PR #4, opened 2020-11-11
switch tl2php to int64_t TL long 

DrDet:
TL long representation in PHP becomes `int64_t` instead of `mixed`.
Transition to it must be performed with taking into account that existing PHP code relies on `mixed` type. PHP code fixes are required for that. 
Moreover, transition must look 'simultaneous' for KPHP and vkext. For this reason the temporary 'toggle' is implemented as a private constant in one PHP TL class. Internal behavior of KPHP and vkext depends on this toggle. When it's off, TL long is represented as `mixed`, when it's on - as `int64_t`. When transition is completed, the toggle will be deleted and behavior will correspond `int64_t` TL long representation.


AlexK0 (approved):
Please, check TeamCity build before merging.

PR #11, opened 2020-11-13
[tests] Add github actions config.

AlexK0:
Hi.

There are 5 builds:
1) buster/g++/c++14
2) focal/g++/c++14
3) focal/clang++/c++17/ubsan
4) focal/g++-10/c++20/asan
5) macos/clang++/c++14

Details
1) All linux builds work via Dockerfile, because it allows to cache prepared env, and reuse it later;
2) Macos build isn't cached, because it fails faster :) May be later it make sense to think about it;
3) Stages: (cmake + make) + run unit tests + build dummy php file and run it;
4) For Macos only cmake works, other fail and they are ignored;
5) The builds are started on creating pull request and pushing something to master.

Examples in my fork:
actions: https://github.com/AlexK0/kphp/actions
pr: https://github.com/AlexK0/kphp/pull/2

### DIFF
--- a/.github/workflows/Build.yml
+++ b/.github/workflows/Build.yml
@@ -1,0 +1,115 @@
+name: Build
+
+on:
+  push:
+    branches:
+    - master
+  pull_request:
+    branches:
+    - master
+
+jobs:
+  build-linux:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        include:
+          - os: buster
+            compiler: g++
+            cpp: 14
+            asan: off
+            ubsan: off
+          - os: focal
+            compiler: g++
+            cpp: 14
+            asan: off
+            ubsan: off
+          - os: focal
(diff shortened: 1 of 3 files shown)


.github/workflows/Build.yml
```diff
@@ -0,0 +1,119 @@
...
+    
+    - name: Setup Environment
+      run: |
+        brew install cmake glib-openssl re2 ccache fmt googletest
+        pip3 install jsonschema
+        
+    # when we fix build for mac, don't forget to combine this step with "Build step"
+    - name: Run cmake
+      run: |
+        mkdir -p ${{runner.workspace}}/build
```
eugene536, line 100: you don't need it, cmake makes a directory when you use `-B` option

eugene536 (approved):

eugene536 (approved):

PR #17, opened 2020-11-16
prevent assumptions inferring for uses of lambda simultaneously

eugene536:
prevent inferring assumptions of typed lamba's uses with reading assumptions at the same time which causes data-race and SIGSEGV from time to time.

### DIFF
--- a/compiler/data/lambda-class-data.cpp
+++ b/compiler/data/lambda-class-data.cpp
@@ -40,6 +40,10 @@ LambdaPtr LambdaClassData::get_from(VertexPtr v) {
 }
 
 void LambdaClassData::infer_uses_assumptions(FunctionPtr parent_function) {
+  auto expected = AssumptionStatus::unknown;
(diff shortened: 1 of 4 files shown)


compiler/data/lambda-class-data.cpp
```diff
@@ -40,6 +40,10 @@ LambdaPtr LambdaClassData::get_from(VertexPtr v) {
 }
 
 void LambdaClassData::infer_uses_assumptions(FunctionPtr parent_function) {
+  auto expected = AssumptionStatus::unknown;
+  if (!assumption_vars_status.compare_exchange_strong(expected, AssumptionStatus::processing)) {
+    return;
```
AlexK0, line 45: Is it make sense to perform `assumption_busy_wait()` here?

eugene536: honestly no, we just need this condition to prevent inferring in the preprocess-functions pass if it's been already inferred before, due-to typed-lambdas.
all in all, everything should be fine

AlexK0 (approved):

PR #22, opened 2020-11-17
openssl critical sections

AlexK0:
Call openssl methods under signal critical sections.

### DIFF
--- a/runtime/critical_section.h
+++ b/runtime/critical_section.h
@@ -23,26 +23,30 @@ struct CriticalSectionGuard : private vk::not_copyable {
 
 class CriticalSectionSmartGuard : private vk::not_copyable {
 public:
-  CriticalSectionSmartGuard() noexcept { enter_critical_section(); }
-
-  void leave_critical_section() noexcept {
-    dl::leave_critical_section();
-    in_critical_section_ = false;
+  CriticalSectionSmartGuard() noexcept {
+    this->enter_critical_section();
   }
 
   void enter_critical_section() noexcept {
-    dl::enter_critical_section();
-    in_critical_section_ = true;
-  }
-
-  ~CriticalSectionSmartGuard() noexcept {
-    if (in_critical_section_) {
-      dl::leave_critical_section();
+    if (!in_critical_section_) {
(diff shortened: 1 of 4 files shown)


runtime/critical_section.h
```diff
@@ -21,6 +21,25 @@ struct CriticalSectionGuard : private vk::not_copyable {
   ~CriticalSectionGuard() noexcept { leave_critical_section(); }
 };
 
+class CriticalSectionSmartGuard : private vk::not_copyable {
+public:
+  CriticalSectionSmartGuard() noexcept { enter_critical_section(); }
+
+  void leave_critical_section() noexcept {
```
DrDet, line 28: Maybe it's worth to add similar `enter_critical_section` method to make it more smart :) ? Having only one of them looks a bit strange

AlexK0: It was added in the next commit

runtime/critical_section.h
```diff
@@ -21,6 +21,30 @@ struct CriticalSectionGuard : private vk::not_copyable {
...
+    in_critical_section_ = false;
+  }
+
+  void enter_critical_section() noexcept {
+    dl::enter_critical_section();
+    in_critical_section_ = true;
+  }
+
+  ~CriticalSectionSmartGuard() noexcept {
+    if (in_critical_section_) {
```
eugene536, line 39: maybe put this if inside the leave_critical_section method?

runtime/critical_section.h
```diff
@@ -26,13 +26,17 @@ class CriticalSectionSmartGuard : private vk::not_copyable {
...
   void enter_critical_section() noexcept {
-    dl::enter_critical_section();
-    in_critical_section_ = true;
+    if (!in_critical_section_) {
+      dl::enter_critical_section();
+      in_critical_section_ = true;
+    }
   }
 
   ~CriticalSectionSmartGuard() noexcept {
```
eugene536, line 42: and now you can just call the `leave_critical_section` method here, without `if`

eugene536 (approved):

PR #24, opened 2020-11-17
add support for simple ${var} string interpolation expressions

quasilyte:
This syntax is preferred by some programmers over {$...}.

The main reasoning behind this change is to make more PHP
programs compile as KPHP programs.

A more compilcated ${<expr>} form is not implemented, so
this will result in a compile error: "${a[0]}".
We may add more complete support for this syntax later.

Note: <expr> inside ${<expr>} is subtle and should
be handled separately, if ever attempted to be implemented.
For example, "${'foo'}" is identical to "$foo".
We don't support ${'foo'} variable syntax yet (which is a
valid freestanding PHP variable syntax btw).

### DIFF
--- a/compiler/lexer.cpp
+++ b/compiler/lexer.cpp
@@ -303,9 +303,12 @@ bool TokenLexerName::parse(LexerData *lexer_data) const {
   }
 
   const char *t = s;
+  bool closing_curly = false; // whether it's a ${varname} and we need to find matching '}'
   if (type == tok_var_name) {
     if (t[0] == '{') {
-      return TokenLexerError("${ is not supported by kPHP").parse(lexer_data);
+      s++; // string value should start after the '{'
+      t++; // consume '{'
+      closing_curly = true;
     }
     if (is_alpha(t[0])) {
       t++;
@@ -342,6 +345,13 @@ bool TokenLexerName::parse(LexerData *lexer_data) const {
   }
 
   vk::string_view name(s, t);
+
+  if (closing_curly) {
+    if (t[0] != '}') {
+      return TokenLexerError("Expected '}' after " + string(s, t)).parse(lexer_data);
+    }
+    t++; // consume '}'
+  }
 
   if (type == tok_func_name) {
     const KeywordType *tp = KeywordsSet::get_type(name.begin(), name.size());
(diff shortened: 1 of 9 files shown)


quasilyte:
Do not merge this until it's fully discussed whether this feature should be implemented.

eugene536:
Well yes, but actually no

this piece of code compiles without any error
```
<?php

$a = 10;
${a} = 30;
var_dump("${a}");

function foo(${a}) {
    var_dump(${a});
}

foo(20);

/**@var ${uu} int*/
$uu = 20;

```

quasilyte:
> this piece of code compiles without any error

Hmm. KPHP gives compile errors for it.

```
Compilation error [gen by /home/isharipov/CODE/kphp3/compiler/lexer.cpp at 294]
In stage = [Split file to tokens]:
  [file = /home/isharipov/CODE/php/hello.php]
  [function = unknown function:4]
  [line = 4]
//   4:${a} = 30;
${ is not supported by kPHP


Compilation error [gen by /home/isharipov/CODE/kphp3/compiler/lexer.cpp at 294]
In stage = [Split file to tokens]:
  [file = /home/isharipov/CODE/php/hello.php]
  [function = unknown function:4]
  [line = 4]
//   4:${a} = 30;
Can't parse


Compilation error [gen by /home/isharipov/CODE/kphp3/compiler/lexer.cpp at 1176]
In stage = [Split file to tokens]:
  [file = /home/isharipov/CODE/php/hello.php]
  [function = unknown function:4]
  [line = 4]
//   4:${a} = 30;
failed to parse
```

PHP also doesn't allow the `${x}` syntax in the param names. The main question for this PR is whether we want to support `${x}` syntax in strings. It's quite rare (one can always use `{$x}` if explicit delimiters are needed.

quasilyte:
Ah. Now I see what you've meant. I'll fix that.

compiler/lexer.cpp
```diff
@@ -780,8 +812,9 @@ void TokenLexerHeredocString::init() {
   h->add_rule("\\[0-7]", &vk::singleton<TokenLexerOctChar>::get());
   h->add_rule("\\x[0-9A-Fa-f]", &vk::singleton<TokenLexerHexChar>::get());
 
-  h->add_rule("$[A-Za-z{]", &vk::singleton<TokenLexerStringSimpleExpr>::get());
+  h->add_rule("$[A-Za-z]", &vk::singleton<TokenLexerStringSimpleExpr>::get());
```
eugene536, line 815: it seems like a subtle bug here is hidden, maybe we unify these Lexers?
image

quasilyte: Fixed. Relevant tests are added.

eugene536 (approved):

quasilyte:
Approve disappeared after rebase. I tested the branch, it gives no translation diff.

eugene536 (approved):

unserialize (approved):

PR #27, opened 2020-11-18
add `final` specifier to vertex classes if there are no derived

eugene536:
add final specifier to almost all vertices to avoid virtual call for `get_string` and similar methods

### DIFF
--- a/compiler/vertex-gen.py
+++ b/compiler/vertex-gen.py
@@ -54,11 +54,11 @@ def output_include_directive(f, name):
         f.write('#include "%s/vertex-%s.h"\n' % (REL_DIR, name))
 
 
-def output_class_header(f, base_name, name):
+def output_class_header(f, base_name, name, final_specifier=""):
     f.write("""
 template<>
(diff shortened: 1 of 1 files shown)


quasilyte (dismissed):

compiler/vertex-gen.py
```diff
@@ -54,11 +54,11 @@ def output_include_directive(f, name):
         f.write('#include "%s/vertex-%s.h"\n' % (REL_DIR, name))
 
 
-def output_class_header(f, base_name, name):
+def output_class_header(f, base_name, name, final=""):
     f.write("""
 template<>
-class vertex_inner<{name}> : public vertex_inner<{base_name}> {{
-public:""".format(name=name, base_name=base_name))
+class vertex_inner<{name}> {final}: public vertex_inner<{base_name}> {{
```
unserialize, line 60: Maybe, `final_specifier`, the same as below?
(`final` seemed to be boolean at first)

eugene536: ok

unserialize (approved):

PR #28, opened 2020-11-18
allow semi-reserved keywords in class member names

quasilyte:
Make it possible to use semi-reserved keywords for class
methods and constants.

A new GenTree::get_identifier() method is used in all places
where we allow such keywords usage.

Some code in lexer hack_last_tokens() method is re-arranged,
so elseif token is not replaced with else+if tokens in
places where we want it to stay a single elseif token.


quasilyte:
TODO: add tests for upper_case keywords, like `const NULL = 10`.

quasilyte:
Ready for review.

AlexK0 (approved):

quasilyte:
Tests: OK.

AlexK0 (approved):

PR #30, opened 2020-11-19
add comments for the generated vertex types

quasilyte:
Add "comment" optional field in vertex-desc.json for
vertices that may benefit from a little summary.

That comment is added to the docummented vertex type
as a C++ comment.

So, "comment": "lhs() and rhs()" for op_log_and_let will produce
something like this:

	// lhs() and rhs()
	template<>
	class vertex_inner<op_log_and_let> : public vertex_inner<meta_op_binary> {

Another example:

	// a simple var, like $foo; variable name is stored in str_val
	template<>
	class vertex_inner<op_var> final: public vertex_inner<meta_op_base> {

The <name>() notation in most comments refer to the named getter
of the vertex. In the example above, lhs() and rhs() are
accessors from meta_op_binary.

This improves readability of both places: json manifest and
generated C++ code (where we end up after "goto definition").

### DIFF
--- a/compiler/vertex-desc.config.json
+++ b/compiler/vertex-desc.config.json
@@ -88,6 +88,9 @@
       }
     ],
     "properties": {
+      "comment": {
+        "type": "string"
+      },
       "name": {
         "type": "string"
       },
--- a/compiler/vertex-desc.json
+++ b/compiler/vertex-desc.json
@@ -1,5 +1,6 @@
 [
   {
+    "comment": "artificial op that represents a part of code that we failed to parse",
     "name": "op_err",
     "base_name": "meta_op_base"
   },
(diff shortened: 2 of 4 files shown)


quasilyte:
image

quasilyte:
Some comments may seem dull, but I think we may improve in that regard. :)

My first frustration was trying to remember what `op_log_and_let` meant.
Seeing `lhs and rhs` for that and `lhs && rhs` for `op_log_and` makes me figure out which one is which faster.

eugene536 (approved):
Great job

Obasanya (approved):
[]() tannylord001@gmail.com

Obasanya (approved):

PR #35, opened 2020-11-21
remove support for multiple main files

quasilyte:
usage: kphp <main-files>
	=>
	usage: kphp <main-file>

Supporting multiple main files doesn't seem to be useful
and it complicates code in some places where we need to
keep in mind that there could be several main files
although only the first one has real meaning.

Giving multiple positional arg to kphp command will result in an error:

	Expected exactly 1 positional argument: main PHP file

### DIFF
--- a/compiler/code-gen/files/init-scripts.cpp
+++ b/compiler/code-gen/files/init-scripts.cpp
@@ -161,8 +161,8 @@ void LibGlobalVarsReset::compile(CodeGenerator &W) const {
 }
 
 
-InitScriptsCpp::InitScriptsCpp(vector<SrcFilePtr> &&main_file_ids, vector<FunctionPtr> &&all_functions) :
-  main_file_ids(std::move(main_file_ids)),
+InitScriptsCpp::InitScriptsCpp(SrcFilePtr main_file_id, vector<FunctionPtr> &&all_functions) :
+  main_file_id(main_file_id),
   all_functions(std::move(all_functions)) {}
 
 void InitScriptsCpp::compile(CodeGenerator &W) const {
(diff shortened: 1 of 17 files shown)


compiler/code-gen/files/init-scripts.cpp
```diff
@@ -184,28 +182,24 @@ void InitScriptsCpp::compile(CodeGenerator &W) const {
...
-    W << FunctionName(i->main_function) << "$global_reset();" << NL;
+  W << FunctionName(main_file_id->main_function) << "$global_reset();" << NL;
 
-    W << "set_script (" <<
-      "\"@" << i->short_file_name << "\", " <<
-      FunctionName(i->main_function) << "$run, " <<
-      FunctionName(i->main_function) << "$global_reset);" << NL;
-  }
+  // TODO: do we need a scripts map if it contains only 1 script?
+  // A further cleanup may be possible.
```
AlexK0, line 198: Let's do it right now. It's easy, just go to `server/php-script.cpp` and 
1) replace `std::map` with single `script_t *`
2) drop `const char *name` from all functions

quasilyte: Done.

AlexK0 (approved):

unserialize (dismissed):

compiler/compiler-settings.h
```diff
@@ -91,7 +91,11 @@ class CompilerSettings : vk::not_copyable {
     colored
   };
 
-  KphpOption<std::vector<std::string>> main_files;
+  // right now we require that there is only 1 file in the input
+  // files list: a main file, but we may allow more than 1 input
+  // files in future (?); it also makes giving an error for
+  // more than 1 positional argument easier
+  KphpOption<std::vector<std::string>> input_files;
```
eugene536, line 98: I don't think that we need to keep a vector of strings here and take a back element every time.
It seems that leaving only one file here would be better.

quasilyte: I haven't found a simple way to give an error for the case where user provides more than 1 positional argument.
Maybe @AlexK0 can give a hint here?
My best bet was to modify the args parser and warn if number of positional arguments is greater than 1.

quasilyte: Done.

eugene536 (approved):

unserialize (approved):

AlexK0 (approved):

PR #36, opened 2020-11-24
replace f"" string with string concat in vertex-gen.py

quasilyte:
So tests can be executed in environments that don't have Python 3.6.


unserialize (commented):
Maybe, you'll add a note, what's the purpose? Because without any comments it seems strange

eugene536 (approved):

PR #37, opened 2020-11-24
Docs: improve Debian repository instructions

loskutov:
`apt-key` is now deprecated. Also, using config directories is considered good practice.


eugene536 (approved):
it works and looks better, thx :+1:

PR #41, opened 2020-11-25
permit FQN forms for the global function names

quasilyte:
Since FQN names always resolve to the name without leading '\',
we can start with this and implement real function namespaces
support later (https://www.php.net/manual/en/language.namespaces.rules.php).

This change makes it possible to compile PHP projects that refer
to global functions using full qualified names.
Here is an example: https://github.com/symfony/polyfill-ctype/blob/main/Ctype.php#L36

### DIFF
--- a/compiler/gentree.cpp
+++ b/compiler/gentree.cpp
@@ -7,6 +7,7 @@
 #include <sstream>
 
 #include "common/algorithms/find.h"
+#include "common/algorithms/contains.h"
 #include "common/type_traits/constexpr_if.h"
 
 #include "common/php-functions.h"
@@ -539,7 +540,17 @@ VertexPtr GenTree::get_expr_top(bool was_arrow) {
         break;
       }
       cur--;
-      res = get_func_call<op_func_call, op_none>();
+
+      // we don't support namespaces for functions yet, but we
+      // permit \func() syntax and interpret it as func();
+      // this logic is compatible with what we'll get after the
+      // function namespaces will be implemented
+      auto func_call = get_func_call<op_func_call, op_none>();
+      if (func_call->str_val[0] == '\\' && !vk::contains(func_call->str_val, "::")) {
+        func_call->str_val.erase(0, 1);
+      }
+
+      res = func_call;
       return_flag = was_arrow;
       break;
     }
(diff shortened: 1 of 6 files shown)


compiler/gentree.cpp
```diff
@@ -539,7 +539,17 @@ VertexPtr GenTree::get_expr_top(bool was_arrow) {
...
       cur--;
-      res = get_func_call<op_func_call, op_none>();
+
+      // we don't support namespaces for functions yet, but we
+      // permit \func() syntax and interpret it as func();
+      // this logic is compatible with what we'll get after the
+      // function namespaces will be implemented
+      auto func_call = get_func_call<op_func_call, op_none>();
+      if (func_call->str_val[0] == '\\') {
+        func_call->str_val = func_call->str_val.substr(1);
```
eugene536, line 549: `func_calll->str_val.erase(0, 1)` won't allocate extra string

eugene536 (commented):
I think that supporting functions in namespaces is not a huge deal and we can implement it.
But if you don't consider it useful right now just left it as it is.

compiler/gentree.cpp
quasilyte: I was using erase (with iterator), but the code that removes the class leading `\` did `substr(1)`, so I followed that pattern. :sweat_smile:

quasilyte:
@eugene536 I'm mostly concerned about unqualified function names resolution that can't be done in `gentree` correctly.
`foo()` call can refer to the global `\foo` function or `foo` function inside the current namespace. Until we have all functions of that namespace available, I don't think it's possible to resolve the `foo()` call properly in the context of namespaces. It would require more changes than the proposed small fix.

compiler/gentree.cpp
eugene536: I think than for classes we used `string_view`

quasilyte: You're right. I'll change the code to use erase.

AlexK0 (dismissed):

quasilyte:
Rebased, tested for diff.

unserialize (approved):

PR #45, opened 2020-11-26
fix race in resolving defines and checking defaults

eugene536:
When we check that the default argument of interface or abstract method has the same value as every overloaded method from derived classes we have a data race. `InlineDefinesUsagesPass` tries can inline defines used as default value at the same time when we investigate such a situation.

Now we separate that check and do it after the synchronization pipe `GenerateVirtualMethods`.

### DIFF
--- a/compiler/pipes/check-abstract-function-defaults.cpp
+++ b/compiler/pipes/check-abstract-function-defaults.cpp
@@ -1,0 +1,49 @@
+// Compiler for PHP (aka KPHP)
+// Copyright (c) 2020 LLC «V Kontakte»
+// Distributed under the GPL v3 License, see LICENSE.notice.txt
+
+#include "compiler/pipes/check-abstract-function-defaults.h"
+
+#include "common/containers/final_action.h"
+#include "common/termformat/termformat.h"
+
+#include "compiler/data/class-data.h"
+#include "compiler/data/function-data.h"
+#include "compiler/pipes/calc-real-defines-values.h"
+
+void CheckAbstractFunctionDefaults::execute(FunctionPtr interface_function, DataStream<FunctionPtr> &os) {
+  stage::set_function(interface_function);
+  auto forward_function_next = vk::finally([&] { os << interface_function; });
+
+  if (!interface_function->modifiers.is_abstract()) {
+    return;
+  }
+
+  auto get_default = [](VertexPtr v) {
+    auto param = v.try_as<op_func_param>();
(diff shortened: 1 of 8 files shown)


compiler/vertex-meta_op_base.h
```diff
@@ -214,3 +214,13 @@ using Vertex = vertex_inner<meta_op_base>;
...
 
+inline bool deep_equal(VertexPtr lhs, VertexPtr rhs) {
+  if (lhs == rhs) {
+    return true;
+  }
+  if (!lhs || !rhs) {
+    return false;
+  }
+
+  return lhs->type() == rhs->type() && lhs->size() == rhs->size() && std::equal(lhs->begin(), lhs->end(), rhs->begin(), deep_equal);
```
AlexK0, line 225: `lhs->type() == rhs->type() && std::equal(lhs->begin(), lhs->end(), rhs->begin(), rhs->end(), deep_equal);`

compiler/pipes/check-abstract-function-defaults.cpp
```diff
@@ -0,0 +1,49 @@
...
+    auto *derived_method = derived_class->members.get_instance_method(interface_function->local_name());
+    if (!derived_method || derived_method->function == interface_function) {
+      continue;
+    }
+    auto derived_params = derived_method->function->get_params();
+
+    for (auto arg_id = interface_function->get_min_argn(); arg_id < interface_params.size(); ++arg_id) {
+      auto derived_default = get_default(derived_params[arg_id]);
+      auto interface_default = get_default(interface_params[arg_id]);
+      kphp_error(deep_equal(derived_default, interface_default),
```
AlexK0, line 42: Please, consider the following example:
```php
<?php

interface II {
    function foo($x = 1234);
}

class AA implements II {
    function foo($x = 56789) { var_dump($x); }
}

class BB implements II {
    function foo($x = 0) { var_dump($x); }
}

function foo(II $ii) { $ii->foo(); }

function demo() {
    foo(new AA);
    foo(new BB);
}

demo();
```

Old code (before this patch) forbids such kind of default value redefining, but new one allows and generates inconsistent with PHP result
PHP out:
```
int(56789)
int(0)
```
new KPHP out:
```
int(1234)
int(1234)
```
:(

eugene536: It's a bug, thx, fixed it

AlexK0 (approved):

PR #47, opened 2020-11-27
drop uint, long, ulong completely

unserialize:


runtime/string_functions.cpp
```diff
@@ -892,7 +891,7 @@ string f$pack(const string &pattern, const array<mixed> &a) {
             case 'P':
             case 'Q': {
               // stored in the host machine order by the default (Q flag)
-              unsigned long long value_byteordered = ULong(arg.to_string()).l;
+              unsigned long long value_byteordered = static_cast<unsigned long long>(arg.to_string().to_int());
```
AlexK0, line 894: 1) I'm not sure if `string::to_int()` parse string with large unsigned integer correct, please check.
2) Types ULong and Long check overflow and fire warning, but `string::to_int()` doesn't.

runtime/string_functions.cpp
```diff
@@ -2537,7 +2537,8 @@ array<mixed> f$unpack(const string &pattern, const string &data) {
                 php_warning("Not enough data to unpack with format \"%s\"", pattern.c_str());
                 return result;
               }
-              value = f$strval(Long(*(long long *)(data.c_str() + data_pos)));
+              long long value_ll = *(long long *)(data.c_str() + data_pos);
```
DrDet, line 2540:
```suggestion
              long long value_ll = *reinterpret_cast<const long long *>(data.c_str() + data_pos);
```

eugene536:
Can we delete these operators too?
`string_buffer &operator<<(string_buffer &sb, long long x)`
`string_buffer &operator<<(string_buffer &sb, unsigned long long x)`

runtime/string_functions.cpp
unserialize: How would you offer to fix this? Or just to ignore?

runtime/string_functions.cpp
unserialize: What's wrong with the original code?)

DrDet: 1. C-style cast here silently discards const, which is absolutely not needed and potentially dangerous. One of the reason to avoid C-style casts is they perform sequence of different casts until the expression is casted even if you need particular one. It happens implicitly and you can't see, for example, that it has dropped const.
2. In general, there's no reason to use C-style casts in C++ code. C++ style casts are more explicit, easier for search and checked by the compiler.

runtime/string_functions.cpp
AlexK0: I have no idea if it matters. Lets ignore.

AlexK0 (approved):

PR #50, opened 2020-11-30
Performance analyze.

AlexK0:
Hi there,

There is a new feature which allows to perform performance static analyze. It uses some heuristics to determine whether code can be improved in terms of performance. There are 4 inspections are implemented:
- _**array-merge-into**_: searches a pattern `expr = array_merge(expr, ...);` and suggests to replace it with `array_merge_into(expr, ...);`. `expr` could be a variable (`$x`), an instance field (`$i->x`), or an array element (`$a["x"]["y"]`);
- **_array-reserve_**: searches patterns where arrays are filling in loops and suggests to use  `array_reserve()` in advance out of loop;
- **_constant-execution-in-loop_**: searches the constant execution patterns and suggests to save them into separate variable in advance out of loop. The following patterns are supported: 
  1) indexing array by constant expressions keys;
  2) building strings from constant expressions;
  3) calling `@kphp-pure-function` functions with constant expression arguments.

  Constant expressions here mean any constants (string, integers, arrays), non modified (inside loop) local variables, array indexing, string building and pure function calling.
- **_implicit-array-cast_**: searches implicit runtime array casts of:
  1) function call arguments (only if function parameters is non modified and passed by const reference);
  2) function return statement; 
  3) array making (e.g. if create `mixed[] $x` from `int[][] $i` with expression `$x = ["i" => $i];`) ;
  4) pushing and inserting arrays into another arrays;
  5) array assignment between different variables.

**_array-reserve_** and **_constant-execution-in-loop_** work only in loops without break, continue and return statements. Also they don't inspect conditional code in if-else, switch-case, ternary true/false operands, right operand of &&, II, ??, and they don't support global, static, instance field variables.

All these inspections can be enabled with `@kphp-analyze-performance` or `@kphp-warn-performance`  annotations. The annotations are propagated by call graph for all reachable functions. Special tag `all` means enable all inspections. Inspection can be disabled with `!`, e.g `@kphp-analyze-performance all !implicit-array-cast` means enable all but **_implicit-array-cast_** or `@kphp-analyze-performance !all` means disable all (can be useful if suppressing warnings is required).
-  `@kphp-analyze-performance` - enables inspections for analyze, it means that after compilation, KPHP will generate json (hard-coded `${KPHP_DEST_DIR}/performance_issues.json`) file with all founded issues. It is supposed that this annotation will never be committed to the repo.
- `@kphp-warn-performance` - enables inspections for warnings, it means that KPHP will generate warnings for all founded issues. It can be useful for fixing optimizations and it is supposed to be used after `@kphp-analyze-performance`.

Also, `KPHP_ENABLE_FULL_PERFORMANCE_ANALYZE (--enable-full-performance-analyze)` can be used for enabling all inspection in analyze mode. The main aim of this option is to enable analyze for all our tests.

Optimization _implicit constant array cast in compile time_ has been improved as well:
- Support more cases for op_array (array creation) and op_return (return statement);
- Don't cast `mixed[]` to `mixed` and always empty array (with unknown type) to array with type;
- Implicitly change type of creating array (op_array) to required type, in order to avoid of runtime conversion;
- Enable cast optimization for variadic parameters as well;


compiler/pipes/analyze-performance.h
```diff
@@ -0,0 +1,97 @@
...
+  bool is_enabled() const noexcept {
+    return get_function_inspections(current_function) & inspection;
+  }
+
+  void on_array_merge(VertexPtr first_arg) noexcept;
+  void trigger_inspection(PerformanceInspections::Inspections inspection, std::string message) noexcept;
+  void trigger_inspection_on_second_pass(VertexPtr inspected_vertex, PerformanceInspections::Inspections inspection, std::string message) noexcept;
+
+  bool warnings_enabled_{false};
+  std::vector<Issue> performance_issues_for_analyse_;
```
DrDet, line 61: ..to_analyze_ or ..for_analysis sounds better

compiler/data/function-data.h
```diff
@@ -127,6 +128,10 @@ class FunctionData {
     enable_as_inline_child,
   } profiler_state = profiler_status::disable;
 
+  PerformanceInspections performance_inspections_for_analyse;
```
DrDet, line 131: and here

compiler/data/performance-inspections.h
```diff
@@ -0,0 +1,50 @@
...
+// Copyright (c) 2020 LLC «V Kontakte»
+// Distributed under the GPL v3 License, see LICENSE.notice.txt
+#pragma once
+#include "compiler/data/data_ptr.h"
+#include "common/wrappers/string_view.h"
+
+
+class PerformanceInspections {
+public:
+  enum Inspections : uint8_t {
```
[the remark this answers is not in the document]
AlexK0: I don't use enum class because this enum is inside `PerformanceInspections` class, so it is required to use this class namespace `PerformanceInspections::`, and `PerformanceInspections::Inspections::` looks wired for me.

eugene536 (changes requested):

compiler/compiler-core.cpp
```diff
@@ -537,6 +537,17 @@ void CompilerCore::init_composer_class_loader() {
   }
 }
 
+vk::string_view CompilerCore::get_base_relative_filename(SrcFilePtr file) const noexcept {
```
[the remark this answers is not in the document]
AlexK0: Sure. I've removed all of them. Hope nothing has broken :shamrock:

compiler/compiler.cpp
```diff
@@ -312,6 +314,13 @@ bool compiler_execute(CompilerSettings *settings) {
     stage::error();
   }
 
+  if (!PerformanceIssuesReport::get().empty()) {
+    if (auto *report_file = fopen(G->settings().performance_analyze_report_path.get().c_str(), "w")) {
+      PerformanceIssuesReport::get().flush_to(report_file);
+      fclose(report_file);
+    }
+  }
```
unserialize, line 322: If there are no issues, the .json file is not re-created, so it remains an old one. 
Here is the situation:
1) I mark a function with @kphp-analyze-performance and see a report with several warnings
2) I fix all errors and execute KPHP again
3) The file remains — I suppose it's confusing

I'd suggest to create an empty file (just '[\n]') in case of any errors. (if there is a function with @kphp-analyze-performance of course)

compiler/data/function-data.h
```diff
@@ -145,8 +150,9 @@ class FunctionData {
 
   string get_resumable_path() const;
   string get_throws_call_chain() const;
-  static string get_human_readable_name(const std::string &name);
-  string get_human_readable_name() const;
+  std::string get_performance_inspections_warning_chain(PerformanceInspections::Inspections inspection, bool search_disabled_inspection = false) const noexcept;
+  static string get_human_readable_name(const std::string &name, bool add_details = true);
```
unserialize, line 154: btw, a static function seems to be useless, as it's called only from an instance one

compiler/data/performance-inspections.cpp
```diff
@@ -0,0 +1,87 @@
+// Compiler for PHP (aka KPHP)
+// Copyright (c) 2020 LLC «V Kontakte»
+// Distributed under the GPL v3 License, see LICENSE.notice.txt
+#include "compiler/data/performance-inspections.h"
```
unserialize, line 4: a newline after the header please

compiler/data/performance-inspections.h
```diff
@@ -0,0 +1,50 @@
+// Compiler for PHP (aka KPHP)
+// Copyright (c) 2020 LLC «V Kontakte»
+// Distributed under the GPL v3 License, see LICENSE.notice.txt
+#pragma once
```
unserialize, line 4: a newline after the header please

compiler/pipes/analyze-performance.cpp
```diff
@@ -0,0 +1,636 @@
+// Compiler for PHP (aka KPHP)
+// Copyright (c) 2020 LLC «V Kontakte»
+// Distributed under the GPL v3 License, see LICENSE.notice.txt
+#include "compiler/pipes/analyze-performance.h"
```
unserialize, line 4: A newline after the header please

compiler/pipes/analyze-performance.cpp
```diff
@@ -0,0 +1,636 @@
...
+#include "compiler/compiler-core.h"
+#include "compiler/data/src-file.h"
+#include "compiler/inferring/public.h"
+#include "compiler/gentree.h"
+
+namespace {
+
+VertexPtr remove_conv_wrap(VertexPtr vertex) noexcept {
+  if (vk::any_of_equal(vertex->type(), op_conv_int, op_conv_float, op_conv_string, op_conv_array,
+                       op_conv_bool, op_conv_mixed, op_conv_drop_false, op_conv_drop_null)) {
```
unserialize, line 17: Maybe, to check for OpInfo::type(vertex->type()) == conv_op?
(also op_move is conv_op by the way, I don't know whether we should handle it here)

compiler/pipes/analyze-performance.cpp
```diff
@@ -0,0 +1,636 @@
...
+
+template<bool full>
+std::string get_description_for_help_impl(VertexPtr expr);
+
+std::string join_string_build_elements(VertexRange elements) noexcept {
+  return vk::join(elements, ".", get_description_for_help_impl<false>);
+}
+
+template<bool full>
+std::string get_description_for_help_impl(VertexPtr expr) {
```
unserialize, line 40: Why full is not just a boolean argument?

compiler/pipes/analyze-performance.cpp
```diff
@@ -0,0 +1,636 @@
...
+      }
+      break;
+    case op_instance_prop:
+      if (lhs.as<op_instance_prop>()->var_id != rhs.as<op_instance_prop>()->var_id) {
+        return false;
+      }
+      break;
+    case op_int_const:
+    case op_float_const:
+      if (lhs.as<meta_op_num>()->str_val != rhs.as<meta_op_num>()->str_val) {
```
unserialize, line 117: '0' != '0x0' in this case, I amn't sure if it matters

compiler/pipes/analyze-performance.cpp
```diff
@@ -0,0 +1,636 @@
...
+    }
+    return false;
+  }
+  if (is_pure_nary_operation(vertex)) {
+    return vk::all_of(*vertex, [this](VertexPtr v) { return is_constant_expression_in_this_loop(v); });
+  }
+  return false;
+}
+
+void AnalyzePerformance::on_array_merge(VertexPtr first_arg) noexcept {
```
unserialize, line 404: trigger_array_merge?

compiler/pipes/analyze-performance.h
```diff
@@ -0,0 +1,97 @@
+// Compiler for PHP (aka KPHP)
+// Copyright (c) 2020 LLC «V Kontakte»
+// Distributed under the GPL v3 License, see LICENSE.notice.txt
+#pragma once
```
unserialize, line 4: a newline after the header please

tests/phpt/performance_inspections/11_implicit_array_cast_warn.php
```diff
@@ -0,0 +1,48 @@
+@kphp_should_warn
+/variable \$a1 is implicitly converted from array< int > to array< mixed >/
+/variable \$b1 is implicitly converted from array< string > to array< mixed >/
+/variable \$a2 is implicitly converted from array< int > to mixed/
+/variable \$b2 is implicitly converted from array< string > to mixed/
+/variable \$a3 is implicitly converted from array< int > to mixed/
+/variable \$b3 is implicitly converted from array< string > to mixed/
+/variable \$a4 is implicitly converted from array< int > to mixed/
+/variable \$b4 is implicitly converted from array< string > to mixed/
```
unserialize, line 9: Maybe, add a more complicated conversion test also? For example:
```
class A {
    /** @var tuple(float, mixed[]) */
    public $a;
}

    (new A)->a = tuple(4, [1,2,3]);

```
Leads to 
expresion <...> is implicitly converted from tuple<int , array< int >> to tuple<float , array< mixed >>

tests/phpt/performance_inspections/1_warn_tag_conflict.php
```diff
@@ -0,0 +1,20 @@
+@kphp_should_fail
+/@kphp-warn-performance conflict, one caller enables 'implicit-array-cast' while other disables it\nEnabled by: a1 -> a2 -> ab\nDisabled by: b1 -> b2 -> ab\
```
unserialize, line 2: It seems that '\\' at the end should be '/'

tests/phpt/performance_inspections/2_analyze_tag_conflict.php
```diff
@@ -0,0 +1,20 @@
+@kphp_should_fail
+/@kphp-analyze-performance conflict, one caller enables 'implicit-array-cast' while other disables it\
```
unserialize, line 2: '\\' at the end?

tests/phpt/performance_inspections/3_warn_no_tag_error.php
```diff
@@ -0,0 +1,10 @@
+@kphp_should_fail
+/@kphp-warn-performance bad tag: there are no any inspection\
```
unserialize, line 2: '\\' at the end?

tests/phpt/performance_inspections/4_analyze_no_tag_error.php
```diff
@@ -0,0 +1,10 @@
+@kphp_should_fail
+/@kphp-analyze-performance bad tag: there are no any inspection\
```
unserialize, line 2: '\\' at the end?

compiler/pipes/analyze-performance.cpp
```diff
@@ -0,0 +1,636 @@
...
+    loop_data_for_second_pass_.back().second_pass_vertexes.emplace_back(vertex);
+    second_pass_saved_top_vertex_ = vertex;
+  }
+}
+
+void AnalyzePerformance::run_second_pass_on_loop_exit(VertexPtr vertex, uint64_t enabled_inspections, VertexPtr loop_vertex) noexcept {
+  if (enabled_inspections & PerformanceInspections::constant_execution_in_loop) {
+    if (vk::any_of_equal(vertex->type(), op_index, op_func_call)) {
+      if (is_constant_expression_in_this_loop(vertex) && is_op_index_sequence_with_non_int_keys(vertex)) {
+        auto message = get_description_for_help(vertex) + " can be saved in separate variable out of loop";
```
unserialize, line 283: a separate variable

compiler/data/function-data.h
AlexK0: unfortunately it is used in preprocess-function.cpp

unserialize: 😳 CLion didn't show me that usage, okay..

eugene536 (approved):

compiler/pipes/analyze-performance.cpp
AlexK0: doesn't matter, just micro optimization

compiler/pipes/analyze-performance.cpp
AlexK0: I suppose we can leave as is for now.

unserialize (approved):

unserialize (approved):

PR #51, opened 2020-11-30
change the assert message header to be a proper location ref for IDE

quasilyte:
It makes it easier to use the location in the source code editor:

	Compilation error [gen by $KPHP_ROOT/compiler/pipes/resolve-self-static-parent.cpp at 87]
	=>
	Compilation error [gen by $KPHP_ROOT/compiler/pipes/resolve-self-static-parent.cpp:87]

Most terminals also support this format to open the file ref in the associated app.


AlexK0 (approved):
Please, check tests in TC before merge, some of them may expect messages with 'at'.

quasilyte:
TC tests are passing.

PR #54, opened 2020-12-02
Some fixes in RPC lib

DrDet:
Here are 2 fixes:
- Update script runs counter on getting typed RPC TL response. Without it `rpc_tl_pending_queries_count()` always returns 0 if only typed RPC requests are used.
- Fix fatal error on inconsistent RPC TL API usage. Without it typed sending followed by untyped getting result or vice versa raises fatal error.

### DIFF
--- a/runtime/tl/rpc_request.h
+++ b/runtime/tl/rpc_request.h
@@ -40,10 +40,15 @@ protected:
 
 class RpcRequestResult : public ManagedThroughDlAllocator {
 public:
-  explicit RpcRequestResult(std::unique_ptr<tl_func_base> &&result_fetcher)
-    : result_fetcher_(std::move(result_fetcher)) {}
+  const bool is_typed{};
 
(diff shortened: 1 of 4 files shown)


runtime/tl/rpc_request.h
```diff
@@ -47,6 +47,7 @@ class RpcRequestResult : public ManagedThroughDlAllocator {
 
   virtual class_instance<C$VK$TL$RpcResponse> fetch_typed_response() = 0;
   virtual std::unique_ptr<tl_func_base> extract_untyped_fetcher() = 0;
+  virtual bool is_typed() const = 0;
```
AlexK0, line 50: nit: you can use public constant bool filed and set it from constructor, this way the virtual function call can be avoided

DrDet: fixed

unserialize (approved):

AlexK0 (approved):

PR #58, opened 2020-12-04
Extra RPC query headers for microservices

DrDet:
Allows to send extra RPC query headers to KPHP microservice. Saves them into `$_SERVER`. 
Without it you can't specify custom timeout for RPC request.

### DIFF
--- a/common/tl/query-header.cpp
+++ b/common/tl/query-header.cpp
@@ -98,7 +98,7 @@ static int tl_fetch_query_flags(tl_query_header_t *header) {
   return 0;
 }
 
-bool tl_fetch_query_header(tl_query_header_t *header) {
+bool tl_fetch_query_header(tl_query_header_t *header, bool increment_queries_counter) {
   assert (header);
   if (vk::tl::fetch_magic(TL_RPC_INVOKE_REQ)) {
     header->qid = tl_fetch_long();
@@ -125,7 +125,7 @@ bool tl_fetch_query_header(tl_query_header_t *header) {
   if (tl_fetch_error()) {
     return false;
   }
-  rpc_queries_received++;
+  rpc_queries_received += increment_queries_counter;
   return true;
 }
 
--- a/common/tl/query-header.cpp
+++ b/common/tl/query-header.cpp
@@ -93,7 +93,7 @@ static int tl_fetch_query_flags(tl_query_header_t *header) {
   return 0;
 }
 
-bool tl_fetch_query_header(tl_query_header_t *header, bool increment_queries_counter) {
+bool tl_fetch_query_header(tl_query_header_t *header) {
   assert (header);
   if (vk::tl::fetch_magic(TL_RPC_INVOKE_REQ)) {
     header->qid = tl_fetch_long();
--- a/common/tl/query-header.cpp
+++ b/common/tl/query-header.cpp
@@ -21,11 +21,6 @@
 #include "common/tl/store.h"
 #include "common/wrappers/optional.h"
 
-static long long rpc_queries_received;
-static long long rpc_answers_received;
-static long long rpc_answers_error;
-
-
 static int tl_fetch_query_flags(tl_query_header_t *header) {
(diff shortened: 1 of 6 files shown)


runtime/interface.cpp
```diff
@@ -1305,64 +1306,107 @@ static void parse_http_authorization_header(const string &header_value) {
...
+  }
+  if (header.flags & flag::supported_compression_version) {
+    v$_SERVER.set_value(string("RPC_EXTRA_SUPPORTED_COMPRESSION_VERSION"), header.supported_compression_version);
+  }
+  if (header.flags & flag::random_delay) {
+    v$_SERVER.set_value(string("RPC_EXTRA_RANDOM_DELAY"), header.random_delay);
+  }
+}
+
+static void init_superglobals(const http_query_data *http_data, const rpc_query_data *rpc_data) {
```
AlexK0, line 1354: nit: logically by code this is const references, not pointers, because we are sure, that they are always non nullptr

runtime/interface.cpp
```diff
@@ -1305,64 +1306,107 @@ static void parse_http_authorization_header(const string &header_value) {
...
+    v$_SERVER.set_value(string("RPC_EXTRA_FLAGS"), header.flags);
+  }
+  if (header.flags & flag::wait_binlog_pos) {
+    v$_SERVER.set_value(string("RPC_EXTRA_WAIT_BINLOG_POS"), static_cast<int64_t>(header.wait_binlog_pos));
+  }
+  if (header.flags & flag::string_forward_keys) {
+    array<string> string_forward_keys;
+    string_forward_keys.reserve(header.string_forward_keys.size(), 0, true);
+    for (const auto &str_key : header.string_forward_keys) {
+      string_forward_keys.emplace_back(string().assign(str_key.c_str()));
```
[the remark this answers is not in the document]
DrDet: fixed

server/php-engine.cpp
```diff
@@ -1545,25 +1554,23 @@ int rpcx_execute(connection *c, int op, raw_message *raw) {
...
-      auto op_from_tl = tl_fetch_int();
-      len -= static_cast<int>(sizeof(op_from_tl));
-      assert(op_from_tl == op);
-      assert(len % sizeof(int) == 0);
 
-      auto req_id = tl_fetch_long();
-      len -= static_cast<int>(sizeof(req_id));
+      int64_t left_bytes_with_headers = tl_fetch_unread();
+      tl_query_header_t header{};
+      assert(tl_fetch_query_header(&header, false) && "Can't fetch RPC query headers, terminate worker");
```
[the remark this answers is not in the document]
DrDet: 1. fixed
2. At first I tried to send RPC error instead of terminating worker, but the problem is that rpc-proxy expects correct `req_id` which is the part of the header... But I think we can assume that `req_id` is almost always fetched correctly and do this anyway. Changed this place

AlexK0 (approved):

common/tl/query-header.cpp
```diff
@@ -98,7 +98,7 @@ static int tl_fetch_query_flags(tl_query_header_t *header) {
   return 0;
 }
 
-bool tl_fetch_query_header(tl_query_header_t *header) {
```
kunyavskiy, line 101: Seams to be a strange hack. Is this stat really useful? Maybe just drop it?

kunyavskiy: Anyway, I think it was a bug, it was not incremented before

runtime/interface.cpp
```diff
@@ -1305,6 +1306,51 @@ static void parse_http_authorization_header(const string &header_value) {
...
```
kunyavskiy: There can be '\0' in this strings. I think it's handled incorrect.

DrDet: fixed

common/tl/query-header.cpp
DrDet: removed this stat completely

AlexK0 (approved):

PR #60, opened 2020-12-07
support PHP8 union type hints and mixed typehint

unserialize:
Previously, type hints (params of functions and return typehints) were parsed separately and stored as strings.

Now, KPHP parses type hints just the same as PHPDoc. This leads to support union type hints, like
function f(int|false $a): string|int
Also, 'mixed' is parsed correctly as a consequence:
function f(mixed $a): mixed

Moreover, a correct syntax is ANY type that can appear in PHPDoc. For instance, these examples are correct for KPHP, though they fail on PHP:
function f(int[] $arr): mixed[]
function f(tuple<int, string> $t): (int|false)[]

Multiple commits are squashed into a single, you can see development history in the PR.

### DIFF
--- a/compiler/data/function-data.cpp
+++ b/compiler/data/function-data.cpp
@@ -284,7 +284,7 @@ void FunctionData::add_kphp_infer_hint(FunctionData::InferHint::InferType infer_
     return;
   }
   type_expr.set_location(root);
-  infer_hints.emplace_back(InferHint{infer_type, param_i, type_expr});
+  infer_hints.emplace_back(infer_type, param_i, type_expr);
 }
 
 bool FunctionData::is_lambda_with_uses() const {
--- a/compiler/data/function-data.cpp
+++ b/compiler/data/function-data.cpp
@@ -42,6 +42,7 @@ FunctionPtr FunctionData::clone_from(const std::string &new_name, FunctionPtr ot
   res->name = new_name;
   res->update_location_in_body();
   res->name_gen_map = {};
+  res->return_typehint = other->return_typehint.clone();
 
   res->assumptions_for_vars = {};
   res->assumption_args_status = AssumptionStatus::unknown;
--- a/compiler/data/function-data.cpp
+++ b/compiler/data/function-data.cpp
@@ -278,12 +278,12 @@ string FunctionData::get_human_readable_name(bool add_details) const {
   return result_name;
 }
 
-void FunctionData::add_kphp_infer_hint(FunctionData::InferHint::infer_mask infer_mask, int param_i, VertexPtr type_rule) {
-  if (!type_rule) {
+void FunctionData::add_kphp_infer_hint(FunctionData::InferHint::InferType infer_type, int param_i, VertexPtr type_expr) {
+  if (!type_expr) {
     return;
   }
-  type_rule.set_location(root);
-  infer_hints.emplace_back(InferHint{infer_mask, param_i, type_rule});
+  type_expr.set_location(root);
+  infer_hints.emplace_back(InferHint{infer_type, param_i, type_expr});
 }
 
 bool FunctionData::is_lambda_with_uses() const {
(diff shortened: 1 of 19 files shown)


compiler/data/function-data.cpp
```diff
@@ -255,12 +256,12 @@ string FunctionData::get_human_readable_name() const {
...
-void FunctionData::add_kphp_infer_hint(FunctionData::InferHint::infer_mask infer_mask, int param_i, VertexPtr type_rule) {
-  if (!type_rule) {
+void FunctionData::add_kphp_infer_hint(FunctionData::InferHint::InferType infer_type, int param_i, VertexPtr type_expr) {
+  if (!type_expr) {
     return;
   }
-  type_rule.set_location(root);
-  infer_hints.emplace_back(InferHint{infer_mask, param_i, type_rule});
+  type_expr.set_location(root);
+  infer_hints.emplace_back(InferHint{infer_type, param_i, type_expr});
```
DrDet, line 264:
```suggestion
  infer_hints.emplace_back(infer_type, param_i, type_expr);
```

compiler/pipes/collect-required-and-classes.cpp
```diff
@@ -77,39 +77,49 @@ class CollectRequiredPass final : public FunctionPassBase {
...
-  inline void require_all_classes_in_phpdoc(vk::string_view phpdoc_str) {
+  inline void require_all_classes_in_field_phpdoc(vk::string_view phpdoc_str) {
     if (auto type_and_var_name = phpdoc_find_tag_as_string(phpdoc_str, php_doc_tag::var)) {
-      require_all_classes_in_phpdoc_type(*type_and_var_name);
+      // we don't use phpdoc_parse_type_and_var_name() here since classes
+      // are not available at this point and we need to fetch these unknown classes
+      std::vector<Token> tokens = phpdoc_to_tokens(*type_and_var_name);
+      std::vector<Token>::const_iterator cur_tok = tokens.begin();
+      PhpDocTypeRuleParser parser(current_function);
+      auto type_expr = parser.parse_from_tokens_silent(cur_tok);
```
DrDet, line 102: It seems that `parse_from_tokens_silent` doesn't handle case when iterator is incremented after it reaches `end` (it's UB in C++). It must know `tokens.end()` at least to handle this. How is it guaranteed it never happens?

eugene536: we add the fake token - `tok_end` to `tokens` therefore we never reach the end of all tokens

compiler/phpdoc.cpp
```diff
@@ -628,3 +616,37 @@ std::vector<std::string> phpdoc_find_tag_as_string_multi(vk::string_view phpdoc,
...
+                                 TermStringFormat::add_text_attribute(class_name, TermStringFormat::underline)));
+      }
+      else {
+        kphp_error(!klass->is_trait(), "You may not use trait as a type-hint");
+      }
+    }
+  }
+
+  for (auto i : *type_expr) {
+    all_resolved &= phpdoc_prepare_type_expr_resolving_classes(cur_function, i);
```
eugene536, line 648: nit) It's a bit strange using a `bool` variables with bitwise `&` it would be better to separate these expressions or use `&&` instead

eugene536 (changes requested):

compiler/pipes/collect-required-and-classes.cpp
```diff
@@ -77,39 +77,49 @@ class CollectRequiredPass final : public FunctionPassBase {
...
-    std::vector<Token>::const_iterator cur_tok = tokens.begin();
-    PhpDocTypeRuleParser parser(current_function);
-    parser.parse_from_tokens_silent(cur_tok);
-
-    for (const auto &class_name : parser.get_unknown_classes()) {
-      require_class(replace_characters(class_name, '\\', '/'));
-    }
+  // Searching for classes inside @var/@param phpdoc as well as inside type hints
+  inline void require_all_classes_in_phpdoc_type(VertexPtr type_expr) {
+    std::function<void(VertexPtr)> find_unknown_and_require = [this, &find_unknown_and_require](VertexPtr type_expr) {
```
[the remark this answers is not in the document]
unserialize: Agree, done

compiler/phpdoc.cpp
unserialize: Okay, now &&

eugene536 (approved):

PR #63, opened 2020-12-08
MacOS build supporting.

rim-man:
Hi guys!

Here is a patch which enables MacOS supporting, (of course only in dev environment).

#### Preparing
1) Make sure you have `brew` and `clang` (at least `Apple clang version 10.0.0`)
2) Install all required depends:
```bash
$ brew install cmake coreutils glib-openssl libiconv re2 fmt h3 yaml-cpp msgpack zstd googletest php@7.4
$ brew link php@7.4 
$ pip3 install jsonschema
```
3) **_OPTIONAL_**: I highly recommend installing `ccache`, it makes you compilation blazing fast!
```bash
$ brew install ccache
```
4) **_OPTIONAL_**: Clone somewhere local `epoll-shim` from https://github.com/VKCOM/epoll-shim and switch to `osx-platform` branch.
Add env variable `EPOLL_SHIM_REPO` to your bash profile (and may be to you IDE, need to check). It allows to avoid of cloning `epoll-shim` on each clean `cmake` call.
```bash
$ echo 'export EPOLL_SHIM_REPO=/local/path/to/epoll-shim/repo' >> ~/.bash_profile
```

#### Compiling
1) If you clone `epoll-shim` in advance, just run `cmake` as usual
```bash
$ cmake -B build
```
If not, you have to allow `cmake` to download `epoll-shim` from github
```bash
$ cmake -DDOWNLOAD_MISSING_LIBRARIES=On -B build
```
2) Make it
```bash
$ make -C build/ -j$(nproc)
```

#### Features
- compiling all targets (with vkext) - ok 🟢 
- running unittests for all targets - ok 🟢  
- kphp2cpp code generation and kphp server compilation - ok 🟢 
- running compiled kphp server in cli mode (with `--once/-o` option) - ok 🟢 
- checking MacOS build with github actions - ok 🟢 
- running compiled kphp server in server mode with workers and listening HTTP port mode - experimental 🟡 
- running tests with `./tests/kphp_tester.py` - experimental 🟡 
- sending stats to statsd port 8125 - experimental 🟡 
- sending `curl` requests from PHP code by IP - experimental 🟡 
- sending `curl` requests from PHP code by hostname - unsupported 🔴 
- running functional tests with `pytest` - unsupported 🔴 
- any backtraces (both from PHP and signal handlers) - unsupported 🔴 
- all RPC stuff - unknown ⚫ 

#### Details
I've tried to put all MacOS stuff in separate file `common/macos-ports.h`.


net/net-connections.cpp
```diff
@@ -1016,17 +1020,22 @@ int accept_new_connections(struct connection *cc) {
...
+#endif
   assert(cc->basic_type == ct_listen);
   do {
     peer_addrlen = sizeof(peer);
     self_addrlen = sizeof(self);
     memset(&peer, 0, sizeof(peer));
     memset(&self, 0, sizeof(self));
 
+#if defined(__APPLE__)
+    const int cfd = accept(cc->fd, (struct sockaddr *)&peer, &peer_addrlen);
```
AlexK0, line 1035: fcntl(cfd, FD_CLOEXEC);

server/php-engine-vars.h
```diff
@@ -8,6 +8,7 @@
 #include <netinet/in.h>
 
 #include "common/kphp-tasks-lease/lease-worker-mode.h"
+#include "common/macos-ports.h"
```
AlexK0, line 11: no need

cmake/external-libraries.cmake
```diff
@@ -40,3 +40,17 @@ if(KPHP_TESTS)
         message(STATUS "---------------------")
     endif()
 endif()
+
+if(APPLE)
+    if (DEFINED ENV{EPOLL_SHIM_REPO})
+        FetchContent_Declare(epoll URL $ENV{EPOLL_SHIM_REPO})
+    else()
+        handle_missing_library("epoll-shim")
+        FetchContent_Declare(epoll GIT_REPOSITORY https://github.com/rim-man/epoll-shim GIT_TAG osx-platform)
```
eugene536, line 49: It would be great having the same code style as the above code

cmake/init-compilation-flags.cmake
```diff
@@ -49,7 +60,9 @@ endif()
 
 if(NOT DEFINED ENV{ENABLE_GRPROF})
```
eugene536, line 61: honestly all of the below options are almost useless, we can safely add `NOT APPLE` just in this condition

flex/flex.cmake
```diff
@@ -5,9 +5,15 @@ set(AUTO_FLEX_DIR ${AUTO_DIR}/flex)
 set(FLEX_DATA_SRC ${AUTO_FLEX_DIR}/vk-flex-data.cpp)
 
 file(MAKE_DIRECTORY ${AUTO_FLEX_DIR})
+
+if(APPLE)
```
eugene536, line 9: it would be great using `find_program` call here

cmake/init-compilation-flags.cmake
rim-man: let me leave as is, also I've added -dead_strip for macos - --gc-sections alternative.

common/kernel-version.cpp
```diff
@@ -27,20 +27,25 @@ static struct utsname* cached_uname() {
   return &kernel_version;
 }
 
-static int kernel_x;
-static int kernel_y;
-static int kernel_z;
+static int kernel_x = 0;
+static int kernel_y = 0;
+static int kernel_z = 0;
+static bool is_macos = false;
```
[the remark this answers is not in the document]
rim-man: `error: use of undeclared identifier 'defined'` :(

unserialize (approved):

AlexK0 (approved):

PR #67, opened 2020-12-09
Allow to change unix_socket_directory in runtime

Tindarid:
__Proof__: https://www.php.net/manual/en/internals2.ze1.zendapi.php

__TL;DR__
`INI_STR(name)` - Returns the current value of entry name as string. Note: This string is not duplicated, but instead points to internal data. Further access requires duplication to local memory.

__Description__:
This commit allows users in php code change unix_socket_directory through `ini_set("vkext.unix_socket_directory", your_new_path")`. Also it fixes possible `SIGSEGV` if someone used mentioned operation before.

### DIFF
--- a/vkext/vkext-rpc.cpp
+++ b/vkext/vkext-rpc.cpp
@@ -69,7 +69,6 @@ struct stats stats;
 double ping_timeout = PING_TIMEOUT;
 
 static int use_unix;
-static const char *unix_socket_directory;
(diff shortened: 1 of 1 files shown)


eugene536 (approved):
Seems reasonable. Tests have passed, thx for the PR.

DrDet (approved):

PR #71, opened 2020-12-10
use std::stack for CFG::cacl_used to avoid stack overflow

eugene536:
For huge functions recursion could cause a SIGSEGV with the corrupted stack, it's better to be on the safe side.


DrDet (commented):
Have you checked diff for the site?

eugene536:
> Have you checked diff for the site?

yeap, no diff :)

DrDet (approved):

PR #72, opened 2020-12-14
Use PrimitiveType with bit filed.

AlexK0:
Hi. 
I've replaced uint8_t with PrimitiveType for TypeData, so it can be well printed in debugger.


eugene536 (approved):

unserialize (approved):

unserialize:
You can also correct the misspelling 'filed' both in a commit message and in PR name :)

PR #78, opened 2020-12-18
Update docs 12.2020

AlexK0:
Hi!

1) Add MacOS build instructions.
2) Add performance phpdoc annotations.


docs/kphp-internals/developing-and-extending-kphp/compiling-kphp-from-sources.md
```diff
@@ -25,23 +26,75 @@ KPHP is compiled with CMake and packed with CPack.
...
+cd epoll-shim
+git checkout osx-platform
+echo 'export "EPOLL_SHIM_REPO=$(pwd)" >> ~/.bash_profile
 ```
 
-It's also recommended having a `libfmt >= 7` installed (if not, pass `-DDOWNLOAD_MISSING_LIBRARIES=On` to CMake). 
+##### Other Linux
+Make sure you are using the same package list. You may use system default libcurl package, it would work, but without DNS resolving. `uber-h3` must be installed from sources.
+
+### Recommendations
```
quasilyte, line 89: I believe most headings `#-things` have a newline after them. I think keeping the consistent style is a nice thing even for doc markdown sources.

unserialize (approved):

PR #80, opened 2020-12-18
Serialize double as float

eugene536:
The new annotation for fields of serializable classes is implemented in the PR.
It should be used wisely in cases where you are 100% sure what you are doing.

Classes that are annotated as `@kphp-serializable` could use the tag `@kphp-serialized-float32` to let KPHP know which fields contain float value inside and will be serialized as 32-bit floats.
Sacrificing precision 4 bytes per float of a payload can be saved.

### DIFF
--- a/compiler/pipes/check-classes.cpp
+++ b/compiler/pipes/check-classes.cpp
@@ -79,25 +79,33 @@ void CheckClassesPass::check_serialized_fields(ClassPtr klass) {
   fill_reserved_serialization_tags(used_serialization_tags_for_fields, klass);
 
   klass->members.for_each([&](ClassMemberInstanceField &f) {
-    if (auto kphp_serialized_field_str = phpdoc_find_tag_as_string(f.phpdoc_str, php_doc_tag::kphp_serialized_field)) {
-      if (!vk::string_view(*kphp_serialized_field_str).starts_with("none")) {
-        kphp_error_return(klass->is_serializable, fmt_format("you may not use @kphp-serialized-field inside non-serializable klass: {}", klass->name));
-        try {
-          size_t processed_pos{0};
-          auto kphp_serialized_field = std::stoi(*kphp_serialized_field_str, &processed_pos);
-          if (processed_pos != kphp_serialized_field_str->size() && vk::none_of_equal((*kphp_serialized_field_str)[processed_pos], ' ', '/', '#', '\n')) {
-            throw std::invalid_argument("");
-          }
-          kphp_error_return(0 <= kphp_serialized_field && kphp_serialized_field < max_serialization_tag_value, fmt_format("kphp-serialized-field({}) must be >=0 and < than {}", kphp_serialized_field, max_serialization_tag_value + 0));
-          kphp_error_return(!used_serialization_tags_for_fields[kphp_serialized_field], fmt_format("kphp-serialized-field: {} is already in use", kphp_serialized_field));
-          f.serialization_tag = kphp_serialized_field;
-          used_serialization_tags_for_fields[kphp_serialized_field] = true;
-        } catch (std::logic_error &) {
-          kphp_error_return(false, fmt_format("bad kphp-serialized-field: '{}'", *kphp_serialized_field_str));
-        }
+    auto kphp_serialized_field_str = phpdoc_find_tag_as_string(f.phpdoc_str, php_doc_tag::kphp_serialized_field);
+    if (!kphp_serialized_field_str) {
+      kphp_error(!klass->is_serializable, fmt_format("kphp-serialized-field is required for field: {}", f.local_name()));
+      return;
+    }
+
+    kphp_error_return(klass->is_serializable, fmt_format("you may not use @kphp-serialized-field inside non-serializable klass: {}", klass->name));
+    if (vk::string_view(*kphp_serialized_field_str).starts_with("none")) {
+        return;
+    }
+
+    try {
+      size_t processed_pos{0};
+      auto kphp_serialized_field = std::stoi(*kphp_serialized_field_str, &processed_pos);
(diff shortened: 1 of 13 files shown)


runtime/msgpack-serialization.h
```diff
@@ -14,6 +14,57 @@
...
+
+template<class StreamT, class ...Args>
+inline void pack_value(msgpack::packer<StreamT> &packer, const std::tuple<Args...>& value) {
+  packer.pack_array(sizeof...(Args));
+  PackValueHelper<sizeof...(Args)>::convert(packer, value);
+}
+
+template<class StreamT, class T>
+inline void pack_value(bool as_float32, msgpack::packer<StreamT> &packer, const T& value) {
+  if (as_float32) {
```
AlexK0, line 59: 1) It seems to me, that we have to reset this value in advance, somewhere at packing start, or remove this checking and always set `serialize_as_float32() = as_float32` , so we can be sure, that it would be false by default (in case of SIGALRM right after setting true).

2) How it would work for instances inside instances? If I put float32 tag under another instance, would it pack all floats from it as float32?

Look at this example
```php
<?php


/**
 * @kphp-serializable
 **/
class B {
    /**
     * @kphp-serialized-field 3
     * @kphp-serialized-float32
     * @var double
     */
    public $int_f = 536;
}

/**
 * @kphp-serializable
 **/
class A {
    /**
     * @kphp-serialized-field 3
     * @kphp-serialized-float32
     * @var B
     */
    public $b = null;
}

function run() {
    instance_serialize(new A());
}
run();
```

eugene536: Thanks for the comment, unfortunately, I forgot about signals and fixed up the changes.
`f$instance_serialize` resets the boolean flag for float32 serialization. 

For the second question about instances, the annotation forces everything inside to be encoded as `float32`, you can look at the test case `phpt/msgpack_serialize/023_compound_float32.php`

AlexK0: Sorry for bad example(
Please, take a look again
example 1
```php
/** @kphp-serializable **/
class FloatHolder {
    /**
     * @kphp-serialized-field 1
     * @var float[]
     */
    public $float32_f = [20.123, .536, .536];

    /**
     * @kphp-serialized-field 2
     * @var tuple(float, int)
     */
    public $tup;

    public function __construct() {
        $this->tup = tuple(.536, 20);
    }
}

/** @kphp-serializable **/
class A {
    /**
     * @kphp-serialized-field 1
     * @kphp-serialized-float32
     * @var FloatHolder
     */
    public $float_f = null;

    public function __construct() {
        $this->float_f = new FloatHolder();
    }
}
```

example 2 
```php
/** @kphp-serializable **/
class FloatHolder {
    /**
     * @kphp-serialized-field 1
     * @kphp-serialized-float32
     * @var float[]
     */
    public $float32_f = [20.123, .536, .536];

    /**
     * @kphp-serialized-field 2
     * @var tuple(float, int)
     */
    public $tup;

    public function __construct() {
        $this->tup = tuple(.536, 20);
    }
}

/** @kphp-serializable **/
class A {
    /**
     * @kphp-serialized-field 1
     * @kphp-serialized-float32
     * @var FloatHolder
     */
    public $float_f = null;

    public function __construct() {
        $this->float_f = new FloatHolder();
    }
}
```
in the first example `A->float_f->tup` would be serialized as float, in the second example - not.
From my point of view it is confusing

eugene536: Wow, thanks, you are definitely right, it works well on PHP, but fails on kphp.
I've fixed the issue in the new commit (I'll fixup it later).

compiler/pipes/check-classes.cpp
```diff
@@ -79,30 +79,40 @@ void CheckClassesPass::check_serialized_fields(ClassPtr klass) {
...
-        }
+    auto kphp_serialized_field_str = phpdoc_find_tag_as_string(f.phpdoc_str, php_doc_tag::kphp_serialized_field);
+    if (!kphp_serialized_field_str) {
+      kphp_error(!klass->is_serializable, fmt_format("kphp-serialized-field is required for field: {}", f.local_name()));
+      return;
+    }
+
+    kphp_error_return(klass->is_serializable, fmt_format("you may not use @kphp-serialized-field inside non-serializable klass: {}", klass->name));
+    if (vk::string_view(*kphp_serialized_field_str).starts_with("none")) {
+        return;
```
unserialize, line 90: identation

compiler/pipes/check-classes.cpp
```diff
@@ -79,30 +79,40 @@ void CheckClassesPass::check_serialized_fields(ClassPtr klass) {
...
+      used_serialization_tags_for_fields[kphp_serialized_field] = true;
+    } catch (std::logic_error &) {
+      kphp_error_return(false, fmt_format("bad kphp-serialized-field: '{}'", *kphp_serialized_field_str));
     }
   });
 
   klass->members.for_each([&](ClassMemberStaticField &f) {
-    kphp_error_return(!phpdoc_tag_exists(f.phpdoc_str, php_doc_tag::kphp_serialized_field),
+    kphp_error_return(!phpdoc_tag_exists(f.phpdoc_str, php_doc_tag::kphp_serialized_field) &&
+                      !phpdoc_tag_exists(f.phpdoc_str, php_doc_tag::kphp_serialized_float32),
```
unserialize, line 115: Why is && here? @kphp-serialized-field must always exist

unserialize (approved):

compiler/pipes/check-classes.cpp
eugene536: It's about `ClassMemberStaticField` we check here that static fields don't contain `kphp_serialized_field` and `serialized_float32` annotations.

compiler/pipes/check-classes.cpp
eugene536: fixed

AlexK0 (approved):

PR #86, opened 2020-12-25
implement strtotime() via timelib to make it identical to PHP strtotime()

quasilyte:
```
PHP uses timelib more intensively and we could replace some of out
lookup tables from datetime.cpp with its help, but it might be a better
idea to do this (if ever) later.

This patch also includes a fix to date_default_timezone_get(),
so if anyone uses a date_default_timezone_set() with a supported
argument, calling date_default_timezone_get() would return that value.
```
Timelib: https://github.com/derickr/timelib
For this WIP I used this fork (cmake is added): https://github.com/quasilyte/timelib/tree/kphp
The fork has 0 changes except for the added CMake file.

TODO:
- [x] Create a `timelib-kphp-vk` package for it (like `curl-kphp-vk`)
- [x] Move that fork to vkcom organization
- [x] Can we do a library lookup better instead of using a hardcoded paths?
- [x] Update the "build from sources" docs
- [x] Make sure it works on macos

Benchmark results:
```
name             old time/op  new time/op   delta
Now              1.62µs ± 1%   5.33µs ± 0%  +229.20%  (p=0.016 n=5+4)
Timestamp        2.20µs ± 1%  20.37µs ± 0%  +825.27%  (p=0.008 n=5+5)
10September2000  2.10µs ± 1%   6.86µs ± 1%  +225.76%  (p=0.008 n=5+5)
NextMonday       1.83µs ± 1%   7.04µs ± 1%  +284.92%  (p=0.008 n=5+5)
TimeWithOffset   7.32µs ± 1%   8.59µs ± 0%   +17.41%  (p=0.008 n=5+5)
[Geo mean]       2.51µs        8.53µs       +239.11%
```
Benchmarks:
```php
<?php

class TimelibBenchmark {
    public function benchmarkNow() {
        return strtotime("now");
    }

    public function benchmarkTimestamp() {
        return strtotime("91248374");
    }

    public function benchmark10September2000() {
        return strtotime("10 September 2000");
    }

    public function benchmarkNextMonday() {
        return strtotime("last Monday");
    }

    public function benchmarkTimeWithOffset() {
        return strtotime("+1 week 2 days 4 hours");
    }
}

```


cmake/external-libraries.cmake
```diff
@@ -41,6 +41,20 @@ if(KPHP_TESTS)
...
+find_library(KPHP_TIMELIB kphp-timelib)
+if(KPHP_TIMELIB)
+    add_library(kphp-timelib STATIC IMPORTED ${KPHP_TIMELIB})
+else()
+    handle_missing_library("kphp-timelib")
+    FetchContent_Declare(kphp-timelib GIT_REPOSITORY https://github.com/VKCOM/timelib)
+    message(STATUS "---------------------")
+    FetchContent_MakeAvailable(kphp-timelib)
+    include_directories(${kphp-timelib_SOURCE_DIR}/include)
+    target_compile_options(kphp-timelib PUBLIC -Wno-float-conversion -std=gnu11)
```
AlexK0, line 53: Do we need this? Can we put this options directly to kphp-timelib CMakeList.txt?

quasilyte: I moved these flags to kphp-timelib repository and it seems to be working.

runtime/timelib_wrapper.cpp
```diff
@@ -0,0 +1,95 @@
...
+
+  now->tz_info = tzi;
+  now->zone_type = TIMELIB_ZONETYPE_ID;
+  timelib_unixtime2local(now, static_cast<timelib_sll>(preset_ts));
+  timelib_error_container *error = nullptr;
+  timelib_time *t = timelib_strtotime(times.c_str(), times.size(), &error, timelib_builtin_db(), timelib_parse_tzfile);
+  auto t_deleter = vk::finally([t]() { timelib_time_dtor(t); });
+
+  int errors_num = error->error_count;
+  timelib_error_container_dtor(error);
```
AlexK0, line 81: What about warnings? Shouldn't we propagate warnings to php_warning? Can you check PHP behavior, please.

runtime/timelib_wrapper.cpp
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  bool use_heap_memory = (dl::get_script_memory_stats().memory_limit == 0);
+  auto malloc_replacement_guard = make_malloc_replacement_with_script_allocator(!use_heap_memory);
+
+  timelib_time *now = timelib_time_ctor();
+  auto now_deleter = vk::finally([now]() { timelib_time_dtor(now); });
+
+  // timelib destructors do timelib_free which is a macro that expands to free();
+  // so we probably don't need a CriticalSectionGuard here
```
AlexK0, line 71: timelib_free call is not the main issue. The main issue is an invalid internal library state. If you sure, that timelib doesn't use any static/global vars inside (which can be spoiled with signal interruption), then we don't need a CriticalSectionGuard here, otherwise we must use the guard.

And if you add the guard, the second question is: Is timelib ready for nullptrs from malloc family calls?

runtime/timelib_wrapper.cpp
quasilyte: strtotime in PHP never gives any warnings.
It returns a `false` result instead.
image

cmake/external-libraries.cmake
```diff
@@ -41,6 +41,19 @@ if(KPHP_TESTS)
     endif()
 endif()
 
+find_library(KPHP_TIMELIB kphp-timelib)
```
eugene536, line 44: It would be better to add `kphp-timelib` to the variable  `CPACK_DEBIAN_KPHP_PACKAGE_DEPENDS`, so the deb package will depend on the library

runtime/timelib_wrapper.h
```diff
@@ -0,0 +1,14 @@
+#pragma once
+
+#include "kphp_core.h"
+
+// php_timelib wraps the https://github.com/derickr/timelib library
+// which is used in PHP to implement several datetime lib functions.
+
+namespace php_timelib {
```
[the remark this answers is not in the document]
quasilyte: Fixed.

runtime/datetime.cpp
```diff
@@ -829,12 +584,12 @@ mixed f$hrtime(bool as_number) {
...
 
-
 void init_datetime_lib() {
   dl::enter_critical_section();//OK
 
-  setenv("TZ", "Etc/GMT-3", 1);
-  tzset();
+  set_default_timezone_id("Europe/Moscow");
+
+  php_timelib::init();
```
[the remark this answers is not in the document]
quasilyte: Seems legit.

runtime/timelib_wrapper.cpp
```diff
@@ -0,0 +1,95 @@
...
+namespace {
+timelib_tzinfo *get_timezone_info(const char *tz_name) {
+  int dummy_error_code = 0;
+  timelib_tzinfo *info = timelib_parse_tzfile(tz_name, timelib_builtin_db(), &dummy_error_code);
+  return info;
+}
+} // namespace
+
+struct TzinfoCache : private vk::not_copyable {
+  timelib_tzinfo *etc_gmt3_;
```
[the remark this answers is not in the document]
quasilyte: Done.

runtime/datetime.cpp
```diff
@@ -372,18 +364,24 @@ string f$date(const string &format, int64_t timestamp) {
 }
 
 bool f$date_default_timezone_set(const string &s) {
-  if (s != string("Etc/GMT-3", 9) && s != string("Europe/Moscow", 13)) {//TODO
-    if (s == string("Etc/GMT-4", 9)) {
-      php_warning("Timezone Etc/GMT-4 is not supported, use Etc/GMT-3 instead");
-      return true;
-    }
-    php_critical_error ("unsupported default timezone \"%s\"", s.c_str());
+  if (s == string("Etc/GMT-3")) {
```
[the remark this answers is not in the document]
quasilyte: Done. I was discouraged from doing it as other parts of this file had 0 strcmp usages (I added 1 though).

runtime/timelib_wrapper.cpp
quasilyte: I updated the comment.

- timelib can't handle nullptr from malloc gracefully
- timelib is thread-safe

So it looks like CriticalSectionGuard is not needed here.

cmake/external-libraries.cmake
quasilyte: Done.

quasilyte:
I addressed the review comments, but the CI build failed.
I'll remove the WIP prefix when it will be clear that failure was unrelated (or fixed).

runtime/datetime.cpp
```diff
@@ -11,17 +11,29 @@
...
 
+static const char *default_timezone_id = "";
+
+// this function should only be called with supported timezone_id values;
+// it could be "Etc/GMT-3" or "Europe/Moscow"
+static void set_default_timezone_id(const char *timezone_id) {
+  if (strcmp(default_timezone_id, timezone_id) == 0) {
+    return;
+  }
+  setenv("TZ", timezone_id, 1);
```
AlexK0, line 26: Please, consider using the critical guard section here. I'm not sure, if it is safe to interrupt `setenv()` and `tzset()` calls.

runtime/datetime.cpp
```diff
@@ -372,18 +364,24 @@ string f$date(const string &format, int64_t timestamp) {
...
-    if (s == string("Etc/GMT-4", 9)) {
-      php_warning("Timezone Etc/GMT-4 is not supported, use Etc/GMT-3 instead");
-      return true;
-    }
-    php_critical_error ("unsupported default timezone \"%s\"", s.c_str());
+  if (strcmp(s.c_str(), "Etc/GMT-3") == 0) {
+    set_default_timezone_id("Etc/GMT-3");
+    return true;
+  }
+  if (strcmp(s.c_str(), "Europe/Moscow") == 0) {
```
AlexK0, line 371: nit: Copy-paste of the magic strings makes me cry.
"Europe/Moscow" - 5 times
"Etc/GMT-3" - 5 times
"Etc/GMT-4" - 2 times

May be it's worth to move them into const char* global constants into one place? 
E.g. declare in timelib_wrapper.h, define and use in timelib_wrapper.cpp, and use in datetime.cpp.

quasilyte: Done.

runtime/datetime.cpp
```diff
@@ -372,18 +367,24 @@ string f$date(const string &format, int64_t timestamp) {
...
+    set_default_timezone_id(PHP_TIMELIB_TZ_GMT3);
+    return true;
+  }
+  if (strcmp(s.c_str(), PHP_TIMELIB_TZ_MOSCOW) == 0) {
+    set_default_timezone_id(PHP_TIMELIB_TZ_MOSCOW);
+    return true;
+  }
+  // TODO: this branch is weird; remove it?
+  if (strcmp(s.c_str(), PHP_TIMELIB_TZ_GMT4) == 0) {
+    php_warning("Timezone Etc/GMT-4 is not supported, use Etc/GMT-3 instead");
```
AlexK0, line 380: may be
`php_warning("Timezone %s is not supported, use %s instead", PHP_TIMELIB_TZ_GMT4, PHP_TIMELIB_TZ_GMT3);`
?

quasilyte: Done.

AlexK0 (approved):

PR #87, opened 2020-12-28
Load cpp dest dir in a separate thread

AlexK0:
1) Move cpp destination directory initialization logic into separate pipe.
2) Move CollectForkableTypes into separate file and pipe.


compiler/pipes/collect-forkable-types.h
```diff
@@ -0,0 +1,41 @@
...
+// Copyright (c) 2020 LLC «V Kontakte»
+// Distributed under the GPL v3 License, see LICENSE.notice.txt
+
+#pragma once
+
+#include "common/mixin/not_copyable.h"
+#include "common/smart_ptrs/singleton.h"
+#include "compiler/function-pass.h"
+
+class CollectForkableTypes final : public FunctionPassBase {
```
unserialize, line 11: CollectForkableTypesPass?

compiler/pipes/collect-forkable-types.h
```diff
@@ -0,0 +1,41 @@
...
+  VertexPtr on_enter_vertex(VertexPtr root) final;
+
+  void on_finish() final;
+
+private:
+  std::vector<const TypeData *> waitable_types_;
+  std::vector<const TypeData *> forkable_types_;
+};
+
+class ForkableTypes : vk::not_copyable {
```
unserialize, line 26: Why not to store these fields in CollectForkableTypesPass? 
For example, name waitable_types_ above to waitable_types_current_, and waitable_types_ as waitable_types_total? 
(It seems, that it's quite unintuitive, that a separate class represents the total result of a pass)

compiler/pipes/init-cpp-dest-dir.cpp
```diff
@@ -0,0 +1,24 @@
...
+
+#include "compiler/pipes/init-cpp-dest-dir.h"
+
+#include "compiler/compiler-core.h"
+
+void InitCppDestDir::execute(FunctionPtr function, DataStream<FunctionPtr> &os) {
+  os << function;
+
+  static std::atomic_flag dest_dir_initiated = ATOMIC_FLAG_INIT;
+  if (!dest_dir_initiated.test_and_set()) {
```
unserialize, line 15: Can you add a comment describing the purpose? (to make in understandable for a random reader)

compiler/pipes/init-cpp-dest-dir.h
```diff
@@ -0,0 +1,15 @@
+// Compiler for PHP (aka KPHP)
+// Copyright (c) 2020 LLC «V Kontakte»
+// Distributed under the GPL v3 License, see LICENSE.notice.txt
+
+#pragma once
+
+#include "compiler/data/data_ptr.h"
+#include "compiler/threading/data-stream.h"
+
+class InitCppDestDir {
```
unserialize, line 10: InitCppDestDirF? (according to other pipes naming rules)

compiler/compiler.cpp
```diff
@@ -248,6 +250,7 @@ bool compiler_execute(CompilerSettings *settings) {
     >> PipeC<CalcEmptyFunctions>{}
     >> PassC<CalcActualCallsEdgesPass>{}
     >> SyncC<FilterOnlyActuallyUsedFunctionsF>{}
+    >> PipeC<InitCppDestDir>{}
```
unserialize, line 253: Is it true, that loading mtime will start here and run in parallel until the next sync pipe (CFGBeginSync) waits for it?
I mean, that if a hard disk is slow, the execution point won't pass CFGBeginSync (it will hang there waiting for mtime)?

compiler/pipes/collect-forkable-types.h
AlexK0: CollectForkableTypesPass objects have limited lifetime, so if I want to store elements there I have to use static fields, so I've chosen singleton object.

compiler/compiler.cpp
AlexK0: yes

compiler/pipes/collect-forkable-types.h
unserialize: Yes, sure, I meant of course to use static fields. Do you just prefer a singleton object over using static fields in a pass, or there is another reason?

AlexK0: Yes, I prefer singleton, so I can split these pipe and storage logic, noting more.

compiler/compiler.cpp
```diff
@@ -212,14 +212,16 @@ bool compiler_execute(CompilerSettings *settings) {
...
+  if (scheduler_threads == 1) {
     scheduler = new OneThreadScheduler();
   } else {
     auto s = new Scheduler();
-    s->set_threads_count(static_cast<int32_t>(G->settings().threads_count.get()));
+    s->set_threads_count(scheduler_threads);
     scheduler = s;
   }
 
+  vk::singleton<CppDestDirAsyncInitializer>::get().start(scheduler_threads + 1);
```
unserialize, line 224: When there is only 1 thread — for local development — let's do in synchronously in the same thread? For easier, one-thread debugging

DrDet (approved):

unserialize (approved):

compiler/compiler.cpp
unserialize: (outdated)

PR #88, opened 2020-12-28
implement hrtime() PHP function

quasilyte:
Not all PHP installations has hrtime enabled, so I had
to add a polyfill version of that function in tests.

### DIFF
--- a/functions.txt
+++ b/functions.txt
@@ -299,6 +299,7 @@ function mktime ($h ::: int = INT_MIN, $m ::: int = INT_MIN, $s ::: int = INT_MI
 function strftime ($format ::: string, $timestamp ::: int = INT_MIN) ::: string;
 function strtotime ($time ::: string, $timestamp ::: int = INT_MIN) ::: int | false;
 function time() ::: int;
+function hrtime (bool $as_number = false): mixed; // array|int
 
 function debug_backtrace() ::: string[][];
 function posix_getpid() ::: int;
--- a/runtime/datetime.cpp
+++ b/runtime/datetime.cpp
@@ -7,6 +7,7 @@
 #include <clocale>
 #include <ctime>
 #include <sys/time.h>
+#include <chrono>
 
 #include "runtime/critical_section.h"
 #include "runtime/string_functions.h"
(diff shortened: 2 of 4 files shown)


runtime/hrtime.cpp
```diff
@@ -0,0 +1,67 @@
...
+void init_hrtime() {
+#if defined(__APPLE__)
+  hrtime_available = (mach_timebase_info(&_timerlib_info) == KERN_SUCCESS);
+#elif defined(_POSIX_MONOTONIC_CLOCK)
+  hrtime_available = true;
+#endif
+}
+
+int64_t hrtime() {
+#if defined(__APPLE__)
```
AlexK0, line 33: You can use clock_gettime for apple as well.

runtime/hrtime.cpp
```diff
@@ -0,0 +1,67 @@
...
+    return false;
+  }
+
+  if (as_number) {
+    // since we don't support 32-bit build anymore,
+    // "as number" means as int64 and never float/double
+    return nanosecs;
+  }
+
+  array<mixed> result(array_size(2, 0, true));
```
AlexK0, line 63: I guess, you can use array::create.

runtime/hrtime.cpp
quasilyte: Done.

runtime/hrtime.cpp
quasilyte: Done.

functions.txt
```diff
@@ -299,6 +299,7 @@ function mktime ($h ::: int = INT_MIN, $m ::: int = INT_MIN, $s ::: int = INT_MI
 function strftime ($format ::: string, $timestamp ::: int = INT_MIN) ::: string;
 function strtotime ($time ::: string, $timestamp ::: int = INT_MIN) ::: int | false;
 function time() ::: int;
+function hrtime (bool $as_number = false): mixed; // array|int|false
```
AlexK0, line 302: Now it doesn't return false.

quasilyte: Fixed.

AlexK0 (approved):

PR #89, opened 2020-12-29
Template serializators

AlexK0:
1) Add template version for `json_encode` functions family 
2) Add template version for `serialize` function
3) minor: check ssh keys before free - not related fix, it is required for fixing eventual sigsegv


runtime/json-functions.cpp
```diff
@@ -0,0 +1,539 @@
...
+    }
+  }
+
+  return false;
+}
+
+} // namespace
+
+mixed f$json_decode(const string &v, bool assoc) noexcept {
+  static_cast<void>(assoc);
```
unserialize, line 527: In php, when assoc = true, an associative array is returned (that's KPHP case). And when assoc = false, stdClass is returned.
We just skip this argument and work as if it's true always.
Maybe, if it's passed false, invoke php_warning?

runtime/json-functions.h
```diff
@@ -0,0 +1,146 @@
...
+      if (arr.get_next_key() == arr.count()) {
+        is_vector = true;
+      } else {
+        php_warning("Corner case in json conversion, [] could be easy transformed to {}");
+      }
+    }
+  }
+  is_vector &= !force_object;
+
+  static_SB << "{["[is_vector];
```
unserialize, line 62: 😮

unserialize (dismissed):

runtime/json-functions.cpp
AlexK0: It was warning before, and it had been commented, I don't know why
image

unserialize: Seems strange. Hopefully, there aren't many warnings on a working site (since PHP behaves completely differently). I'd suggest to uncomment it, what do you think?

AlexK0: Added comment about it.

AlexK0: False is default value here, I think it may produce a huge number of warnings.

unserialize (approved):

PR #92, opened 2021-01-13
Filter special chars from json logs.

AlexK0:
This patch allows to remove special chars from json logs.


quasilyte (dismissed):
Maybe add a test for non-space iscntrl character, so we have `?` in the output?
Otherwise LGTM.

unserialize (approved):

quasilyte (approved):

PR #94, opened 2021-01-15
Show pretty errors on type mismatch

unserialize:
# This PR greatly improves compilation errors on type mismatch

Take this simple php code:
```
class A {
    public $i = 0;
}

function f(int $arg) {
    $a = new A;
    $a->i = $arg;
}

$arr = [];
$arr[] = 'str';
$v = $arr[0];
f($v);
```

The error here is passing `string $v` to `$arg of f()`, declared as `@param int`.

# How did type mismatch errors look before this PR?

At first, all type inferring process was finished, and then really inferred types were checked against phpdoc.

PHPDocs and type hints were considered as _hints_: if `int $arg`, then `$arg` deemed to be _at least_ int. When passed string `$v`, `$arg` became int+string = mixed. This mixed was assigned to `A::$i` leading to another type mismatch as a consequence.

As a result, having only one actual error here, two were printed. And it's a very simple example! In real code, an occasional error usually led to lots of type mismatches as a propagating chain reaction.

It looked like this:
image

# How do type mismatch errors look now?

Like this:

image

Now we see what is really going on, no more, no less. 
We see an error *pass string to argument $arg of f, but it's declared as @param int* and a stacktrace explaining "Why is it string, but not int?". This stack trace also contains lines of code, so you even don't have an IDE to understand the reason.

Pretty errors are detected and printed in various patters: returning a mismatched value, assigning to a field, etc. — without spoiling types and a mismatch chain reaction.

# How does it work internally?

The main purpose was not to spoil types of variables that have some phpdoc / type hint.
Thus, having `int $arg` it shouldn't be mixed with a string (giving mixed): it should remain int.
The main point is adding `type_restriction` to `VarNode`. Variables, arguments and return values (which are VarNodes, they and only they can have phpdoc) can now be restricted with `TypeData`.
On recalculation (see `var-node.php, do_recalc()`), `new_type_` is calculated only using edges that don't trigger mismatch. If any edge mismatches, then it is stored to be printed after, and does not take part in inferring anymore.

To achieve this, we have to convert phpdoc type_rule to `TypeData` (to set it as `type_restriction`). Now it's implemented as a separate inferring launch on demand, but later this will become more native (after type hints refactoring).

An important note. A PHPDoc rule does not always define a full type. It can contain `any`, it's often just `array`, which is actually `any[]`. It's also a valid type restriction, and moreover: we need to infer that `any`, but we don't allow nothing but array to be passed. Here is an example:

```
/** @var array $a */
$a = [1,2,3];  // becomes int[]
$a = ['str'];  // becomes mixed[]
$a = 3; // remains mixed[], a type mismatch is printed: assign int to array $a
```

# Optimization: create less set edges

If a phpdoc defines a full type (that doesn't contain `any` inside), we can avoid creating edges. See:
```
function f(int $x) {}

f($a);
f($b);
```
Since these two `f()` invocations **can't change the type of $x**, we don't have to create set edges `$x -> $a` and `$x -> $b`. Whenever `$a` and `$b` are recalculated, they don't trigger `$x` recalculations. What we have to, is just to check that they match int afterwards. 
So, we create two `RestrictionMatchPhpdoc` and zero edges.
Same for `return` statement and assignments to fields: when they have type restrictions without `any` inside, no edges are created. This greatly reduces amount of edge dependencies in a well-typed code and speeds up type inferring.

# Moreover

This PR contains some more enhancements or refactorings:
* type hints can be used in functions.txt, they would be checked like in regular functions
* less and greater type rules were deleted: in functions.txt, you either use type hints or ::: cast params
* RestrictionLess was replaced with RestrictionMatchPhpdoc
* `tp_Unknown` and `tp_CreateAny` are deleted
* PrepareFunctionF pipe was removed, now there is a ParseAndApplyPhpdocF pipe that handles phpdocs and phpdoc inheritance for both functions and classes, and does nothing more
* simplified assumptions: dropped ptype, or_null, or_false and get_type_data


AlexK0:
image

compiler/inferring/type-data.cpp
```diff
@@ -759,10 +738,6 @@ int type_strlen(const TypeData *type) {
       return STRLEN_FUTURE_QUEUE;
     case tp_Error:
       return STRLEN_ERROR;
-    case tp_Any:
-      return STRLEN_ANY;
-    case tp_CreateAny:
-      return STRLEN_CREATE_ANY;
```
AlexK0, line 765: Let's completely remove `STRLEN_ANY` and `STRLEN_CREATE_ANY` from code.

compiler/inferring/type-data.cpp
```diff
@@ -799,22 +775,18 @@ bool are_equal_types(const TypeData *type1, const TypeData *type2) {
...
-  const PrimitiveType tp = type1->get_real_ptype();
+
+  const PrimitiveType tp = type1->ptype();
 
   if (tp == tp_Class) {
     return type1->class_types() == type2->class_types();
   }
 
-  if (tp == tp_array) {
+  if (tp == tp_array || tp == tp_future || tp == tp_future_queue) {
```
AlexK0, line 789: `vk::any_of_equal`? And the same below at line 842.

compiler/inferring/restriction-match-phpdoc.cpp
```diff
@@ -0,0 +1,124 @@
...
+}
+
+bool RestrictionMatchPhpdoc::is_restriction_broken() {
+  return !is_less_or_equal_type(actual_node->get_type(), expected_type);
+}
+
+std::string RestrictionMatchPhpdoc::get_description() {
+  std::string desc;
+  std::string actual_type = this->actual_node->get_type()->as_human_readable();
+  std::string expected_type = this->expected_type->as_human_readable();
```
AlexK0, line 38: local `expected_type` hides class `expected_type` let's use another name

compiler/stage.cpp
```diff
@@ -49,14 +46,11 @@ void on_compilation_error(const char *description __attribute__((unused)), const
...
-  stage::print(file);
-  string correct_description;
-  if (!stage::should_be_colored(file)) {
-    correct_description = TermStringFormat::remove_special_symbols(full_description);
-  } else {
-    correct_description = full_description;
-  }
+  std::string error_title = get_assert_level_desc(assert_level);
+  fmt_fprintf(file, "\n~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~\n{} at stage: {}, gen by {}:{}\n",
+              stage::should_be_colored(file) ? TermStringFormat::paint_red(error_title) : error_title, stage::get_name(), file_name, line_number);
```
AlexK0, line 51: I suggest write only file name (without full path) for 'gen by'.
Please, consider to use `kbasename(file_name)` for that.

compiler/pipes/parse-and-apply-phpdoc.cpp
```diff
@@ -361,60 +361,90 @@ class ParseAndApplyPHPDoc {
       }
     }
   }
+};
 
+// Inspects @var above all fields and converts them to type hints
+// Also analyzes @kphp-serializable and other tags above the class itself
+class ParseAndApplyPhpDocForClass {
+  ClassPtr klass;
```
DrDet, line 369: Maybe it would be better to place it below `private:`?

unserialize: For me, it's better when members are above implementation. The same for kbasename

AlexK0 (approved):

quasilyte (approved):

PR #95, opened 2021-01-15
Emit more info about warm up on graceful restart

DrDet:
For tuning warm up parameters we need to know 'hotness' of instance cache at the end of warm up, e.g. when warm up timeout is expired. Here are 2 things to help with this:

- new stat metrics `new IC size` and `old IC size`, where `IC size` is the size of instance cache in corresponding master (new or old) at the end of warm up on graceful restart
- log instance cache sizes during warm up

### DIFF
--- a/server/php-master.cpp
+++ b/server/php-master.cpp
@@ -1536,6 +1536,8 @@ STATS_PROVIDER_TAGGED(kphp_stats, 100, STATS_TAG_KPHP_SERVER) {
   add_histogram_stat_double(stats, "requests.incoming_queries_per_second", qps_calculator.get_incoming_qps());
   add_histogram_stat_double(stats, "requests.outgoing_queries_per_second", qps_calculator.get_outgoing_qps());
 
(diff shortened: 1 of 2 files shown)


server/php-master.cpp
```diff
@@ -1536,6 +1536,8 @@ STATS_PROVIDER_TAGGED(kphp_stats, 100, STATS_TAG_KPHP_SERVER) {
   add_histogram_stat_double(stats, "requests.incoming_queries_per_second", qps_calculator.get_incoming_qps());
   add_histogram_stat_double(stats, "requests.outgoing_queries_per_second", qps_calculator.get_outgoing_qps());
 
+  add_histogram_stat_double(stats, "graceful_restart.warmup.final_instance_cache_sizes_ratio", WarmUpContext::get().calc_final_instance_cache_sizes_ratio());
```
AlexK0, line 1539: Let's write both values instead of ratio. so we can catch the case if `final_old_instance_cache_size`  is small.

DrDet: done

AlexK0 (approved):

PR #102, opened 2021-01-25
Type inferring performance/memory optimizations

unserialize:
This PR has several optimizations reducing CPU/memory footprint during transpilation.

Metrics for VK.com code:
Before:
```
memory.rss: 14322671616
memory.rss_peak: 14661017600
```
After:
```
memory.rss: 13196865536
memory.rss_peak: 13501394944
```
Pipes "type inferring" and "collect main edges" became noticably faster, memory consumption became less everywhere, because sizeof(TypeData) and sizeof(tinf::Node) were greatly reduced.
No codegen diff.

* use std::forward_list instead of std::vector for tinf edges
* drop field tinf::Node::recalc_cnt
* drop TypeData::generation
* drop TypeData::parent_
* drop TypeData::anykey_value

These commits are supposed to be rebased, not squashed.

### DIFF
--- a/compiler/inferring/node-recalc.cpp
+++ b/compiler/inferring/node-recalc.cpp
@@ -100,10 +100,6 @@ void NodeRecalc::set_lca_at(const MultiKey *key, const RValue &rvalue) {
     key = &MultiKey::any_key(0);
   }
 
-  if (type->error_flag()) {
-    return;
-  }
-
   if (types_stack.empty()) {
     new_type_->set_lca_at(*key, type, !rvalue.drop_or_false, !rvalue.drop_or_null);
   } else {
--- a/compiler/inferring/node-recalc.cpp
+++ b/compiler/inferring/node-recalc.cpp
@@ -86,7 +86,6 @@ void NodeRecalc::set_lca_at(const MultiKey *key, const RValue &rvalue) {
     if (auto_edge_flag()) {
       add_dependency_impl(node_, rvalue.node, key);
     }
-    __sync_synchronize();
     type = rvalue.node->get_type();
   } else if (rvalue.type != nullptr) {
     type = rvalue.type;
@@ -160,14 +159,11 @@ NodeRecalc::NodeRecalc(tinf::Node *node, tinf::TypeInferer *inferer) :
   inferer_(inferer) {}
 
 void NodeRecalc::on_changed() {
-  __sync_synchronize();
-
   //fmt_fprintf(stderr, "{} : {} -> {}\n", node_->get_description(), type_out(node_->get_type()), type_out(new_type()));
   node_->set_type(new_type_);
   new_type_ = nullptr;
 
-  __sync_synchronize();
-
+  // here we need a lock: in case another thread updates edges_to_this_ just now via auto edge
   AutoLocker<Lockable *> locker(node_);
   for (auto e : node_->get_rev_next()) {
     inferer_->recalc_node(e->from);
--- a/compiler/inferring/node-recalc.cpp
+++ b/compiler/inferring/node-recalc.cpp
@@ -160,6 +160,7 @@ NodeRecalc::NodeRecalc(tinf::Node *node, tinf::TypeInferer *inferer) :
 
 void NodeRecalc::on_changed() {
   //fmt_fprintf(stderr, "{} : {} -> {}\n", node_->get_description(), type_out(node_->get_type()), type_out(new_type()));
+  new_type_->mark_classes_used();
   node_->set_type(new_type_);
   new_type_ = nullptr;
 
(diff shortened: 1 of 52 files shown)


compiler/inferring/type-data.cpp
```diff
@@ -891,30 +856,46 @@ size_t TypeData::get_tuple_max_index() const {
...
+  if (subkeys_values.empty() && before->subkeys_values.empty()) {
+    return false;
+  }
+
+  auto i1 = subkeys_values.begin();
+  auto i2 = before->subkeys_values.begin();
+  auto e1 = subkeys_values.end();
+  auto e2 = before->subkeys_values.end();
+  for (; i1 != e1 && i2 != e2; ++i1, ++i2) {
+    if (i1->second->did_type_data_change_after_tinf_step(i2->second)) {
```
kunyavskiy, line 863: Shouldn't we check key here?

compiler/inferring/node.h
```diff
@@ -30,26 +29,37 @@ class Node : public Lockable {
...
 
-  volatile int recalc_state_;
+protected:
+  enum {
+    recalc_st_waiting = 0,
+    recalc_st_processing = 1,
+    recalc_st_need_relaunch = 2,
+    // and three more states — the same by meaning, but with bit 16 on (16, 17, 18)
+    // that 3 more states — with bit 16 on — mean that recalc has once finished <=> recalc count > 0
+    recalc_bit_16 = 16,
```
kunyavskiy, line 39: I don't like name of this enum value. Maybe recalc_bit_at_least_once?

compiler/inferring/restriction-isset.cpp
```diff
@@ -158,5 +157,12 @@ bool RestrictionIsset::find_dangerous_isset_dfs(int isset_flags, tinf::Node *nod
 
 bool RestrictionIsset::is_restriction_broken() {
   vector<tinf::Node *> bt;
-  return find_dangerous_isset_dfs(a_->isset_flags, a_, &bt);
+  int isset_flags = a_->isset_flags;
+  kphp_assert(isset_flags);
+  // before dfs, Node::isset_flags contains what ifi mask we need to search for
+  // during dfs, we use the same field to store what mask has already been checked for
+  a_->isset_flags = 0;
```
kunyavskiy, line 164: This change seams to be very strange. You fixed flags in one vertex, but a lot of others, which this one can possibly depend, will contain some trash inside. Why this could work?

compiler/inferring/node-recalc.cpp
```diff
@@ -160,14 +159,11 @@ NodeRecalc::NodeRecalc(tinf::Node *node, tinf::TypeInferer *inferer) :
   inferer_(inferer) {}
 
 void NodeRecalc::on_changed() {
-  __sync_synchronize();
-
   //fmt_fprintf(stderr, "{} : {} -> {}\n", node_->get_description(), type_out(node_->get_type()), type_out(new_type()));
   node_->set_type(new_type_);
   new_type_ = nullptr;
 
-  __sync_synchronize();
```
kunyavskiy, line 169: I'm not sure if this correct, even if it works on tests. As AutoLocker does nothing with std::atomic, happens-before can be not conducted here, and compiler could be allowed to rearrange code leading to something wrong. Probably, this wouldn't happen, but maybe it's better to wait untill normal lockers are used.

compiler/inferring/type-data.cpp
unserialize: No, we shouldn't. Because even if keys were swapped, we should return that this was changed

unserialize: Sounds reasonable. I'll add `i1->first != i2->first || ` condition

compiler/pipes/collect-forkable-types.cpp
```diff
@@ -13,8 +13,8 @@ VertexPtr CollectForkableTypesPass::on_enter_vertex(VertexPtr root) {
       if (call->str_val == "wait") {
         waitable_types_.push_back(tinf::get_type(root));
       } else if (call->str_val == "wait_multi") {
-        forkable_types_.push_back(tinf::get_type(root)->const_read_at(Key::any_key()));
-        waitable_types_.push_back(tinf::get_type(root)->const_read_at(Key::any_key()));
+        forkable_types_.push_back(tinf::get_type(root)->lookup_at_any_key());
```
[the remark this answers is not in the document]
unserialize: wait_multi() accepts `future<any>[]`, so it's definitely an array if it passed type checks

compiler/inferring/node-recalc.cpp
```diff
@@ -160,16 +155,14 @@ NodeRecalc::NodeRecalc(tinf::Node *node, tinf::TypeInferer *inferer) :
...
-
   //fmt_fprintf(stderr, "{} : {} -> {}\n", node_->get_description(), type_out(node_->get_type()), type_out(new_type()));
+  new_type_->mark_classes_used();
   node_->set_type(new_type_);
   new_type_ = nullptr;
 
-  __sync_synchronize();
-
+  // here we need a lock: in case another thread updates rev_next_ just now via audo edge
```
AlexK0, line 163: misspelling: `audo edge`

compiler/inferring/node.h
unserialize: Ok

compiler/inferring/restriction-isset.cpp
unserialize: Lazha :( You're right. BTW, it shows, that we have no tests for such cases. 
I've reverted this commit for now. Probably, I'll think how to deal with isset_flags/isset_was and this restriction later on.

compiler/inferring/node.h
```diff
@@ -23,38 +22,55 @@ class Node : public Lockable {
...
   }
 
-  void add_edge(Edge *edge);
-  void add_rev_edge(Edge *edge);
+  void register_edge_from_this(const tinf::Edge *edge);
+  void register_edge_to_this(const tinf::Edge *edge);
 
-  inline const std::vector<Edge *> &get_next() const {
-    return next_;
+  inline const std::forward_list<const Edge *> &get_edges_from_this() const {
```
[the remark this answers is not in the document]
unserialize: Ok, deleted inline keyword

compiler/inferring/node-recalc.cpp
unserialize: Fixed

compiler/inferring/node-recalc.cpp
unserialize: Finally we decided, that __sync_synchronize() is not needed here

AlexK0 (approved):

PR #104, opened 2021-01-25
Task workers: Step 1 - basic design

DrDet:
**Step 1 in task workers implementation**
Basic design described in `KPHP-1231`:
- network design
- communication protocol via shared memory
- communication with master process
- array x^2 interface for tests
- and a lot of other technical details


server/php-master.cpp
```diff
@@ -825,6 +834,7 @@ int run_worker() {
   worker->logname_id = worker_logname_id;
   worker->last_activity_time = my_now;
 
+  worker->task_result_fd_idx = task_result_fd_idx;
```
AlexK0, line 837: we already have `logname_id` and I guess it does exactly the same stuff as `task_result_fd_idx`. I believe, that we can reuse it somehow, maybe it's worth to rename `logname_id` for better understanding.

server/php-master.cpp
```diff
@@ -1924,6 +1936,8 @@ void run_master() {
 
     me->generation = generation;
 
+    vk::singleton<TaskWorkersContext>::get().master_update_task_workers();
```
AlexK0, line 1939: may be use another name, like `master_run_task_workers`?

server/task-workers/shared-context.cpp
```diff
@@ -0,0 +1,23 @@
...
+
+#include <cassert>
+#include <new>
+#include <sys/mman.h>
+
+#include "server/task-workers/shared-context.h"
+
+SharedContext &SharedContext::get() {
+  assert(inited);
+  static void *shared_mem = mmap(nullptr, sizeof(SharedContext), PROT_READ | PROT_WRITE, MAP_SHARED | MAP_ANONYMOUS, -1, 0);
```
AlexK0, line 13: let's create separate make function.

server/task-workers/shared-context.h
```diff
@@ -0,0 +1,23 @@
...
+
+class SharedContext : vk::not_copyable {
+public:
+  std::atomic<int> task_queue_size{0};
+
+  // must be called once from master process
+  static void init();
+  static SharedContext &get();
+private:
+  static bool inited;
```
AlexK0, line 20: I suggest to remove this flag.

server/task-workers/task-worker-client.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+#include "net/net-reactor.h"
+
+#include "server/php-queries.h"
+#include "server/task-workers/task-worker-client.h"
+#include "server/task-workers/task-workers-context.h"
+#include "server/task-workers/shared-context.h"
+
+int TaskWorkerClient::read_task_results(int fd, void *data __attribute__((unused)), event_t *ev) {
+  static int read_buf[PIPE_BUF / sizeof(int)];
```
AlexK0, line 18: I suggest to create POD structure with named fields, and use it for writing and reading.

server/php-master.cpp
```diff
@@ -363,6 +365,11 @@ struct pipe_info_t {
   connection pending_stat_queue;
 };
 
+enum class WorkerType {
```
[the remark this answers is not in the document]
DrDet: On one hand `worker_info_t` must contain field identifying type of worker which is one out of two: `http_worker` or `task_worker`, it can't be `master`. On the other hand `generic_event_loop` can be called by master process as well, so it should accept such option. So `ProcessType` wouldn't be enough.

server/php-queries.cpp
```diff
@@ -858,6 +859,29 @@ int create_rpc_answer_event(slot_id_t slot_id, int len, net_event_t **res) {
...
+  auto deallocate_shared_memory_slice = vk::finally([&]() { memory_manager.deallocate_slice(task_result_memory_ptr); });
+
+  net_event_t *event = nullptr;
+  int status = alloc_net_event(ready_task_id, net_event_type_t::task_worker_answer, &event);
+  if (status <= 0) {
+    return status;
+  }
+  // script is running here, because is_valid_slot(ready_task_id) == true
+
+  size_t slice_payload_size = vk::singleton<task_workers::SharedMemoryManager>::get().get_slice_payload_size();
```
AlexK0, line 873: use local variable `memory_manager`?

server/task-workers/pipe-io.h
```diff
@@ -0,0 +1,45 @@
...
+// Distributed under the GPL v3 License, see LICENSE.notice.txt
+
+#pragma once
+
+#include <climits>
+#include <cstdint>
+
+namespace task_workers {
+
+class PipeIO {
```
AlexK0, line 12: Let's add protected default destructor.

server/task-workers/shared-memory-manager.h
```diff
@@ -0,0 +1,74 @@
...
+      pid_t old_pid = 0;
+      return owner_pid.compare_exchange_strong(old_pid, getpid());
+    }
+
+    bool release() {
+      pid_t old_pid = owner_pid.exchange(0);
+      return old_pid != 0;
+    }
+
+    bool is_vacant() const {
```
AlexK0, line 57: Le't remove this unused function.

server/task-workers/shared-memory-manager.h
```diff
@@ -0,0 +1,74 @@
...
+    std::atomic<pid_t> owner_pid{0};
+
+    bool acquire() {
+      pid_t old_pid = 0;
+      return owner_pid.compare_exchange_strong(old_pid, getpid());
+    }
+
+    bool release() {
+      pid_t old_pid = owner_pid.exchange(0);
+      return old_pid != 0;
```
AlexK0, line 54: I guess it should be an assertion that `old_pid == getpid()` and release should return nothing (void).

server/task-workers/shared-memory-manager.cpp
```diff
@@ -0,0 +1,85 @@
...
+  memory = static_cast<unsigned char *>(mmap(nullptr, memory_size, PROT_READ | PROT_WRITE, MAP_SHARED | MAP_ANONYMOUS, -1, 0));
+  for (size_t i = 0; i < total_slices_count; ++i) {
+    unsigned char *slice_ptr = get_slice_ptr(i);
+    new (slice_ptr + slice_payload_size) SliceMetaInfo{};
+  }
+}
+
+void *SharedMemoryManager::allocate_slice() {
+  std::random_device rd;
+  std::mt19937 gen(rd());
```
AlexK0, line 32: You need to generate only one random number, so I suggest to use `std::random_device` directly without the mt generator.
Also, please, move `std::random_device` into the class, we don't need to create it on the each call (I'm not sure about the cost of `std::random_device` creation).

server/task-workers/task-worker-server.cpp
```diff
@@ -0,0 +1,165 @@
...
+    if (last_stats.is_started() && last_stats.time() > std::chrono::duration<int>{TaskWorkersContext::MAX_HANGING_TIME_SEC / 2}) {
+      tvkprintf(task_workers, 1, "Too many tasks. Task worker will complete them later. Goes back to event loop now\n");
+      has_delayed_tasks = true;
+      return 0;
+    }
+
+    task_reader.reset();
+    ssize_t read_bytes = task_reader.read_task_from_pipe(TASK_BYTE_SIZE);
+
+    if (read_bytes == -2) {
```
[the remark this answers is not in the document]
DrDet: Done. Sorry for skipping your previous comment about this, it seems I didn't see it for some reason.

unserialize:
You are going to squash these commits into a single, right?

DrDet:
> You are going to squash these commits into a single, right? @DrDet

Yep

functions.txt
```diff
@@ -1327,3 +1327,9 @@ final class UberH3 {
   static public function polyfill($polygon_boundary ::: tuple(float, float)[], $holes ::: tuple(float, float)[][], $resolution ::: int) ::: int[] | false;
   static public function maxPolyfillSize($polygon_boundary ::: tuple(float, float)[], $holes ::: tuple(float, float)[][], $resolution ::: int) ::: int;
 }
+
+/** Task workers interface **/
+
+function async_x2($arr ::: int[]) ::: int;
```
unserialize, line 1333: int[] $arr

functions.txt
```diff
@@ -1327,3 +1327,9 @@ final class UberH3 {
...
+function await_x2($task_id ::: int) ::: int[];
```
unserialize, line 1334: int $task_id

server/task-workers/pipe-io.cpp
```diff
@@ -0,0 +1,78 @@
+#include <cstring>
```
unserialize, line 1: a license header

AlexK0 (dismissed):

AlexK0 (approved):

PR #106, opened 2021-01-27
Get rid of >, >=, !=, <> operations from AST.

AlexK0:
Hi, this patch replaces some compare operations, and removes them from AST.
1) `$a  > $b`  to `$b < $a`
2) `$a  >= $b`  to `$b <= $a`
3) `$a  != $b`  to `!($b == $a)`
4) `$a  <> $b`  to `!($b == $a)`

### DIFF
--- a/compiler/operation.cpp
+++ b/compiler/operation.cpp
@@ -90,7 +90,6 @@ void OpInfo::init_static() {
   curP++;
 
   add_binary_op(curP, tok_eq2, op_eq2);
-  add_binary_op(curP, tok_neq2, op_neq2);
   add_binary_op(curP, tok_eq3, op_eq3);
   add_binary_op(curP, tok_neq3, op_neq3);
   add_binary_op(curP, tok_spaceship, op_spaceship);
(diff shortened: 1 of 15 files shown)


compiler/operation.cpp
```diff
@@ -90,15 +90,13 @@ void OpInfo::init_static() {
...
   add_binary_op(curP, tok_eq2, op_eq2);
-  add_binary_op(curP, tok_neq2, op_neq2);
   add_binary_op(curP, tok_eq3, op_eq3);
   add_binary_op(curP, tok_neq3, op_neq3);
   add_binary_op(curP, tok_spaceship, op_spaceship);
   curP++;
 
   add_binary_op(curP, tok_lt, op_lt);
   add_binary_op(curP, tok_le, op_le);
-  add_binary_op(curP, tok_neq_lg, op_neq2);
```
AlexK0, line 103: It looks like was an issue, because `<>` has same priority as `==` and `!=`
https://www.php.net/manual/en/language.operators.precedence.php
image

quasilyte (approved):
LGTM.

unserialize (approved):

PR #107, opened 2021-01-28
use elements_cached instead of elements_stored in warmup on graceful restart

DrDet:
`elements_stored` is the global number of elements stored in cache since server start, it always increases.
`elements_cached` is the current size of cache.


server/php-master.cpp
```diff
@@ -1947,7 +1947,7 @@ void run_master() {
 
     me->running_workers_n = me_running_workers_n;
     me->dying_workers_n = me_dying_workers_n;
-    me->instance_cache_elements_stored = static_cast<uint32_t>(instance_cache_get_stats().elements_stored.load(std::memory_order_relaxed));
+    me->instance_cache_elements_stored = static_cast<uint32_t>(instance_cache_get_stats().elements_cached.load(std::memory_order_relaxed));
```
AlexK0, line 1950: let's rename instance_cache_elements_stored -> instance_cache_elements_cached as well

DrDet: fixed

AlexK0 (approved):

PR #108, opened 2021-01-28
allow literals and '()'-expressions to be postfix expr operands

quasilyte:
This change allows these expressions to be parsed:

	"foo"[index]         string literal indexing
	[$a, $b][index]      array literal indexing
	("foo")[index]       parentheses expr indexing
	(function() {...})() immediately invoked function expressions
	(new Foo())->f()     parenthesized expr arrow access

All cases above are valid in PHP.

These are not valid in PHP, therefore they are forbidden in KPHP as well:

	($x)++
	($x)--

No codegen diff on vkcom.

### DIFF
--- a/compiler/gentree.cpp
+++ b/compiler/gentree.cpp
@@ -296,7 +296,7 @@ VertexAdaptor<op_string_build> GenTree::get_string_build() {
   return VertexAdaptor<op_string_build>::create(strings).set_location(sb_location);
 }
 
-VertexPtr GenTree::get_postfix_expression(VertexPtr res) {
+VertexPtr GenTree::get_postfix_expression(VertexPtr res, bool parenthesized) {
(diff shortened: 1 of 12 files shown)


unserialize:
"parenthesized expr arrow access" was also parsed correctly before, by the way :)

compiler/gentree.cpp
```diff
@@ -587,6 +592,7 @@ VertexPtr GenTree::get_expr_top(bool was_arrow) {
       break;
     case tok_opbrk:
       res = get_short_array();
+      return_flag = false; // array is valid postfix expression operand
```
unserialize, line 595: "array" in comment seems like copy-paste

unserialize (approved):

AlexK0 (approved):

PR #109, opened 2021-01-28
fix null file ptr crash

quasilyte:
auto_location() returns a location without a file,
restriction-stacktrace-finder.cpp tries to access a file
and dereferences a null pointer in `loc.file->get_line`.

By making auto_location() return a more precise location,
we can fix that null pointer dereference *and* improve
error locations in many places (some parse errors had "unknown file"
in their error messages).


unserialize:
BTW, this PR makes codegeneration unstable (in differs between launches).
(I wanted to port this idea, but stopped after this fact, it needs more investigation)

PR #117, opened 2021-02-04
Get rid of op_neq3.

AlexK0:
Replace `$a !== $b` with `!($a === $b)` in AST.


quasilyte (commented):
Maybe add a test with `!==` which triggers a warning?
It seems like there is none, although I may be wrong.

quasilyte (dismissed):

unserialize (approved):

PR #119, opened 2021-02-08
replacing comparison of array value by key with null on isset function for instance field

troy4eg:
x['key'] !== null to x.isset('key')

### DIFF
--- a/compiler/pipes/preprocess-eq3.cpp
+++ b/compiler/pipes/preprocess-eq3.cpp
@@ -29,10 +29,11 @@ inline VertexPtr PreprocessEq3Pass::convert_eq3_null_to_isset(VertexPtr eq_op, V
   while (v->type() == op_index) {
     v = v.as<op_index>()->array();
   }
-  if (auto var_v = v.try_as<op_var>()) {
+  if (vk::any_of_equal(v->type(), op_var, op_instance_prop, op_array)) {
     // it's a kludge to make "real world" PHP code compile
     // TODO: can we get rid of this?
-    if (var_v->str_val == "connection" || vk::contains(var_v->str_val, "MC")) {
+    if (v->has_get_string() &&
+        (v->get_string() == "connection" || vk::contains(v->get_string(), "MC"))) {
       return eq_op;
     }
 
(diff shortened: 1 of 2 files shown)


compiler/pipes/preprocess-eq3.cpp
```diff
@@ -29,10 +29,10 @@ inline VertexPtr PreprocessEq3Pass::convert_eq3_null_to_isset(VertexPtr eq_op, V
   while (v->type() == op_index) {
     v = v.as<op_index>()->array();
   }
-  if (auto var_v = v.try_as<op_var>()) {
+  if (v->type() == op_var || v->type() == op_instance_prop) {
```
AlexK0, line 32: Please, consider using `vk::any_of_equal`.

compiler/pipes/preprocess-eq3.cpp
```diff
@@ -29,10 +29,11 @@ inline VertexPtr PreprocessEq3Pass::convert_eq3_null_to_isset(VertexPtr eq_op, V
   while (v->type() == op_index) {
     v = v.as<op_index>()->array();
   }
-  if (auto var_v = v.try_as<op_var>()) {
+  if (vk::any_of_equal(v->type(), op_var, op_instance_prop, op_array)) {
```
unserialize, line 32: Seems strange...
op_instance_prop seems strange, because it means "$a->x === null" here
op_array seems strange, because $a['x'] is not op_array — it's op_index (whereas op_array is just [1,2,3])

AlexK0: I guess it's a lack of description, because 
op_instance_prop allows to replace both `$a->x === null`  and `$a->x["key"] === null`
op_array enables a case with a class constant, like `A::SOME_CONST_ARR["key"] === null` or even `["a" => "b"]["a"] === null`
so this preprocess works the same way for variables, instance pros and class constants.

unserialize: Nevertheless, we shouldn't replace "$a->x === null" with isset

unserialize (approved):

PR #124, opened 2021-02-15
User friendly resumable expressions.

AlexK0:
Hi there! 
This patch enables the following resumable call expressions:
1) For ternary operator
2) For && and || operations
3) For lhs of ?? operator

It forbids complex expressions in the lhs part of op_set with the resumable rhs: only non global vars are allowed.

So, the following code examples are available now (imagine that all functions are resumable)

```php
$v = check() ? true_case() : false_case();
$v = check() ?: false_case();
 
$a[] = check() ? true_case() : false_case();
$a[$key] = check() ? true_case() : false_case();

$i->a = check() ? true_case() : false_case();

if (check() ? true_case() : false_case()) {
// ...
}
```

```php
$v = check1() && check2();
$v = check1() || check2();
 
if ((check1() && check2()) || (check3() && check4())) {
// ...
}
```

```php
$v = get() ?? "null";
```

```php
if ((check1() && check2()) ? true_case() : (get() ?? "null")) {
// ...
}
```

The following code examples are forbidden (imagine that all functions are resumable)

```php
$v = get() ?? fallback();

// function_call() is not resumable
$a[function_call()] = check() ? true_case() : false_case();

$a[$global_var] = get() ?? "null";

// get_instance() is not resumable
get_instance()->a = resumable_func_call();

```

### DIFF
--- a/compiler/pipes/extract-resumable-calls.cpp
+++ b/compiler/pipes/extract-resumable-calls.cpp
@@ -52,10 +52,23 @@ VertexPtr *ExtractResumableCallsPass::extract_resumable_expr(VertexPtr vertex) n
     if ((*resumable_expr)->type() == op_func_call && vertex->type() == op_set && set_modify->lhs()->type() != op_instance_prop) {
       return {};
     }
+    auto lhs_var = set_modify->lhs().try_as<op_var>();
+    auto instance_prop = set_modify->lhs().try_as<op_instance_prop>();
+    while (instance_prop) {
+      lhs_var = instance_prop->instance().try_as<op_var>();
+      instance_prop = instance_prop->instance().try_as<op_instance_prop>();
+    }
+    if (!lhs_var || lhs_var->var_id->is_in_global_scope()) {
+      return {};
+    }
   } else if (auto list = vertex.try_as<op_list>()) {
     resumable_expr = &list->array();
   } else if (auto set_value = vertex.try_as<op_set_value>()) {
-    resumable_expr = &set_value->value();
+    if (auto var_key = set_value->key().try_as<op_var>()) {
+      if (!var_key->var_id->is_in_global_scope()) {
+        resumable_expr = &set_value->value();
+      }
+    }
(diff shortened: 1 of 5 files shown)


compiler/pipes/extract-resumable-calls.cpp
```diff
@@ -5,78 +5,160 @@
...
+    case op_log_or:
+      *skip_conv_and_sets(&set_if_lhs_true->rhs()) = temp_var.first;
+      *skip_conv_and_sets(&set_if_lhs_false->rhs()) = operation->rhs();
+      break;
+    case op_log_and:
+      *skip_conv_and_sets(&set_if_lhs_false->rhs()) = temp_var.first;
+      *skip_conv_and_sets(&set_if_lhs_true->rhs()) = operation->rhs();
+      break;
+    default:
+      kphp_assert(!"unexpected type");
```
unserialize, line 126: We usually write kphp_assert(0 && "message")?

compiler/pipes/extract-resumable-calls.cpp
```diff
@@ -5,78 +5,160 @@
...
+  }
 
-  return temp_var;
+  auto check_lhs = VertexAdaptor<op_if>::create(temp_var.second, GenTree::embrace(set_if_lhs_true), GenTree::embrace(set_if_lhs_false));
+  return GenTree::embrace(check_lhs.set_rl_type(val_none).set_location(operation));
+}
+
+VertexPtr ExtractResumableCallsPass::replace_resumable_expr_with_temp_var(VertexPtr *resumable_expr, VertexPtr expr_user) noexcept {
+  auto tem_var = make_temp_resumable_var(*resumable_expr);
```
unserialize, line 134: temp_var

compiler/pipes/extract-resumable-calls.cpp
AlexK0: Interesting :)

unserialize (approved):

DrDet (approved):

PR #130, opened 2021-02-19
improve preg_replace and preg_replace_callback typing

quasilyte:
If `$subject` is a string, the result will be `?string`, not `mixed`.

We achieve this by the same trick as in `str_replace` return type.

The `^3|string|null` will resolve into either `string|null` or `mixed`.
It's `?string` when `^3` is a string; otherwise, any T+string will result in `mixed`.

### DIFF
--- a/runtime/regexp.h
+++ b/runtime/regexp.h
@@ -91,7 +91,7 @@ public:
   Optional<array<mixed>> split(const string &subject, int64_t limit, int64_t flags) const;
 
   template<class T>
-  mixed replace(const T &replace_val, const string &subject, int64_t limit, int64_t &replace_count) const;
+  Optional<string> replace(const T &replace_val, const string &subject, int64_t limit, int64_t &replace_count) const;
 
   static int64_t last_error();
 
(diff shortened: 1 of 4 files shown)


unserialize (approved):

runtime/regexp.h
```diff
@@ -263,18 +266,18 @@ string regexp::get_replacement(const T &replace_val, const string &subject, int6
...
 template<class T>
-mixed regexp::replace(const T &replace_val, const string &subject, int64_t limit, int64_t &replace_count) const {
+Optional<string> regexp::replace(const T &replace_val, const string &subject, int64_t limit, int64_t &replace_count) const {
   pcre_last_error = 0;
   int64_t result_count = 0;//calls can be recursive, can't write to replace_count directly
 
   check_pattern_compilation_warning();
   if (pcre_regexp == nullptr && RE2_regexp == nullptr) {
-    return mixed();
+    return false;
```
AlexK0, line 275: Is it ok?
In functions.txt it is written, that this function returns only `string|null` without `false`:
```
function preg_replace (....) ::: ^3|string|null;
```

The same about other `false` below.

AlexK0: Ok, now I see below that many of the `f $ preg_replace` functions return `false`. It might make sense to fix the signature in functions.txt.

quasilyte: Good observations!
I changed all `return false` to `return {}` to return a proper null there.
Added `002_null_error.php` test to check that this behavior is consistent with PHP.

AlexK0 (approved):

AlexK0 (approved):

AlexK0 (approved):

PR #133, opened 2021-02-20
Fix race condition in task (lease) communication protocol

DrDet:
Here is a try to fix bug with error -2005, "Task invoker is invalid".
It seems that this bug happens, because KPHP treats old lease target incorrect immediately after send `STOP_READY` to it on changing lease target, then gets task from lease target that was sent before `STOP_READY` receiving and rejects it.

To fix this, we will wait with some timeout for an acknowledgement that `STOP_READY` was received, and only after that treat lease target incorrect.

### DIFF
--- a/common/pid.h
+++ b/common/pid.h
@@ -31,4 +31,8 @@ typedef enum pid_match pid_match_t;
 
 pid_match_t matches_pid (const struct process_id *X, const struct process_id *Y);
 
+inline pid_match_t matches_pid_ref (const process_id_t &X, const process_id_t &Y) {
+  return matches_pid(&X, &Y);
+}
+
 #endif
--- a/common/pid.h
+++ b/common/pid.h
@@ -31,8 +31,4 @@ typedef enum pid_match pid_match_t;
 
 pid_match_t matches_pid (const struct process_id *X, const struct process_id *Y);
 
-inline pid_match_t matches_pid_ref (const process_id_t &X, const process_id_t &Y) {
-  return matches_pid(&X, &Y);
-}
-
 #endif
--- a/server/lease-context.h
+++ b/server/lease-context.h
@@ -1,0 +1,21 @@
+// Compiler for PHP (aka KPHP)
+// Copyright (c) 2020 LLC «V Kontakte»
+// Distributed under the GPL v3 License, see LICENSE.notice.txt
+
+#pragma once
+
+#include "common/mixin/not_copyable.h"
+#include "common/smart_ptrs/singleton.h"
+#include "common/kphp-tasks-lease/lease-worker-mode.h"
+
+
+class LeaseContext : vk::not_copyable {
+public:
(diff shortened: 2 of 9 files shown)


AlexK0 (approved):

AlexK0 (approved):

common/pid.h
```diff
@@ -31,4 +31,8 @@ typedef enum pid_match pid_match_t;
 
 pid_match_t matches_pid (const struct process_id *X, const struct process_id *Y);
 
+inline pid_match_t matches_pid_ref (const process_id_t &X, const process_id_t &Y) {
+  return matches_pid(&X, &Y);
+}
+
```
Tindarid, line 37: unnecessary?

server/php-lease.cpp
```diff
@@ -4,46 +4,74 @@
...
-  lst_start,          // if !has_pending_scripts -> change state to lst_on, wait for connection to target ready, do lease_set_ready() && run_rpc_lease();
-  lst_on,             // connecting to target, send RPC_READY, doing work, if too much time doing work -> change state to lst_finish
-  lst_finish          // wait finishing current task, send RPC_STOP to target, send TL_KPHP_LEASE_STATS to rpc-proxy, change state to lst_off
+const char *lease_state_to_str[] = {
+  "off",
+  "start",
+  "on",
+  "initiating_finish",
+  "finish",
 };
```
Tindarid, line 69: Maybe,
```
const char *lease_state_to_str[] = {
  [static_cast<int>(lease_state_t::off)] = "off",
  ...
};
```
is better

server/php-lease.cpp
```diff
@@ -4,46 +4,74 @@
...
 
-static lease_state_t lease_state = lst_off;
-static int lease_ready_flag = 0; // waiting for something -> equal 0
+lease_state_t lease_state = lease_state_t::off;
+int lease_ready_flag = 0; // waiting for something -> equal 0
```
Tindarid, line 72: `bool lease_ready`?

common/pid.h
DrDet: yep, removed

server/php-lease.cpp
DrDet: done

server/lease-context.h
```diff
@@ -0,0 +1,18 @@
...
+#pragma once
+
+#include "common/mixin/not_copyable.h"
+#include "common/smart_ptrs/singleton.h"
+#include "common/kphp-tasks-lease/lease-worker-mode.h"
+
+
+class LeaseContext : vk::not_copyable {
+public:
+  vk::optional<QueueTypesLeaseWorkerMode> cur_lease_mode;
```
AlexK0, line 14: pls, private default constructor for singleton.

DrDet: done

AlexK0 (approved):

PR #135, opened 2021-02-24
optimize cfg, uninited and smart casts

unserialize:
# This PR reduces the "CFGBeginF" memory consumption

image

All commits modify cfg.cpp or something nearby. 
I dropped "create_full_cfg" completely. Instead, cfg for all vertex types are manually written.
Speed up dsu calculation for usages.
Reduced size of VertexPtr by dropping the `id` field.
Split smart casts and uninited vars searching, as uninited dfs can be done much faster.
Dropped generalization of IdGen and IdMap: use in-place memory management instead (they were used in cfg only).

All site diff is connected to better type detection after smart casts — because of create_full_cfg() removal.
For instance, a ternary f($a ? f2($a) : 0) — f2() is now called as check_not_false(check_not_null($a)). Before this PR, it was just $a.

Reviewing is supposed to be done commit-by-commit.
All these commits will be squashed into a single one. You can see detailed history in the PR.


quasilyte:
I want this merged soon. :sweat_smile:
`create_full_cfg` causes some troubles for me as well

tests/phpt/smart_casts/05_inside_operators.php
```diff
@@ -0,0 +1,66 @@
+@ok
+<?php
+require_once 'kphp_tester_include.php';
+
+// these tests didn't fork until create_full_cfg was used
```
quasilyte, line 5: I'm not sure I understand what this comment means. :)
Does it have a typo?

fork -> work?
until -> when?

```
these tests didn't work when create_full_cfg was used
```

compiler/pipes/cfg.cpp
```diff
@@ -138,23 +154,39 @@ class CFG {
   }
 };
 
-Node CFG::new_node() {
-  Node res;
-  node_gen.init_id(&res);
+inline Node CFG::new_node() {
+  Node res(n_nodes++);
+
+  if ((n_nodes & 0x3f) == 1) {
```
DrDet, line 160: Can we make it a bit more understandable here? 
```suggestion
  if ((n_nodes & ((1 << 6) - 1)) == 1) {
```
Maybe it worth to extract constant with the mask value

compiler/pipes/cfg.cpp
```diff
@@ -138,23 +154,39 @@ class CFG {
...
+    reserve_capacity_for_cfg_idmaps();
+  }
+
   return res;
 }
 
+void CFG::reserve_capacity_for_cfg_idmaps() {
+  int capacity = (n_nodes | 0x3f) + 1;
```
DrDet, line 168: and here

compiler/pipes/cfg.cpp
unserialize: 0x3f is really much understandable for everybody who knows hex notation, imho :)

DrDet: I think the option with `<<` highlights that it's the mask for power of two, rather than just some hex value
But OK, if you're not agree, just resolve this)

AlexK0: more understandable is: `if (n_nodes % 64 == 1)`

AlexK0 (approved):

DrDet (approved):

PR #136, opened 2021-02-24
optimize bad_vars calculation performed in a single thread

unserialize:
# This PR speeds up "CalcBadVars on_finish" single-thread step

image

"bad vars" for a function are used in check-ub.cpp.
They are modified non-primitive globals, that can lead to a potential ub when modified and accessed in subcalls.

Added comments about optimization points and some tests.

### DIFF
--- a/compiler/pipes/calc-bad-vars.cpp
+++ b/compiler/pipes/calc-bad-vars.cpp
@@ -65,6 +65,7 @@ public:
     int vertex_n = (int)vertices.size();
     IdMap<int> was(vertex_n);
     vector<VertexT> topsorted;
+    topsorted.reserve(vertex_n);
     for (const auto &vertex : vertices) {
       dfs(vertex, rev_graph, &was, &topsorted);
     }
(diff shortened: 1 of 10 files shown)


compiler/pipes/calc-func-dep.cpp
```diff
@@ -76,30 +78,33 @@ VertexPtr CalcFuncDepPass::on_enter_vertex(VertexPtr vertex) {
...
       }
     }
   } else if (auto ptr = vertex.try_as<op_func_ptr>()) {
     data.dep.push_back(ptr->func_id);
   } else if (auto var_vertex = vertex.try_as<op_var>()) {
     VarPtr var = var_vertex->var_id;
-    if (var->is_in_global_scope()) {
-      data.used_global_vars.push_back(var);
+    // here we add var for later ub check; we store only vars that can potentially lead to ub, see check-ub.cpp
+    if ((var->is_global_var() || var->is_class_static_var()) &&   // only globals (not even static inside functions)
```
AlexK0, line 96: May it's worth to use `var->is_in_global_scope()` here?

compiler/pipes/calc-func-dep.cpp
```diff
@@ -49,7 +49,9 @@ VertexPtr CalcFuncDepPass::on_enter_vertex(VertexPtr vertex) {
   //NB: There is no user functions in default values of any kind.
   if (auto call = vertex.try_as<op_func_call>()) {
     FunctionPtr other_function = call->func_id;
-    data.dep.push_back(other_function);
+    if (!other_function->is_extern() || other_function->is_resumable) {
```
AlexK0, line 52: Please, add a comment why we skip resumable functions.

compiler/pipes/calc-bad-vars.cpp
```diff
@@ -118,22 +119,80 @@ struct FuncCallGraph {
...
+          largest_e = e->bad_vars;
+        }
+      }
+      if (largest_e) {
+        bool all_in_largest_e = true;
+        for (FunctionPtr f : component) {
+          for (VarPtr v_in_self : modified_global_vars[f]) {
+            if (largest_e->find(v_in_self) == largest_e->end()) {
+              all_in_largest_e = false;
+              break;
```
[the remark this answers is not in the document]
unserialize: You're right, but line 165 has a prevention from starting an inner loop. And the loop in line 157 is likely to loop over an empty array. All in all, this doesn't cause any overhead (I've checked).
I really don't want to extract these part of lines into a separate function, as I even can't give it a name. They are just a part of this function body, no more

compiler/pipes/calc-func-dep.cpp
unserialize: No, as we don't have to deal with function static vars

AlexK0: But this logic is applied for static vars in recursion calls as well.

compiler/pipes/calc-bad-vars.cpp
```diff
@@ -118,22 +119,80 @@ struct FuncCallGraph {
...
+      // optimization 1: lots of functions don't modify globals at all
+      // then we leave f->bad_vars nullptr
+      bool empty = true;
       for (FunctionPtr f : component) {
-        bad_vars_uniq.insert(tmp_vars[f].begin(), tmp_vars[f].end());
+        empty &= modified_global_vars[f].empty();
+      }
+      for (FunctionPtr f : edges) {
+        empty &= f->bad_vars == nullptr;
+      }
```
eugene536, line 141: You can shorten the code above, on top of that it wouldn't iterate over edges if not empty `modified_global_vars` has been found
```
bool empty = vk::all_of(component, [this](auto f) { return modified_global_vars[f].empty(); })
          && vk::all_of(edges, [](auto f) { return f->bad_vars == nullptr; });
```

compiler/pipes/calc-bad-vars.cpp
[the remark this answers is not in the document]
eugene536: and the loop above over `component` is not necessary if we hit the break.

I suggest you two simplifications for this nested code
```
bool all_in_largest_e = true;
for (FunctionPtr f : component) {
  all_in_largest_e &= vk::any_of(modified_global_vars[f], [&](VarPtr v_in_self) { return vk::contains(*largest_e, v_in_self); });
  if (!all_in_largest_e) {
    break;
  }
}
```

second one
```
bool all_in_largest_e = vk::any_of(component, [&](FunctionPtr f) {
    return vk::any_of(modified_global_vars[f], [&](VarPtr v_in_self) { return vk::contains(*largest_e, v_in_self); });
  });
```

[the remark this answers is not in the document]
eugene536: and the loop above over `component` is not necessary if we hit the break.

I suggest you two simplifications for this nested code

```
for (FunctionPtr f : component) {
  all_in_largest_e &= vk::all_of(modified_global_vars[f], [&](VarPtr v_in_self) { return vk::contains(*largest_e, v_in_self); });
  if (!all_in_largest_e) {
    break;
  }
}
```
Second
```
bool all_in_largest_e = vk::all_of(component, [&](FunctionPtr f) {
  return vk::all_of(modified_global_vars[f], [&](VarPtr v_in_self) { return vk::contains(*largest_e, v_in_self); });
});
```

compiler/pipes/calc-bad-vars.cpp
```diff
@@ -118,22 +119,80 @@ struct FuncCallGraph {
...
+          for (VarPtr v_in_self : modified_global_vars[f]) {
+            if (largest_e->find(v_in_self) == largest_e->end()) {
+              all_in_largest_e = false;
+              break;
+            }
+          }
+        }
+        for (FunctionPtr e : edges) {
+          if (e->bad_vars != largest_e && e->bad_vars != nullptr && all_in_largest_e) {
+            for (VarPtr v_in_e : *e->bad_vars) {
```
eugene536, line 166: here we can shorten too
```
all_in_largest_e &= vk::all_of(*e->bad_vars, [&](VarPtr v_in_e) { return !vk::contains(*largest_e, v_in_e); });
if (!all_in_largest_e) {
  break;
}
```

unserialize: +

compiler/pipes/calc-bad-vars.cpp
unserialize: ok

compiler/pipes/calc-bad-vars.cpp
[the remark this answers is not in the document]
unserialize: ok

compiler/pipes/calc-func-dep.cpp
unserialize: Not exactly. Other functions can't have direct references to static var of current function. Recursion calls are supposed to be safe, as the context is known. `modified_global_vars` that we collect here are used only to find potential misbehavior of global references across different functions. BTW, including static vars in this analysis doesn't affect vkcom codegen at all (no SAFE_OP added), it only heavies execution.

compiler/pipes/calc-func-dep.cpp
unserialize: Added a large comment above

compiler/pipes/calc-bad-vars.cpp
```diff
@@ -118,22 +119,66 @@ struct FuncCallGraph {
...
+          largest_e = e->bad_vars;
+        }
       }
+      if (largest_e) {
+        bool all_in_largest_e = vk::all_of(component, [&](FunctionPtr f) {
+          return vk::all_of(modified_global_vars[f], [&](VarPtr v_in_self) {
+            return largest_e->find(v_in_self) != largest_e->end();
+          });
+        });
+        all_in_largest_e &= vk::all_of(edges, [&](FunctionPtr e) {
```
quasilyte, line 155: I'm not 100% against joining bool results with `&`, but when it leads to another edges traversal I'm more inclined that it's not the most beautiful thing.

It's up to you whether it should be changed or not.

quasilyte (dismissed):

compiler/pipes/calc-bad-vars.cpp
unserialize: How'd you suggest to correct this not very beautiful code? ) I tried to use
```
bool all_in_largest_e = vk::all_of(
  ...
) && vk::all_of( 
  ...
);
```
but it seems even less beautiful than the final way.

DrDet (approved):

PR #137, opened 2021-02-24
optimize CodeGenF::on_finish() performed in a single thread

unserialize:
# This PR speeds up "CodeGen on_finish" single-thread step

image

Dropped, or paralleled, or removed some computations that were previously done in a single thread:
* Don't get a vector of functions and sort it
* Extract preparing functions to execute() to be run in parallel.
* Async launch of tl-schema generation and other previously sync tasks.

### DIFF
--- a/compiler/pipes/code-gen.cpp
+++ b/compiler/pipes/code-gen.cpp
@@ -32,6 +32,15 @@ size_t CodeGenF::calc_count_of_parts(size_t cnt_global_vars) {
   return 1u + cnt_global_vars / G->settings().globals_split_count.get();
 }
 
+
+void CodeGenF::execute(FunctionPtr function, DataStream<WriterData> &unused_os __attribute__ ((unused))) {
+  if (FunctionData::does_need_codegen(function)) {
+    prepare_generate_function(function);
+    G->stats.on_function_processed(function);
+    tmp_stream << function;
+  }
+}
+
 void CodeGenF::on_finish(DataStream<WriterData> &os) {
   vk::singleton<CppDestDirInitializer>::get().wait();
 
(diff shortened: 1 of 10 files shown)


compiler/pipes/code-gen.cpp
```diff
@@ -154,34 +132,12 @@ void CodeGenF::prepare_generate_function(FunctionPtr func) {
     ? func->file_id->owner_lib->headers_dir() + func->header_name
     : func->subdir + "/" + func->header_name;
 
-  my_unique(&func->static_var_ids);
-  my_unique(&func->global_var_ids);
-  my_unique(&func->local_var_ids);
+  std::sort(func->static_var_ids.begin(), func->static_var_ids.end());
```
AlexK0, line 135: Why did you remove unique?

unserialize: Because these vars are already unique. I mean, that kphp_assert(count_before == count_after) is true always.

compiler/pipes/code-gen.h
```diff
@@ -14,15 +14,14 @@
 class CodeGenerator;
 
 class CodeGenF final : public SyncPipeF<FunctionPtr, WriterData> {
-  using need_profiler = std::false_type;
+  using need_profiler = std::true_type;
```
DrDet, line 17: Why was it disabled before?

unserialize: Because it had no logic for each function, it had only `on_finish`. That's why, it was disabled, as it always showed zero measurements.
Now, I've implemented `CodeGenF::execute` that works in parallel.

DrDet (approved):

PR #138, opened 2021-02-24
speedup code gen, calc hashes on the fly

unserialize:
# This PR speeds up comparing codegen diff with a previous launch

image

When KPHP is run, it codegenerates all functions/classes cpp code, but rewrites only files that were changed since a previous run.
Typically, when you write code, there are very few changes relatively to the entire codebase amount.

Before this PR, all classes/functions were codegenerated, hashes of cpp code were calculated and compared with an indexfile (from a previous run).
```
>> SyncC<CodeGenF>{}             // produces cpp/h code
>> PipeC<WriteFilesF, false>{};  // calculates hashes and re-writes changed files
```
Now, the algorithm works like this: at first step, just calculate hashes on the fly; the second step is launched only for commands that triggered hash changes:
```
>> SyncC<CodeGenF>{}              // create all codegen commands and launch them in "just calc hashes" mode
>> PipeC<CodeGenForDiffF>{}       // re-launch codegen commands that diff from the previous kphp launch
>> PipeC<WriteFilesF, false>{};   // store files that differ from the previous kphp launch
```

Pay attention to the second column of the performance image: for vkcom, codegeneration now typically works ~2 second faster than before, also taking ~1G less memory.

One of disadvantages of calculating a hash on the fly is that 
```
W << "while ("
// vs
W << "while" << " ("
```
will lead to different hashes. So, refactoring of codegen that doesn't lead to real cpp diff can trigger files rewriting. Personally I consider this not as a bery bad case, I don't think, that it would happen often and will be annoying.

These commits will be rebased, not squashed.


compiler/code-gen/files/vars-cpp.h
```diff
@@ -6,13 +6,23 @@
...
 };
+
+struct VarsCppPart : CodeGenRootCmd {
+  VarsCppPart(std::vector<VarPtr> &&vars_of_part, size_t part_id);
+  void compile(CodeGenerator &W) const final;
+
+private:
+  mutable std::vector<VarPtr> vars_of_part_;
+  size_t part_id;
+};
```
AlexK0, line 28: new line

compiler/compiler-core.cpp
```diff
@@ -494,6 +494,14 @@ const Index &CompilerCore::get_index() {
 }
 
 File *CompilerCore::get_file_info(std::string &&file_name) {
+  File *f = cpp_index.get_file(file_name);
```
AlexK0, line 497: Here a race condition with `insert_file` call is possible.

compiler/compiler-core.cpp
```diff
@@ -494,6 +494,14 @@ const Index &CompilerCore::get_index() {
...
+  if (f != nullptr) {
+    return f;
+  }
+//  printf("%s not found in index, creating\n", file_name.c_str());
+
+  static std::mutex mutex;
```
AlexK0, line 503: nit: from my point of view it would be much better if this mutex was placed as CompilerCore field.

compiler/code-gen/code-generator.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+    bool cur_hashes_diff = cur_file->crc64 != hash_of_cpp || cur_file->crc64_with_comments != hash_of_comments;
+    if (cur_hashes_diff) {
+      diff_files_count++;
+      // in this case, the same codegen command will be passed to the next pipe, see CodeGenForDiffF
+    }
+    cur_file = nullptr;
+
+  } else {
+
```
AlexK0, line 60: empty line

common/algorithms/hashes.h
```diff
@@ -56,7 +56,7 @@ class range_hasher : Hasher {
 };
 
 template<class T>
-size_t std_hash(const T &obj) {
+inline size_t std_hash(const T &obj) {
```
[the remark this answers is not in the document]
unserialize: Ok, removed

compiler/code-gen/files/vars-cpp.h
unserialize: ok

compiler/compiler-core.cpp
unserialize: Outdated, this function was refactored

compiler/code-gen/code-generator.cpp
unserialize: fixed

compiler/compiler-core.cpp
unserialize: Well... It's funny, that this race condition existed earlier (see WriteFilesF in the current master). But somehow, we didn't encounter it in a real life. Why? I don't know. Maybe, its probability is too small (e.g. the number of newly-appearing files is negligibly small compared to the number of files from a previous launch). Maybe, it was just a luck.

I couldn't find the way to fix this simply. That's why I have written a complicated fix, see the last commit. 
Its main idea is the following:
* don't keep a single map<string, File*> files:
* instead, have two maps: files_prev_launch and files_only_cur_launch
* the first map is accessed in a single thread (when reading from dir / indexfile)
* the second is accessed from multiple threads, that's why it's guarded with a mutex
Please, consider the last commit and comments within it.

compiler/code-gen/code-generator.cpp
```diff
@@ -0,0 +1,84 @@
...
+  cur_file->needed = true;
+  // also, cur_file->includes will be modified on demand, see add_include()
+
+  if (is_step_just_calc_hashes) {
+
+  } else {
+    kphp_assert(data == nullptr);
+    data = new WriterData(cur_file, compile_with_crc);
+    data->begin_line();
+  }
```
i582, line 31: Can rewrite this condition with negation to get rid of the empty body?

unserialize: It's intentionally. See the same if/else below (in the opposite function). This emphasizes, that on a file open, if we are calculating hashes, we do nothing.

unserialize: But yeah, it's a good point. I'll add a comment inside this empty if

compiler/index.cpp
```diff
@@ -195,36 +199,69 @@ void Index::fix_path(std::string &path) const {
...
-      f->subdir = vk::string_view{};
-    }
-    f->ext = f->name.substr(dot_i);
-    f->name_without_ext = f->name.substr(0, dot_i);
+
+  File *f = get_file(path);
+  if (f != nullptr) {
+    return f;
   }
+//  printf("%s not found in index, creating\n", file_name.c_str());
```
i582, line 225: Commented out code. Is it needed here?

compiler/pipes/write-files.cpp
```diff
@@ -70,6 +33,10 @@ void WriteFilesF::execute(WriterData data, EmptyStream &) {
   }
 
   if (need_fix) {
+    string code_str;
+    data->dump(code_str);
+//    printf("overwrite file %s need_fix=%d need_del=%d need_save_time=%d\n", file->path.c_str(), need_fix, need_del, need_save_time);
```
i582, line 38: Commented out code. Is it needed here?

compiler/index.cpp
unserialize: Just for potential logging/debugging in the future. We have some commented printfs throughout all our code. While not being the best practice, they point out, that this exact place is somehow important for logging.

compiler/pipes/write-files.cpp
unserialize: Same as above

compiler/code-gen/code-generator.cpp
i582: Understand. It will look good with a comment.

compiler/index.cpp
i582: Thanks for the explanation, I will remember it for the future.

AlexK0 (approved):

PR #139, opened 2021-02-26
support php 7.4 class type hints

unserialize:
Now we can write simple types near class fields instead of `@var`:
```
class A {
  public int $field = 0;
  public ?self $me = null;
```
Both for instance and static fields.

KPHP can parse any syntaxically correct type hints (int[] and string|false for example), but PHP supports only some primitives and nullable (and union since php 8).


AlexK0 (approved):

quasilyte (approved):

quasilyte:
Let's merge it?

PR #141, opened 2021-02-26
Fix warning in array_rand when num less then array size

proggga:
Php throw warning, if you pass in array_rand NUM that greater then array size.
Kphp does not throw it. But throws only when array size. == 0.
Message looks like: Parameter num of array_rand must be positive.
That happens cause
if num > size: we make num equal to size
so if we have num 5, size == 0, then we have num == 0
and that errors triggers.
I've made fix for right alert.

### DIFF
--- a/runtime/array_functions.h
+++ b/runtime/array_functions.h
@@ -924,11 +924,9 @@ mixed f$array_rand(const array<T> &a, int64_t num) {
   }
 
   int64_t size = a.count();
-  if (num > size) {
-    num = size;
-  }
-  if (unlikely(num <= 0)) {
-    php_warning("Parameter num of array_rand must be positive");
+
+  if (unlikely(num <= 0 || num > size)) {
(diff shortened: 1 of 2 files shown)


runtime/array_functions.h
```diff
@@ -924,14 +924,17 @@ mixed f$array_rand(const array<T> &a, int64_t num) {
...
-  if (num > size) {
-    num = size;
-  }
+
   if (unlikely(num <= 0)) {
     php_warning("Parameter num of array_rand must be positive");
     return {};
   }
 
+  if (num > size) {
```
AlexK0, line 933: May be make same waning as in PHP?
`Second argument has to be between 1 and the number of elements in the array`

Also, please add some tests.

proggga: Fixed, added tests

quasilyte (approved):

PR #143, opened 2021-03-01
make `strict_types=1` disable automatic cast insertion for a current file

quasilyte:
In PHP, strict_types=1 would break the code at the run time if given
argument types mismatch expected types (expressed via type hints).

Previously, in KPHP it would silently cast these types and we'll
get inconsistent behavior between PHP and KPHP.

This change makes KPHP code a little bit closer to PHP,
but instead of failing at the run time we enable more type checking
for cast params in `strict_types=1` files.

A lot of existing PHP code is not prepared for this, so
this behavior is activated only if `declare()` has a phpdoc
with `@kphp-strict-types-enabled`.

As in PHP, strict_types is checked at the call site.

### DIFF
--- a/compiler/pipes/collect-main-edges.cpp
+++ b/compiler/pipes/collect-main-edges.cpp
@@ -283,6 +283,7 @@ void CollectMainEdgesPass::on_func_call(VertexAdaptor<op_func_call> call) {
   for (int i = 0; i < call->args().size(); ++i) {
     VertexPtr arg = call->args()[i];
     auto param = function_params[i].as<op_func_param>();
+    bool is_cast_param = !stage::get_function()->file_id->is_strict_types && param->is_cast_param;
 
     // call an extern function having a callback type description, like 'callable(^1[*]) : bool'
     if (function->is_extern() && param->type_hint && param->type_hint->try_as<TypeHintCallable>()) {
@@ -298,11 +299,11 @@ void CollectMainEdgesPass::on_func_call(VertexAdaptor<op_func_call> call) {
     // * the paragraph above about 'any' refers only to PHP functions; built-ins accepting 'any' need only a postponed check
     // * cast params go their own way
     bool should_create_set_to_infer = (!param->type_hint || param->type_hint->has_tp_any_inside())
-                                      && !function->is_extern() && !param->is_cast_param;
+                                      && !function->is_extern() && !is_cast_param;
 
     if (should_create_set_to_infer) {
       create_set(as_lvalue(function, i), arg);
-    } else if (param->type_hint && param->type_hint->is_typedata_constexpr() && !param->is_cast_param) {
+    } else if (param->type_hint && param->type_hint->is_typedata_constexpr() && !is_cast_param) {
       create_postponed_type_check(as_rvalue(function, i), as_rvalue(arg), param->type_hint);
     }
 
--- a/compiler/pipes/preprocess-function.cpp
+++ b/compiler/pipes/preprocess-function.cpp
@@ -396,7 +396,10 @@ private:
         kphp_error_act(call_arg->type() != op_varg,
                        fmt_format("Passed unpacked arguments to a non-vararg function"), continue);
 
-        if (param->is_cast_param) {   // ::: cast param in functions.txt or @kphp-infer cast function
+        // for cast params (::: in functions.txt or '@kphp-infer cast') we add
+        // conversions automatically (implicit casts), unless the file from where
+        // we're calling this function is annotated with strict_types=1
+        if (!current_function->file_id->is_strict_types && param->is_cast_param) {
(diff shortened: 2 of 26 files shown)


compiler/pipes/collect-main-edges.cpp
```diff
@@ -283,6 +283,7 @@ void CollectMainEdgesPass::on_func_call(VertexAdaptor<op_func_call> call) {
   for (int i = 0; i < call->args().size(); ++i) {
     VertexPtr arg = call->args()[i];
     auto param = function_params[i].as<op_func_param>();
+    bool is_cast_param = !stage::get_function()->file_id->is_strict_types && param->is_cast_param;
```
unserialize, line 286: Please, avoid using `stage::get_function()`, as it's not clear, what function we are talking about.
Here we should use `function` (a variable above), I suppose

compiler/pipes/preprocess-function.cpp
```diff
@@ -396,7 +396,10 @@ class PreprocessFunctionPass final : public FunctionPassBase {
         kphp_error_act(call_arg->type() != op_varg,
                        fmt_format("Passed unpacked arguments to a non-vararg function"), continue);
 
-        if (param->is_cast_param) {   // ::: cast param in functions.txt or @kphp-infer cast function
+        // for cast params (::: in functions.txt or '@kphp-infer cast') we add
+        // conversions automatically (implicit casts), unless the file from where
+        // we're calling this function is annotated with strict_types=1
+        if (!stage::get_function()->file_id->is_strict_types && param->is_cast_param) {
```
unserialize, line 402: The same about get_function()

tests/phpt/declare/01_strict_types.phpt
```diff
@@ -0,0 +1,4 @@
+<?php
```
unserialize, line 1: Is it okay, that some files have the '.phpt' extension and don't have the '@ok' tag?

quasilyte: I use `php` with + ok/fail tag for everything if it's possible.
`declare` can't be combined with tags as it will lead to a parse error in both KPHP and PHP.

compiler/pipes/collect-main-edges.cpp
quasilyte: Maybe I need to access current file directly, without going through the function.

If we would use `function` variable above, we would make `strict_types` checked based on the declarations, but we want to apply it on the caller side. So if `f()` is declared in strict file, but called from the non-strict file, there should be no error. In that case, `f` is a called `function`, so it's file is a declaration file.

AlexK0: May be `current_function` would be better in this case? 🤔

tests/phpt/declare/01_strict_types.phpt
AlexK0: Files with `.phpt` extension implicitly have `@ok` tag - this is a hack for zend tests..
But for me, It's very confusing: some cases use `.php.inc`, but others `.phpt`, and from my opinion we shouldn't abuse this hack for our tests. I suggest remove all `.phpt` tests from `tests/phpt/` dir and use one style with `.php.inc`.

compiler/pipes/collect-main-edges.cpp
quasilyte: > May be current_function would be better in this case? thinking

It is. But all methods of CollectMainEdgesPass are static. So in order to access it, we need to make them non-static, presumably.

quasilyte: I used `current_function` in `preprocess-function.cpp` though.

compiler/pipes/preprocess-function.cpp
quasilyte: Replaced to `current_function`.

tests/phpt/declare/01_strict_types.phpt
quasilyte: Replaced `phpt` with `@ok` + require.

tests/phpt/declare/07_strict_types_kphpinfercast_error.php
```diff
@@ -0,0 +1,8 @@
+@kphp_should_fail
+/pass string to argument \$s of infercast_int_param/
+/but it's declared as @param int/
+<?php
+
+require_once __DIR__ . '/infercast_func.php';
+require_once __DIR__ . '/can_call_with_cast.php'; // OK
+require_once __DIR__ . '/cant_call_with_cast.php'; // Error
```
unserialize, line 8: Shouldn't these two files have a '.inc' extenstion?

quasilyte: Done. I added `inc` to all php files that don't have `@` tag and are intended to be included from other test files (that have `@` tag).

unserialize (approved):

PR #144, opened 2021-03-02
make strip_tags accept array as $allow, remove implicit cast from substr_replace

quasilyte:
This change makes `string_tags()` compatible with PHP7.4 version of that function.
$allow can accept array of allowed tags.

Also fixed an issue where self-closing tags were not ignored.
They are ignored since PHP5.3.4, so using them is a logical error.
Instead of `<br/>`, users should use `<br>` which will match both
`<br>` and `<br/>` tags.

For `substr_replace()` function, implicit casts for string arguments
are removed to avoid unwanted array->string conversions.
Result type changed from `Optional<string>` to just `string`;
the corner-case behavior is adjusted to be identical to what PHP does.

This code is an example of natural `substr_replace()` usage:
```php
$s = substr_replace($s, $rep, $offset);
```
If `substr_replace()` returns optional, `$s` stops being string here.

We may add support for the array arguments in the future, but for now it's
good enough to reject suspicious code that will behave differently
in KPHP and PHP7.4.

The problem with `substr_replace()` array signatures is that
there are multiple scalar vs array combinations, not all of them
are valid; even PHP does not check for all possible errors
(for example, when `$length` and `$offset` arrays have different sizes).

### DIFF
--- a/functions.txt
+++ b/functions.txt
@@ -611,7 +611,7 @@ function ord ($c ::: string) ::: int;
 function strcasecmp ($str1 ::: string, $str2 ::: string) ::: int;
 function strcmp ($str1 ::: string, $str2 ::: string) ::: int;
 function stripslashes ($str ::: string) ::: string;
-function strip_tags ($str ::: string, $allow ::: string = "") ::: string;
+function strip_tags ($str ::: string, $allow ::: string|string[] = "") ::: string;
 function strncmp ($str1 ::: string, $str2 ::: string, $len ::: int) ::: int;
 function strnatcmp ($str1 ::: string, $str2 ::: string) ::: int;
 function wordwrap ($str ::: string, $width ::: int = 75, $break ::: string = '\n', $cut ::: bool = false) ::: string;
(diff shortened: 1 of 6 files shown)


functions.txt
```diff
@@ -611,7 +611,7 @@ function ord ($c ::: string) ::: int;
 function strcasecmp ($str1 ::: string, $str2 ::: string) ::: int;
 function strcmp ($str1 ::: string, $str2 ::: string) ::: int;
 function stripslashes ($str ::: string) ::: string;
-function strip_tags ($str ::: string, $allow ::: string = "") ::: string;
+function strip_tags ($str ::: string, $allow ::: string|string[] = "") ::: string;
```
AlexK0, line 614: Please, consider the following example

```php
<?php

function demo($s, $t) {
  strip_tags($s, $t);
}

demo("x", ["y"]);
demo("x", "z");
```

If you compile it, it will produce C++ compilation error:
```
error: no matching function for call to ‘f$strip_tags(const string&, const mixed&)’
```

quasilyte: I'll probably add a mixed argument overloading there and then I'll remove WIP prefix.

quasilyte: Added `mixed` and `array<Unknown>` overloadings.
`array<Unknown>` is needed for empty arrays.

AlexK0 (approved):

PR #146, opened 2021-03-03
implemented negative offsets for regexps + offset overflow errors

### LINKED ISSUE #145: preg_match with bad offset arg results differ from PHP
```php
<?php

var_dump(preg_match('/\d+/', 'hello world', $m1, 0, -1));
var_dump(preg_match('/\d+/', 'hello world', $m2, 0, 100));
```
KPHP:
```
bool(false)
int(0)
```
PHP (tested on `7.2`, `7.4` and `8.0`):
```
int(0)
bool(false)
```

quasilyte:
Negative offset is interpreted as an offset from the end of the string.

If offset exceeds the `$subject` length, PHP returns false
with `PHP_PCRE_INTERNAL_ERROR` error (internally, it sets `PCRE2_ERROR_BADOFFSET`).

This behavior is tested on `PHP7.2`, `PHP7.4` and `PHP8.0`.

Fixes #145

----

Could be useful for the review: https://github.com/php/php-src/blob/2dc2436e08e1c80247a940dedeae67135e1c233a/ext/pcre/php_pcre.c#L1224-L1237

### DIFF
--- a/runtime/regexp.cpp
+++ b/runtime/regexp.cpp
@@ -13,6 +13,11 @@
 
 int64_t preg_replace_count_dummy;
 
+// TODO: remove when/if we migrate to pcre2
+#ifndef PCRE2_ERROR_BADOFFSET
+#  define PCRE2_ERROR_BADOFFSET -33
+#endif
+
 static re2::StringPiece RE2_submatch[MAX_SUBPATTERNS];
 // refactor me please :(
 // for i-th match(capturing group)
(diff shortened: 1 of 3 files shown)


runtime/regexp.cpp
```diff
@@ -706,6 +720,12 @@ Optional<int64_t> regexp::match(const string &subject, mixed &matches, bool all_
   pcre_last_error = 0;
 
   int64_t result = 0;
+  offset = initial_offset(subject, offset);
```
AlexK0, line 723: Please, consider using `get_correct_index` instead of `initial_offset`
Something like
```cpp
offset = std::max(subject.get_correct_index(offset), 0_i64)
```

AlexK0 (approved):

runtime/regexp.cpp
quasilyte: Replaced initial_offset with your suggestion.
Nice one. :+1:

AlexK0 (approved):

PR #148, opened 2021-03-04
Run async PHP script in job workers

DrDet:


server/job-workers/job-worker-server.cpp
```diff
@@ -6,131 +6,149 @@
...
+
+} // namespace
+
+int JobWorkerServer::job_parse_execute(connection *c) {
+  assert(c == read_job_connection);
+
+  if (running_job.has_value()) {
+    tvkprintf(job_workers, 1, "Get new job while another job is running. Goes back to event loop now\n");
+    has_delayed_jobs = true;
+    JobStats::get().job_worker_skip_job_due_another_is_running++;
```
DrDet, line 111: I'm afraid it will wake up many times on the same job here...

runtime/job-workers/server-functions.cpp
```diff
@@ -0,0 +1,73 @@
...
+
+void free_job_server_interface_lib() noexcept {
+  if (current_job.request_memory) {
+    vk::singleton<job_workers::SharedMemoryManager>::get().release_slice(current_job.request_memory);
+  }
+  current_job.request_memory = nullptr;
+  current_job.send_reply = nullptr;
+}
+
+class_instance<C$KphpJobWorkerRequest> f$kphp_job_worker_fetch_request() noexcept {
```
DrDet, line 31: Now it's possible to fetch the same job request several times. It's not consistent with `fetch` semantic in other places. Maybe it worth to forbid it somehow?

server/php-queries.h
```diff
@@ -34,7 +38,7 @@ struct net_event_t {
       const char *error_message;
     };
     struct { // job_worker_answer
-      void *job_result_script_memory_ptr;
+      job_workers::SharedMemorySlice *job_result_script_memory_ptr;
```
DrDet, line 41: It's not script memory anymore. Let's remove it to `job_result_memory_slice_ptr`

runtime/job-workers/server-functions.cpp
AlexK0: It make sense, let's leave as is for a while, and add to TODO.

AlexK0 (approved):

PR #153, opened 2021-03-10
Support sha224, sha384, sha512 for hash() and hash_hmac().

AlexK0:
This patch enables `sha224`, `sha384`, `sha512` algorithms for `hash()` and `hash_hmac()`.
Also `hash_hmac_algos()` was implemented.

### DIFF
--- a/runtime/openssl.cpp
+++ b/runtime/openssl.cpp
@@ -26,6 +26,7 @@
 #include "common/smart_ptrs/unique_ptr_with_delete_function.h"
 #include "common/wrappers/openssl.h"
 #include "common/wrappers/string_view.h"
+#include "common/wrappers/to_array.h"
 
 #include "runtime/array_functions.h"
 #include "runtime/critical_section.h"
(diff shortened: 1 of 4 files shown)


unserialize (approved):

runtime/openssl.cpp
```diff
@@ -36,127 +37,110 @@
...
-      res.assign(64, false);
-    }
+struct HashTraits {
+private:
+  template<class F>
+  string call_hash_algo(bool raw_output, const F &hash_alho) const noexcept {
+    string res{hash_len * (raw_output ? 1 : 2), false};
 
-    dl::critical_section_call(SHA256, reinterpret_cast<const unsigned char *>(s.c_str()), s.size(), reinterpret_cast<unsigned char *>(res.buffer()));
+    hash_alho(res);
```
eugene536, line 48: nit) hash_АЛFА

runtime/openssl.cpp
```diff
@@ -253,6 +237,18 @@ int64_t f$crc32_file(const string &file_name) {
   return res ^ std::numeric_limits<uint32_t>::max();
 }
 
+bool f$hash_equals(const string &known_string, const string &user_string) noexcept {
+  if (known_string.size() != user_string.size()) {
+    return false;
+  }
+  int result = 0;
+  // This is security sensitive code. Do not optimize this for speed
```
eugene536, line 245: nit) я бы еще вот этот коментал скопипастил
/* {{{ Compares two strings using the same time whether they're equal or not.
   A difference in length will leak */

unserialize: Sorry, we don't understand Russian in Team KPHP 😀

eugene536: LOL, I forgot about it

eugene536 (approved):

PR #159, opened 2021-03-16
Increase http server limits.

AlexK0:
Subj

### DIFF
--- a/net/net-http-server.h
+++ b/net/net-http-server.h
@@ -9,7 +9,9 @@
 
 #include "net/net-connections.h"
 
-#define	MAX_HTTP_HEADER_SIZE	16384
+#define	MAX_HTTP_HEADER_SIZE (64 * 1024)
+#define	MAX_HTTP_HEADER_QUERY_WORD_SIZE (16 * 1024)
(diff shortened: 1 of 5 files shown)


net/net-http-server.h
```diff
@@ -9,7 +9,9 @@
 
 #include "net/net-connections.h"
 
-#define	MAX_HTTP_HEADER_SIZE	16384
+#define	MAX_HTTP_HEADER_SIZE (64 * 1024)
```
DrDet, line 12: I suspect it must correlate with macro `NET_BUFFER_SIZE` from `net/net-buffers.cpp` somehow. But I'm not sure. This buffers are used via `nbit_iterator_t c->Q`, if I'm not mistaken.

AlexK0: I've found the following logic: later in the execute call it combines (glues) these net buffers into one buffer with MAX_HTTP_HEADER_SIZE size.

https://github.com/VKCOM/kphp/blob/master/server/php-engine.cpp#L544

DrDet (approved):

quasilyte (approved):

unserialize (approved):

PR #162, opened 2021-03-17
WIP: function colors

i582:
TODO: new description


compiler/pipes/check-color.h
```diff
@@ -0,0 +1,230 @@
...
+      if (i != 0) {
+        desc += "\n\n";
+      }
+    }
+
+    return desc;
+  }
+
+  void on_start() final {
+    Stacktrace stacktrace{};
```
i582, line 217: Unnecessary variable. Remained after refactoring.

compiler/function-colors.cpp
```diff
@@ -0,0 +1,116 @@
...
+  for (const auto &pair : str2color_type) {
+    if (pair.second == color) {
+      return pair.first;
+    }
+  }
+  return "";
+}
+
+const ColorContainer::ColorsRaw &ColorContainer::colors() {
+  return this->data;
```
quasilyte, line 41: We usually don't use explicit `this->` member access in a new code.

compiler/pipes/check-color.h
```diff
@@ -0,0 +1,373 @@
+// Compiler for PHP (aka KPHP)
+// Copyright (c) 2020 LLC «V Kontakte»
```
quasilyte, line 2: It's not a big deal, but I would use the 2021 copyright year for files that were created in 2021.

compiler/pipes/check-color.h
```diff
@@ -0,0 +1,373 @@
...
+  }
+
+  bool check_function(FunctionPtr function) const override {
+    return !function->is_extern();
+  }
+
+private:
+  // check_func_caller is a function that checks functions above the given in callstack.
+  void check_func_caller(const Callstack &stacktrace, fp::Node *rule, const FunctionPtr &func) const {
+    // If the current rule describes an error.
```
quasilyte, line 102: We don't have a good document that describes the comments style, but there is a consensus that small inline comments inside a function/method should not start from a capital letter and are preferred not to have a period that ends them.

Take a look at the recently committed code to infer the comment rules. They're not strict, but it's easier if we try to keep that style in sync.

So:
```diff
- // If the current rule describes an error.
+ // if the current rule describes an error
```

compiler/function-colors.h
```diff
@@ -0,0 +1,306 @@
...
+
+#include <common/termformat/termformat.h>
+#include <compiler/stage.h>
+
+// function_palette namespace contains all the necessary things to
+// implement the concept of colors for functions.
+namespace function_palette {
+
+// color_t enumeration describes all possible colors.
+enum class color_t {
```
quasilyte, line 19: I don't think that we use `_t` suffix convention.
(Or if we do somewhere, it's probably a legacy code.)
I'm not sure about `Color` vs `color` (upper camel case or just camel case); maybe someone else can suggest something. But aside from that, consider using a enum named `FunctionColor` instead of namespace+Color enum; but @AlexK0 may give more details on that, whether it's OK to make a new namespace here.

compiler/pipes/check-color.h
```diff
@@ -0,0 +1,373 @@
...
+
+#include <utility>
+
+#include "common/algorithms/contains.h"
+#include "common/termformat/termformat.h"
+#include "compiler/compiler-core.h"
+#include "compiler/function-colors.h"
+#include "compiler/function-pass.h"
+
+namespace fp = function_palette;
```
quasilyte, line 15: `fp` is not an obvious shorthand for the `function_palette` in my opinion.

compiler/function-colors.cpp
i582: Thanks, corrected.

compiler/pipes/check-color.h
i582: Thanks, corrected.

compiler/pipes/check-color.h
i582: Yep, removed.

compiler/function-colors.h
i582: Now `color_t` is an alias for `uint64_t`, so it's not a class, but a type, so the suffix is probably needed here.

PR #171, opened 2021-03-23
implement type hints compatibility checks for methods implementation

quasilyte:
implement type hints compatibility checks for methods implementation

This change makes KPHP compiler give an error if implemented methods
do not satisfy PHP signatures compatibility rules.

Signature compatibility rules:

> https://www.php.net/manual/en/language.oop5.basic.php#language.oop.lsp

Since PHP7.4, these rules were improved and it became possible to
use more specific types as implementation return type and
less specific types as implementation param types.

PHP covariance and contravariance:

> https://www.php.net/manual/en/language.oop5.variance.php

These tests were removed from the Zend test list:

        Zend/tests/type_declarations/parameter_type_variance.phpt
        Zend/tests/type_declarations/variance/parent_in_class_failure2.phpt
        Zend/tests/bug60573.phpt
        Zend/tests/bug60573_2.phpt

They test that PHP would give an error for the incompatible signatures;
KPHP gives a compile-time error, so we can't map these tests directly.

They're ported as:

        Zend/tests/type_declarations/parameter_type_variance.phpt -> zend_test_error.php
        Zend/tests/type_declarations/variance/parent_in_class_failure2.phpt -> zend_test_error1.php
        Zend/tests/bug60573.phpt -> zend_test_error2.php
        Zend/tests/bug60573_2.phpt -> zend_test_error3.php

### DIFF
--- a/compiler/data/virtual-method-generator.cpp
+++ b/compiler/data/virtual-method-generator.cpp
@@ -12,7 +12,7 @@
 #include "compiler/data/class-data.h"
 #include "compiler/data/function-data.h"
 #include "compiler/data/src-file.h"
-#include "compiler/debug.h"
+#include "compiler/type-hint.h"
 #include "compiler/gentree.h"
 #include "compiler/stage.h"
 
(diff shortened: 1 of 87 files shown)


compiler/data/virtual-method-generator.cpp
```diff
@@ -79,6 +183,22 @@ bool check_that_signatures_are_same(FunctionPtr interface_function, ClassPtr con
...
+                                derived_method->get_human_readable_name(),
+                                interface_function->get_human_readable_name());
+      if (!error_info.empty()) {
+        message += "\n" + error_info;
+      }
+      kphp_error(false, message);
+    }
+  }
+
+  // do we want to return `check_signatures_variance()` result here instead of `true`?
```
unserialize, line 201: I think, true is enough

compiler/pipes/check-type-hint-variance.cpp
```diff
@@ -0,0 +1,82 @@
...
+
+  for (const auto &derived_class : f->class_id->derived_classes) {
+    const FunctionPtr derived_f = get_derived_method(f, derived_class);
+    if (!derived_f) {
+      continue;
+    }
+
+    // if a base method has a return type hint, derived method should have it too
+    if (f->return_typehint && !derived_f->return_typehint) {
+      kphp_error(false, fmt_format("{}() should provide a return type hint compatible with {}()",
```
unserialize, line 59: Don't you like inserting an expression into `kphp_error` instead of `false`?
```
kphp_error(!f->return_typehint || derived_f->return_typehint, 
           fmt_format(...));
```
Same for below.

compiler/pipes/check-type-hint-variance.cpp
```diff
@@ -0,0 +1,82 @@
...
+#include "compiler/pipes/check-type-hint-variance.h"
+#include "compiler/data/class-data.h"
+#include "compiler/data/function-data.h"
+#include "compiler/type-hint.h"
+
+#include "common/containers/final_action.h"
+
+namespace {
+
+FunctionPtr get_derived_method(FunctionPtr f, ClassPtr derived_class) {
```
unserialize, line 14: f -> parent_method?

compiler/pipes/check-type-hint-variance.cpp
```diff
@@ -0,0 +1,82 @@
...
+    auto *member = derived_class->members.get_static_method(f->local_name());
+    return member ? member->function : FunctionPtr{};
+  }
+  auto *member = derived_class->members.get_instance_method(f->local_name());
+  return member ? member->function : FunctionPtr{};
+}
+
+} // namespace
+
+void CheckTypeHintVariance::execute(FunctionPtr f, DataStream<FunctionPtr> &os) {
```
unserialize, line 25: It's really hard to understand, why some checks are made here, in a separate pipe, and some in `check_signatures_variance()` — in a function named the same, but in another file and another pipe... Can't they be joined? Can you please add a comment of such dictinction?

compiler/data/virtual-method-generator.cpp
```diff
@@ -31,6 +31,110 @@ VertexAdaptor<op_func_call> create_instance_cast_to(VertexAdaptor<op_var> instan
...
+// derived type can be less specific than the base type (contravariance)
+// see https://www.php.net/manual/en/language.oop5.variance.php
+bool param_type_hints_compatible(const TypeHint* base, const TypeHint* derived) {
+  // doing a swap + same argument order to avoid "suspicious arguments order" inspection;
+  // we need to call a return_type_hints_compatible with reversed arguments order
+  std::swap(base, derived);
+  return return_type_hints_compatible(base, derived);
+}
+
+bool check_signatures_variance(FunctionPtr interface_function, FunctionPtr derived_method, std::string &error_info) {
```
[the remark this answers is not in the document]
quasilyte: Base function, for sure.

compiler/pipes/check-type-hint-variance.cpp
quasilyte: There is a comment above the pipe class.

image


The reason is: we merge phpdoc and typehints after the parse-and-apply-phpdocs.
To avoid missing some incompatible signatures that are only compatible due to the phpdocs (which PHP ignores) we have an extra pass that verifies that we follow the PHP rules and will require type hints if they're needed.
So that pipe checks "whether you should add a type hint here" and "whether you can't use a type hint here". It doesn't really check types compatiblity.

Later, in virtual-method-generator.cpp we use the combined type info to check if signatures are actually compatible. So we're probably more strict than PHP, but that the intent: we don't want the code to compile as KPHP and fail as PHP, but we don't have a goal of 100% strict PHP-to-KPHP match here. It's OK to reject some badly typed code for us.

unserialize (approved):

unserialize (approved):

PR #173, opened 2021-03-24
Improve backtrace experience

AlexK0:
The aim of the patch is to improve the sentry backtraces. 

The main pain is that we have a over-inlining of the compiled code and as a result addr2line misses and produces very wired traces, e.g. a custom function call from the mixed destructor. 

I've made some experiments and found that a moving the mixed destructor to the cpp file helps here and after that addr2line shows fine traces.

So, I suggest:
- move string and mixed desctructors to cpp files
- remove force inlining from some descturtors

Yeh, it's unclear how it would affect the performance, but I believe, that we won't notice it, may be only with micro benchmarks.

### DIFF
--- a/runtime/string.cpp
+++ b/runtime/string.cpp
@@ -1,0 +1,10 @@
+// Compiler for PHP (aka KPHP)
+// Copyright (c) 2021 LLC «V Kontakte»
+// Distributed under the GPL v3 License, see LICENSE.notice.txt
+
+#include "runtime/kphp_core.h"
+
+// Don't move this destructor to the headers, it spoils addr2line traces
+string::~string() noexcept {
+  destroy();
+}
--- a/common/smart_ptrs/intrusive_ptr.h
+++ b/common/smart_ptrs/intrusive_ptr.h
@@ -50,7 +50,7 @@ public:
     return *this;
   }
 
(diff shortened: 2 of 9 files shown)


quasilyte (approved):

unserialize:
Does it affect the resulting binary size somehow?

runtime/string.cpp
```diff
@@ -0,0 +1,10 @@
+// Compiler for PHP (aka KPHP)
+// Copyright (c) 2021 LLC «V Kontakte»
+// Distributed under the GPL v3 License, see LICENSE.notice.txt
+
+#include "runtime/kphp_core.h"
+
+// Don't move this destructor to the headers, it spoils addr2line traces
+string::~string() noexcept {
+  destroy();
```
unserialize, line 9: Why mixed::destroy() was moved to a .cpp file, but string::destroy() wasn't?

AlexK0: string::destroy() is quite simple and it is used in string.inl.

mixed::destroy() is used in in mixed.inl as well, but it produces much more asm (due to array destructor), therefore I moved it here.

AlexK0:
> Does it affect the resulting binary size somehow?

Yes, it reduces the size of the binary a bit.

unserialize (approved):

PR #174, opened 2021-03-25
allow `T->regexp` implicit casts even with `strict_types=1`

quasilyte:
When `strict_types=1` is used, there will be no implicit `string->regexp` and `mixed->regexp`
conversions, but in fact we don't need them as there are various overloadings
that can work with all such types (string, regexp, mixed).
    
With this change, it will be possible to use string-typed arguments
to `preg_*` functions when `declare(strict_types=1)` is used.

### DIFF
--- a/compiler/pipes/preprocess-function.cpp
+++ b/compiler/pipes/preprocess-function.cpp
@@ -16,6 +16,21 @@
 #include "compiler/name-gen.h"
 #include "compiler/type-hint.h"
 
+namespace {
+
+bool implicit_cast_allowed(bool strict_types, const TypeHint *type) {
+  if (!strict_types) {
+    return true;
+  }
+  // converting to regexp is OK even in strict_types mode
+  if (const auto *primitive_type = type->try_as<TypeHintPrimitive>()) {
+    return primitive_type->ptype == tp_regexp;
+  }
+  return false;
+}
+
+} // namespace
+
 class PreprocessFunctionPass final : public FunctionPassBase {
 public:
   DataStream<FunctionPtr> &instance_of_function_template_stream;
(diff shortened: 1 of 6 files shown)


quasilyte:
This change makes kphp2cpp produce string extern patterns instead of regexp-typed constants.
```diff
-extern regexp v$const_regexp$use34ccc993b56c26e;
+extern string v$const_string$use34ccc993b56c26e;
 
-  f$preg_match(v$const_regexp$use34ccc993b56c26e, ...);
+  f$preg_match(v$const_string$use34ccc993b56c26e, ...);
```
Maybe adding an ad-hoc check inside implicit casting code is a better way after all.

quasilyte:
With a new approach, there is no generated code diff anymore.

quasilyte:
It would probably be better to only allow `string->regexp` and disallow all other implicit conversions.
But I don't think that we have enough type information during the function preprocess phase.

compiler/pipes/preprocess-function.cpp
```diff
@@ -16,6 +16,21 @@
 #include "compiler/name-gen.h"
 #include "compiler/type-hint.h"
 
+namespace {
+
+bool implicit_cast_allowed(bool strict_types, const TypeHint *type) {
```
AlexK0, line 21: please, consider the following example

```php
<?php

/** @kphp-strict-types-enable */
declare(strict_types=1);


class A {
  public $x = 1;
}

function test_regexp_crap($re) {
  var_dump(preg_match($re, '123') > 0);
  var_dump(preg_match_all($re, '123') > 0);
}

test_regexp_crap(tuple(1));
```

before this patch kphp2cpp generated a compilation error
```
Starting php to cpp transpiling...

Compilation error at stage: Calc actual calls, gen by type-inferer.cpp:53
  dev.php:12  in test_regexp_crap
    var_dump(preg_match($re, '123') > 0);

pass tuple(int) to argument $regex of preg_match
but it's declared as @param regexp

  dev.php:12  in test_regexp_crap
    var_dump(preg_match($re, '123') > 0);
    $re is tuple(int)

  dev.php:11  in test_regexp_crap
    function test_regexp_crap($re) {
    argument $re inferred as tuple(int)

  dev.php:16
    
    tuple<> is tuple(int)
    1 is int


```

but with this patch kphp2cpp skips this error and generates C++ code which leads a gcc compilation error
```
test_regexp_crap.cpp:11:74: error: no matching function for call to ‘f$preg_match(const std::tuple<long int>&, string&)’
   11 |   f$var_dump(0_i64 < f$preg_match(v$re, v$const_string$us8c039e9e9dfbadf3));
```

quasilyte: Fixed the tuple->regexp conversion.

AlexK0 (approved):

PR #178, opened 2021-03-26
More memory stat.

AlexK0:
This patch adds a bunch of different memory stats.

WIP while #160 is not merged.

### DIFF
--- a/common/dl-utils-lite.h
+++ b/common/dl-utils-lite.h
@@ -42,24 +42,24 @@ void dl_assert__ (const char *expr, const char *file_name, const char *func_name
 #define dl_passert(f, str) dl_assert_impl (f, str, 1)
 #define dl_unreachable(str) dl_assert (0, str)
 
-typedef struct {
+struct pid_info_t {
   unsigned long long utime;
   unsigned long long stime;
   unsigned long long cutime;
   unsigned long long cstime;
   unsigned long long starttime;
-} pid_info_t;
+};
 
-typedef struct {
-  unsigned long long vm_peak;
-  unsigned long long vm;
-  unsigned long long rss_peak;
-  unsigned long long rss;
-  unsigned long long rss_file;
-  unsigned long long rss_shmem;
-} mem_info_t;
+struct mem_info_t {
+  uint32_t vm_peak;
+  uint32_t vm;
+  uint32_t rss_peak;
+  uint32_t rss;
+  uint32_t rss_file;
+  uint32_t rss_shmem;
+};
 
-int get_mem_stats (pid_t pid, mem_info_t *info);
+mem_info_t get_self_mem_stats();
 int get_pid_info (pid_t pid, pid_info_t *info);
 unsigned long long get_pid_start_time (pid_t pid);
 int get_cpu_total (unsigned long long *cpu_total);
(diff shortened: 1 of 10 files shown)


common/dl-utils-lite.h
```diff
@@ -42,24 +42,24 @@ void dl_assert__ (const char *expr, const char *file_name, const char *func_name
...
-typedef struct {
-  unsigned long long vm_peak;
-  unsigned long long vm;
-  unsigned long long rss_peak;
-  unsigned long long rss;
-  unsigned long long rss_file;
-  unsigned long long rss_shmem;
-} mem_info_t;
+struct mem_info_t {
+  uint32_t vm_peak;
```
eugene536, line 54: will `uint32_t` be enough?

server/php-worker-stats.h
```diff
@@ -57,7 +64,24 @@ class PhpWorkerStats {
...
+
+  std::array<int64_t, PERCENTILE_SAMPLES> workers_rss_usage_{};
+  std::array<int64_t, PERCENTILE_SAMPLES> workers_vms_usage_{};
+  std::array<int64_t, PERCENTILE_SAMPLES> workers_shm_usage_{};
+
+  uint64_t rss_kb_sum_{0};
+  uint64_t vms_kb_sum_{0};
+
+  uint32_t rss_kb_max_{0};
+  uint32_t vms_kb_max_{0};
```
eugene536, line 82: Why do we use `uint64_t` for `rss_kb_sum_` but `uint32_t` for `rss_kb_max`?

common/dl-utils-lite.h
AlexK0: Yep, they are in kbytes.

server/php-worker-stats.h
AlexK0: Yeah..

rss_kb_max - in kbytes, so uint32_t is enough here
rss_kb_sum_ - sum for all processes in kbytes, and I'm not sure if uint32_t is enough for sum, but uint64_t is enough
workers_rss_usage_ - in bytes..

I'm changing rss_kb_sum_ to int64, to be more consistent.

DrDet (approved):

server/php-master.cpp
```diff
@@ -1493,20 +1467,23 @@ class QPSCalculator {
   }
 
   void update(double time_point, const PhpWorkerStats &new_worker_stats) {
-    if (!prev_worker_stats_.total_queries()) {
-      prev_worker_stats_ = new_worker_stats;
+    if (!prev_total_queries_) {
```
unserialize, line 1470: I can't find, when this field is set to 0. I mean, while execution, when multiple requests are processed one by one

AlexK0: It is set to 0 only on this struct initialization (1502 line), and then it is only incrementally updated.

unserialize: It's not annulated between requests. Is it on purpose?

AlexK0: Yes, it is ok. new_worker_stats.total_queries() returns the monotonically incremented requests count, so we get a delta between the old value and the new one.

unserialize (approved):

PR #189, opened 2021-04-08
Function colors and palette

i582:
# Function colors

The concept of function colors consists of two parts: colors and their mixing palette. 
The colors define the properties of the functions, and the palette determines the order 
in which the colors of the functions can be mixed.

Function colors allow us to control function calls, disallowing some of them if they 
contradict some logic or rules.

A mixing palette is usually represented as a set of rule groups. Each group consists of
a set of rules, where the first rule describes an error, that is, a certain set of colors that 
cannot be mixed, and the rest are exceptions when it is allowed to do so.

For example:

```php
[
    "red green"        => "error",  // you can not do it this way
    "red yellow green" => "ok"      // but if there is also a yellow color, then you can
]
```

It is forbidden to mix `red` with `green`, but if `yellow` is mixed between them, then it is allowed.

Color mixing means chaining of function calls, where the chaining can be of length `>= 1`.

That is, we can say that if a function is `red`, then a `green` function cannot be called from it. 
But if the function has both `red` and `yellow` colors, we can call the function with `green` color. 
We can also say that if a function has a `red` color, we can call a function with `yellow` and `green`
 color from it.

To set the color for the function, added the annotation `@kphp-color`, which takes one argument.
To set multiple colors, you need to add several annotations (order is important).

```php
/**
 * @kphp-color red
 */
function f1() {}

/**
 * @kphp-color red
 * @kphp-color yellow
 */
function f2() {}
```

To define the mixing rules, used the `FUNCTION_PALETTE` constant in the `KphpConfiguration.php` file.

For example:

```php
const FUNCTION_PALETTE = [
  [
    "highload no-highload"              => "Calling no-highload function from highload function",
  ],
  [
    "ssr has-db-access"                 => "Calling function working with the database in the server side rendering function",
    "ssr ssr-allow-db has-db-access"    => 1,
  ],
]
```

A constant is a two-dimensional array or an array of rule groups. Each group consists of a set 
of rules. A rule is a key-value pair, where a string containing a set of colors separated by a space 
is written in the key, and a string describing the error or the number 1 (the fact that the rule does 
not result in an error) is written in the value.

In each group, the rule with an error must be the first rule. All further rules should not lead to an 
error (that is, they should have 1 in the value). All rules after the first are called exception rules, 
that is, they must contain the colors from the first rule and add new ones to them, thereby clarifying.

### DIFF
--- a/compiler/compiler-core.h
+++ b/compiler/compiler-core.h
@@ -19,6 +19,7 @@
 #include "compiler/threading/hash-table.h"
 #include "compiler/tl-classes.h"
 #include "compiler/composer.h"
+#include "compiler/function-colors.h"
 
 class CompilerCore {
 private:
@@ -36,6 +37,7 @@ private:
   TlClasses tl_classes;
   std::vector<std::string> kphp_runtime_opts;
   bool is_untyped_rpc_tl_used{false};
+  function_palette::Palette function_palette;
 
   inline bool try_require_file(SrcFilePtr file);
 
@@ -106,6 +108,14 @@ public:
 
   void add_kphp_runtime_opt(std::string opt) { kphp_runtime_opts.emplace_back(std::move(opt)); }
   const std::vector<std::string> &get_kphp_runtime_opts() const { return kphp_runtime_opts; }
+
+  void set_function_palette(function_palette::Palette &&palette) {
+    function_palette = palette;
+  }
+
+  function_palette::Palette &get_function_palette() {
+    return function_palette;
+  }
 
   void set_untyped_rpc_tl_used() {
     is_untyped_rpc_tl_used = true;
--- a/compiler/compiler.cmake
+++ b/compiler/compiler.cmake
@@ -168,6 +168,7 @@ prepend(KPHP_COMPILER_SOURCES ${KPHP_COMPILER_DIR}/
         cpp-dest-dir-initializer.cpp
         debug.cpp
         compiler-settings.cpp
+        function-colors.cpp
         gentree.cpp
         index.cpp
         lexer.cpp
--- a/compiler/function-colors.cpp
+++ b/compiler/function-colors.cpp
@@ -1,0 +1,79 @@
+// Compiler for PHP (aka KPHP)
+// Copyright (c) 2021 LLC «V Kontakte»
+// Distributed under the GPL v3 License, see LICENSE.notice.txt
+
+#include "compiler/function-colors.h"
+
+#include "common/algorithms/string-algorithms.h"
+
+using namespace function_palette;
+
+PaletteRule::PaletteRule(std::vector<color_t> &&colors, std::string error)
+  : colors(std::move(colors))
+  , error(std::move(error)) {
+
+  for (color_t color : this->colors) {
+    mask |= color;
+  }
+}
+
+std::string PaletteRule::as_human_readable(const Palette &palette) const {
+  return vk::join(colors, " ", [palette](auto color) { return palette.get_name_by_color(color); });
+}
+
+// --------------------------------------------
+
+color_t Palette::register_color_name(const std::string &color_name) {
+  auto found = color_names_mapping.find(color_name);
+  if (found != color_names_mapping.end()) {
+    return found->second;
+  }
+
+  int bit_shift = 1 + color_names_mapping.size();
+  color_t color = 1ULL << bit_shift;
+
+  color_names_mapping[color_name] = color;
+  kphp_error(color_names_mapping.size() != 62, "Reached limit of colors in palette — more than 62 colors");
+  return color;
+}
+
+color_t Palette::get_color_by_name(const std::string &color_name) const {
+  const auto found = color_names_mapping.find(color_name);
+  kphp_assert(found != color_names_mapping.end());
+  return found->second;
+}
+
+std::string Palette::get_name_by_color(color_t color) const {
+  for (const auto &pair : color_names_mapping) {
+    if (pair.second == color) {
+      return pair.first;
+    }
+  }
+  return std::to_string(color);
+}
+
+// --------------------------------------------
+
+void ColorContainer::add(color_t color) {
+  // append color to the end of forward list
+  // a bit inconvenient, but we use forward_list intentionally to consume only 8 bytes on emptiness (the most common case)
+  auto before_end = std::next(sep_colors.before_begin(), std::distance(sep_colors.begin(), sep_colors.end()));
+  sep_colors.emplace_after(before_end, color);
+}
+
+bool ColorContainer::contains(color_t color) const {
+  return std::any_of(sep_colors.begin(), sep_colors.end(), [=](color_t c) { return c == color; });
+}
+
+std::string ColorContainer::to_string(const Palette &palette, colors_mask_t with_highlights) const {
+  std::string desc;
+
+  for (color_t color : sep_colors) {
+    if (with_highlights == 0 || with_highlights & color) {
+      desc += "@";
+      desc += palette.get_name_by_color(color);
+    }
+  }
+
+  return desc;
+}
--- a/compiler/function-colors.h
+++ b/compiler/function-colors.h
@@ -1,0 +1,88 @@
+// Compiler for PHP (aka KPHP)
+// Copyright (c) 2021 LLC «V Kontakte»
+// Distributed under the GPL v3 License, see LICENSE.notice.txt
+
+#pragma once
+
+#include <forward_list>
+#include <map>
+
+#include "common/termformat/termformat.h"
+#include "compiler/stage.h"
+
+namespace function_palette {
+
+using color_t = uint64_t;
+using colors_mask_t = uint64_t;
+
+class Palette;
+
+// rules are representation of human-written "api has-curl" => "error text" or "api allow-curl has-curl" => 1
+// all colors are pre-converted to numeric while reading strings
+struct PaletteRule {
+  std::vector<color_t> colors;
+  colors_mask_t mask{0};
+  std::string error;
+
+  PaletteRule(std::vector<color_t> &&colors, std::string error);
+
+  std::string as_human_readable(const Palette &palette) const;
+
+  bool is_error() const {
+    return !error.empty();
+  }
+
+  bool contains_in(colors_mask_t colors_mask) const {
+    return (mask & colors_mask) == mask;
+  }
+};
+
+// a ruleset is a group of rules, order is important
+// typically, it looks like one error rule and some "exceptions" — more specific color chains with no error
+using PaletteRuleset = std::vector<PaletteRule>;
+
+// a palette is a group of rulesets
+// all colors are stored as numbers, not as strings:
+// they are numbers 1<<n, as we want to use bitmasks to quickly test whether to check a rule for callstack
+class Palette {
+  std::vector<PaletteRuleset> rulesets;
+  std::map<std::string, color_t> color_names_mapping;
+
+public:
+  color_t register_color_name(const std::string &color_name);
+
+  void add_ruleset(PaletteRuleset &&ruleset) {
+    rulesets.emplace_back(std::move(ruleset));
+  }
+
+  const std::vector<PaletteRuleset> &get_rulesets() const {
+    return rulesets;
+  }
+
+  bool color_exists(const std::string &color_name) const {
+    return color_names_mapping.find(color_name) != color_names_mapping.end();
+  }
+
+  color_t get_color_by_name(const std::string &color_name) const;
+  std::string get_name_by_color(color_t color) const;
+};
+
+// a class containing colors after @kphp-color parsing above each function (order is important)
+class ColorContainer {
+  std::forward_list<color_t> sep_colors; // empty almost everywhere
+
+public:
+  ColorContainer() = default;
+
+  bool empty() const { return sep_colors.empty(); }
+
+  auto begin() const { return sep_colors.begin(); }
+  auto end() const { return sep_colors.end(); }
+
+  void add(color_t color);
+  bool contains(color_t color) const;
+
+  std::string to_string(const Palette &palette, colors_mask_t with_highlights) const;
+};
+
+} // namespace function_palette
--- a/compiler/pipes/calc-bad-vars.cpp
+++ b/compiler/pipes/calc-bad-vars.cpp
@@ -116,12 +116,255 @@ struct FuncCallGraph {
   }
 };
 
+class CallstackOfColoredFunctions {
+  std::vector<FunctionPtr> stack;                       // functions placed in order, only colored functions
+  std::vector<function_palette::color_t> colors_chain;  // blended colors of stacked functions, one-by-one
+  std::unordered_set<FunctionPtr> index_set;            // a quick index for contains(), has the same elements as stack
+  function_palette::colors_mask_t colors_mask{0};       // mask of all colors_chain
+
+  void recalc_mask() {
+    colors_mask = 0;
+    for (function_palette::color_t c : colors_chain) {
+      colors_mask |= c;
+    }
+  }
+
+public:
+  CallstackOfColoredFunctions() {
+    stack.reserve(10);
+    index_set.reserve(10);
+  }
+
+  size_t size() const {
+    return stack.size();
+  }
+
+  const std::vector<FunctionPtr> &as_vector() const {
+    return stack;
+  }
+
+  const std::vector<function_palette::color_t> &get_colors_chain() const {
+    return colors_chain;
+  }
+
+  function_palette::colors_mask_t get_colors_mask() const {
+    return colors_mask;
+  }
+
+  void push_back(const FunctionPtr &el) {
+    stack.push_back(el);
+    index_set.insert(el);
+    for (const auto &sep_color : el->colors) {
+      colors_chain.push_back(sep_color);
+    }
+    recalc_mask();
+  }
+
+  void pop_back() {
+    FunctionPtr pop = stack.back();
+    index_set.erase(pop);
+    stack.pop_back();
+    for (const auto &sep_color __attribute__((unused)) : pop->colors) {
+      colors_chain.pop_back();
+    }
+    recalc_mask();
+  }
+
+  bool contains(const FunctionPtr &el) const {
+    return index_set.find(el) != index_set.end();
+  }
+};
+
+// here we check all palette rules like "api has-curl" or "messages-module messages-internals"
+// start from the root, make dfs expanding next_with_colors and check all rules for each dfs chain
+// these chains are left-to-right, e.g. colored f1->f2->f3, we find and check f1, then f1->f2, then f1->f2->f3
+class CheckFunctionsColors {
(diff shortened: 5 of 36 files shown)


compiler/compiler-core.h
```diff
@@ -1,5 +1,5 @@
 // Compiler for PHP (aka KPHP)
-// Copyright (c) 2020 LLC «V Kontakte»
+// Copyright (c) 2021 LLC «V Kontakte»
```
unserialize, line 2: Let's not change any file headers (this comment is actual for all files below). It's okay, that the year 2020 remains here.

compiler/compiler.cmake
```diff
@@ -177,7 +177,8 @@ prepend(KPHP_COMPILER_SOURCES ${KPHP_COMPILER_DIR}/
         stats.cpp
         type-hint.cpp
         tl-classes.cpp
-        vertex.cpp)
+        vertex.cpp
+        function-colors.cpp)
```
unserialize, line 181: Lexicographical sorting

compiler/function-colors.cpp
```diff
@@ -0,0 +1,84 @@
...
+  return sep_colors_;
+}
+
+void ColorContainer::add(color_t color) {
+  data |= color;
+  count++;
+  sep_colors_.push_back(color);
+}
+
+size_t ColorContainer::size() const noexcept {
```
unserialize, line 23: It's better to implement such very small functions right in the .h file, so that they can be inlined

compiler/pipes/calc-bad-vars.cpp
```diff
@@ -115,6 +115,343 @@ struct FuncCallGraph {
   }
 };
 
+using CallstackColors = std::vector<function_palette::color_t>;
+
+class Callstack {
+  using raw = std::vector<FunctionPtr>;
+  using raw_set = std::unordered_set<FunctionPtr>;
+  using it = std::vector<FunctionPtr>::iterator;
+  using element = FunctionPtr;
```
unserialize, line 124: As for me, such usings create more mess than benefit. It's really hard to read "it begin;" and others: you should always keep in mind, that "it" means.
I'd suggest not to use them at all.

compiler/pipes/calc-bad-vars.cpp
```diff
@@ -115,6 +115,343 @@ struct FuncCallGraph {
...
+
+private:
+  raw data;
+  raw_set data_set;
+  it begin;
+  it end;
+
+public:
+  Callstack(size_t cap) {
```
unserialize, line 133: explicit

compiler/pipes/calc-bad-vars.cpp
```diff
@@ -115,6 +115,343 @@ struct FuncCallGraph {
...
+    data.reserve(cap);
+    data_set.reserve(cap);
+    begin = data.begin();
+    end = data.end();
+  }
+
+public:
```
unserialize, line 140: 'public' duplicated

compiler/pipes/calc-bad-vars.cpp
```diff
@@ -115,6 +115,343 @@ struct FuncCallGraph {
...
+    data_set.erase(data.back());
+    data.pop_back();
+  }
+
+  bool contains(const element &el) const {
+    return vk::contains(data_set, el);
+  }
+
+  size_t hash() {
+    return front()->id + (back()->id * 100);
```
unserialize, line 170: Is it okay, that all elements between back and front are skipped?

compiler/pipes/filter-only-actually-used.cpp
```diff
@@ -86,6 +140,12 @@ void calc_throws_and_body_value_through_call_edges(std::vector<FunctionAndEdges>
     if (fun->body_seq == FunctionData::body_value::non_empty) {
       calc_non_empty_body_dfs(fun, non_empty_body_graph);
     }
+    if (fun->type == FunctionData::func_local && !fun->colors.empty()) {
+      {
+        AutoProfiler pp{get_p_check_func()};
```
unserialize, line 145: Forgotten debug

compiler/pipes/parse-and-apply-phpdoc.cpp
```diff
@@ -146,6 +146,20 @@ class ParseAndApplyPhpDocForFunction {
 
   void parse_kphp_doc_tag(const php_doc_tag &tag) {
     switch (tag.type) {
+      case php_doc_tag::kphp_color: {
+        const auto raw_colors = split(tag.value, ' ');
```
unserialize, line 150: Let's deny multiple colors in one tag. The format of @kphp-color should be:
@kphp-color color [optional comment]
So, if a user wants multiple colors – he writes multiple tags.
It will be much easier to find used colors in a regular code using a simple text search.

compiler/pipes/register-kphp-configuration.cpp
```diff
@@ -19,7 +19,7 @@ bool RegisterKphpConfiguration::check_function(FunctionPtr function) const {
 void RegisterKphpConfiguration::on_start() {
   const auto klass = current_function->class_id;
   kphp_error(
-    !klass->members.has_any_instance_method() &&
+//    !klass->members.has_any_instance_method() &&
```
i582, line 22: Due to the transfer of processing `KphpConfiguration`, an unfiltered constructor remains in it and because of this, the check here is true and therefore an error is produced.
The line will be uncommented when the color definition goes to the `json` file.

unserialize, line 22: ?

i582: We now process `RegisterKphpConfiguration` before `ParseAndApplyPhpdocF`, and 
all this together is processed before filtering functions and the `KphpConfiguration` 
is left with a constructor due to which an error occurs.

compiler/data/function-data.cpp
```diff
@@ -267,6 +267,8 @@ string FunctionData::get_human_readable_name(bool add_details) const {
   std::string result_name;
   if (is_lambda()) {
     result_name = get_params().empty() ? "anonymous()" : "anonymous(...)";
+  } else if(type == FunctionData::func_main) {
+    return "todo";//todo
```
i582, line 271: Are we going to make readable names for the `main` functions in this PR or put them in a separate one?

compiler/pipes/parse-and-apply-phpdoc.cpp
```diff
@@ -146,6 +146,18 @@ class ParseAndApplyPhpDocForFunction {
...
+        std::string color_name;
+        is >> color_name;
+
+        kphp_error_return(!color_name.empty(), "An empty tag value");
+        kphp_error_return(G->get_function_palette().color_exists(color_name), "Color missing in palette (either a misprint or a new color that needs to be added)");
+
+        f_->colors.add(G->get_function_palette().get_color_by_name(color_name));
+        break;
+      }
+
```
unserialize, line 160: Move this case below, please

compiler/pipes/calc-bad-vars.h
```diff
@@ -8,6 +8,7 @@
 #include "compiler/pipes/calc-func-dep.h"
 #include "compiler/pipes/sync.h"
 #include "compiler/threading/data-stream.h"
+#include "common/algorithms/contains.h"
```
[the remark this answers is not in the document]
i582: This doesn't seem to be used and can be removed.

compiler/pipes/parse-and-apply-phpdoc.cpp
i582: Moved.

compiler/pipes/calc-bad-vars.cpp
```diff
@@ -115,13 +115,297 @@ struct FuncCallGraph {
...
+  void show_error_on_rule_broken(const CallstackOfColoredFunctions &color_callstack, const function_palette::PaletteRule &rule) {
+    kphp_assert(rule.is_error());
+    if (shown_errors.size() > max_shown_errors_count) {
+      return;
+    }
+
+    auto vector = color_callstack.as_vector();    // f1, f2, f3 — all of them are colored, and their chain breaks the rule
+    std::vector<FunctionPtr> full_callstack;      // will be: src_main -> ... -> f1 -> ... -> f2 -> ... -> f3
+    for (int i = 0; i < vector.size() - 1; ++i) {
+      FunctionPtr cur = vector[i], next = vector[i+1];
```
DrDet, line 318: If vector is empty, `vector.size() - 1` is `2^64 - 1` because of unsigned type. So I would prefer rewrite it like this:
```suggestion
    for (int i = 1; i < vector.size(); ++i) {
      FunctionPtr cur = vector[i - 1], next = vector[i];
```

compiler/pipes/calc-bad-vars.cpp
```diff
@@ -115,13 +115,297 @@ struct FuncCallGraph {
...
+      std::unordered_set<FunctionPtr> shouldnt_appear;
+      kphp_assert(cur->next_with_colors);
+      for (FunctionPtr exclude : *cur->next_with_colors) {
+        if (exclude != next && exclude != cur) {
+          shouldnt_appear.insert(exclude);
+        }
+      }
+      auto next_callstack_part = find_callstack_between_two_functions_bfs(cur, next, shouldnt_appear);
+      full_callstack.insert(full_callstack.end(), next_callstack_part.begin(), next_callstack_part.end() - 1);
```
DrDet, line 327:
```suggestion
      full_callstack.insert(full_callstack.end(), next_callstack_part.begin(), next_callstack_part.rbegin());
```

compiler/pipes/register-kphp-configuration.cpp
```diff
@@ -31,43 +31,110 @@ void RegisterKphpConfiguration::on_start() {
...
+    const std::string *errtext_val = GenTree::get_constexpr_string(value);
+    const auto ok_val = value.try_as<op_int_const>();
+    kphp_error(errtext_val || (ok_val && ok_val->get_string() == "1"), fmt_format("{}::{} map values must be constexpr strings or number 1",
+                                                                                  configuration_class_name_, function_color_palette_name_));
+    return errtext_val != nullptr ? *errtext_val : "";
+  };
+
+  auto colors = parse_rule_key_colors(pair->key());
+  auto error = parse_rule_value_error(pair->value());
+  add_to.add_rule(function_palette::PaletteRule(std::move(colors), std::move(error)));
```
DrDet, line 101:
```suggestion
  add_to.add_rule(function_palette::PaletteRule{std::move(colors), std::move(error)});
```
I think, it's a bit more readable, because it highlights object creation among function calls

compiler/pipes/calc-bad-vars.cpp
i582: Maybe cast to int is easier here?
We have similar explicitly cast pieces of code.

DrDet: I think, the option without `- 1` in size calculation will look more clear and graceful. Also I can't say that it makes code more complicated somehow. Why do you think cast is easier?

compiler/function-colors.cpp
```diff
@@ -0,0 +1,82 @@
...
+
+// --------------------------------------------
+
+void ColorContainer::add(color_t color) {
+  // append color to the end of forward list
+  // a bit inconvenient, but we use forward_list intentionally to consume only 8 bytes on emptiness (the most common case)
+  auto before_end = sep_colors.before_begin();
+  for (const auto &_ __attribute__((unused)) : sep_colors) {
+    ++before_end;
+  }
```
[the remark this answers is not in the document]
i582: Done.

compiler/pipes/calc-bad-vars.cpp
```diff
@@ -115,13 +115,297 @@ struct FuncCallGraph {
...
+    }
+    full_callstack.emplace_back(vector[vector.size() - 1]);
```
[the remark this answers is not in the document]
i582: Done.

compiler/pipes/register-kphp-configuration.cpp
i582: Done.

compiler/function-colors.h
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  bool contains_in(colors_mask_t colors_mask) const {
+    return (mask & colors_mask) == mask;
+  }
+};
+
+// a ruleset is a group of rules, order is important
+// typically, it looks like one error rule and some "exceptions" — more specific color chains with no error
+class PaletteRuleset {
```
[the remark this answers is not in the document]
i582: I think so.
Done.

compiler/pipes/calc-bad-vars.cpp
i582: Hmm, it seems we can't use `rbegin` here because it's a reverse iterator.

DrDet: Hmm, sorry. Additionally we need to call `base()` to get the corresponding base iterator, so it will be like this:
`next_callstack_part.rbegin().base()`

DrDet: ok, sorry, let's just leave it as is :)

compiler/pipes/calc-bad-vars.cpp
i582: Loop
```cpp
for (int i = 1; i < vector.size(); ++i) {
```
does not look like a loop until the penultimate element, to understand this you need to look at the code further.

In a loop with an explicit cast, you can see it right away.
```cpp
for (int i = 0; i < static_cast<int>(vector.size()) - 1; ++i) {
```

DrDet: Ok. Alternatively you can just add `assert(!vector.empty())` before the loop and don't add explicit cast

compiler/pipes/calc-real-defines-values.cpp
```diff
@@ -25,6 +27,18 @@ void CalcRealDefinesValuesF::on_finish(DataStream<FunctionPtr> &os) {
   }
 
   stage::die_if_global_errors();
+
+  // after all constants have been inlined, call RegisterKphpConfiguration
+  // do this INSIDE this sync point, BEFORE passing all functions to the next pipe
+  // this is done intentionally, so that further phpdoc parsing has the KphpConfiguration registered
+  {
+    ClassPtr kphpConfigurationClass = G->get_class("KphpConfiguration");
```
AlexK0, line 35: may be 
`SyncC<RegisterKphpConfiguration>{} ` right after `SyncC<CalcRealDefinesValuesF>{}`
with logic in `RegisterKphpConfiguration::execute()` and simple `on_finish` would be better?

I don't like this hack, it looks wired, because in this case `RegisterKphpConfiguration` is not a pipe anymore and should be moved into another place.

compiler/function-colors.h
```diff
@@ -0,0 +1,89 @@
...
+// they are numbers 1<<n, as we want to use bitmasks to quickly test whether to check a rule for callstack
+class Palette {
+  std::vector<PaletteRuleset> rulesets;
+  std::map<std::string, color_t> color_names_mapping;
+
+public:
+  color_t register_color_name(const std::string &color_name);
+
+  void add_ruleset(PaletteRuleset &&ruleset) {
+    rulesets.emplace_back(ruleset);
```
AlexK0, line 56: std::move(ruleset)

compiler/pipes/calc-bad-vars.cpp
i582: Replace with assert.

compiler/pipes/calc-real-defines-values.cpp
unserialize: Okay, I don't like this moment also. But two sync pipes one after one seemed strange for me also :)
How'd you suggest to make this finally?
1) Two sync pipes
2) make RegisterKphpConfiguration not a pipe, but a standalone logic
?

unserialize: Okay, I don't like this moment also. But two sync pipes one after one seemed strange for me also :)
How'd you suggest to make this finally?
1) Two sync pipes
2) make RegisterKphpConfiguration not a pipe, but a standalone logic
?

compiler/pipes/calc-bad-vars.cpp
```diff
@@ -115,13 +115,297 @@ struct FuncCallGraph {
...
+          show_error_on_rule_broken(callstack, rule);
+          was_any_error = true;
+        }
+        break;
+      }
+    }
+
+    if (!was_any_error && func->next_with_colors != nullptr) {
+      for (FunctionPtr next : *func->next_with_colors) {
+        if (callstack.size() < 50 && !callstack.contains(next)) {
```
AlexK0, line 218: What would happen if callstack.size() > 50? Just skip it? Shouldn't we fire some warning?

compiler/pipes/calc-bad-vars.cpp
```diff
@@ -115,13 +115,297 @@ struct FuncCallGraph {
...
+    }
+    recalc_mask();
+  }
+
+  void pop_back() {
+    FunctionPtr pop = stack.back();
+    index_set.erase(pop);
+    stack.pop_back();
+    for (const auto &sep_color __attribute__((unused)) : pop->colors) {
+      colors_chain.pop_back();
```
AlexK0, line 167: `colors_chain.resize(colors_chain.size() - pop->colors.size());`?

compiler/pipes/calc-bad-vars.cpp
unserialize: Just skip, nothing more

compiler/pipes/calc-bad-vars.cpp
unserialize: Hm, I didn't know, that it works ) I'll test. If ok — I'll replace with your suggestion

compiler/pipes/calc-real-defines-values.cpp
AlexK0: From my point of view two sync pipes is the easiest way, and I believe it wouldn't make any performance penalty. An in additional, if we move it into another place, then the `CalcRealDefinesValuesF` pipe should be renamed as well to something like
`CalcRealDefinesValuesAndRegisterConfigurationF`. Therefore, yes, I vote for two sync pipes.

DrDet (approved):

compiler/data/function-data.cpp
unserialize: I decided to do it later, as it has no relation to colors, what this PR is about

compiler/pipes/calc-real-defines-values.cpp
unserialize: Ok, I made two sync pipes. Maybe, we'll discuss this approach later :)

compiler/pipes/calc-bad-vars.cpp
unserialize: pop->colors is a forward_list, so calculating its size will result in the same loop

AlexK0 (approved):

unserialize (approved):

PR #193, opened 2021-04-15
Job workers memory refactoring

AlexK0:
Hi.
This patch is about the job workers memory usage refactoring:
1) Enable attaching/detaching logic for messages, so they don't leak anymore
2) Use lock free algorithms for memory allocations, so it works faster
3) Allow allocate an extra memory (by 1/2/4/8/16/32/64 mb pieces)
4) Copy a response into the script memory in advance (on accepting response) 
5) Send an error if the job didn't send any reply
6) More stats
7) More tests
8) kphp_job_worker_wait accepts |false and returns |null


runtime/job-workers/shared-memory-manager.cpp
```diff
@@ -5,54 +5,132 @@
...
-  slice->resource.init(slice->resource.memory_begin(), slice->resource.get_memory_stats().memory_limit);
-  slice->job_id = 0;
-  slice->job_result_fd_idx = -1;
-  const pid_t prev_pid = slice->owner_pid.exchange(0);
-  php_assert(prev_pid);
-  JobStats::get().currently_memory_slices_used--;
+void SharedMemoryManager::release_shared_message(JobSharedMessage *message) noexcept {
+  dl::CriticalSectionGuard critical_section;
+  owners_table_->workers_table[logname_id].detach(message);
+  if (--message->owners_counter == 0) {
```
DrDet: `message->owners_counter` is always 1 or 0. I think it would be more logical to consider it as bool flag `message->released` and set it to `false` here instead of decrement

server/job-workers/shared-memory-manager.cpp
```diff
@@ -0,0 +1,189 @@
...
+    return message;
+  }
+  ++control_block_->stats.messages.acquire_fails;
+  return nullptr;
+}
+
+void SharedMemoryManager::release_shared_message(JobSharedMessage *message) noexcept {
+  dl::CriticalSectionGuard critical_section;
+  control_block_->workers_table[logname_id].detach(message);
+  if (--message->owners_counter == 0) {
```
DrDet, line 124: `message->owners_counter` is always 1 or 0. I think it would be more logical to consider it as bool flag `message->released` and set it to `false` here instead of decrement

server/job-workers/shared-memory-manager.cpp
```diff
@@ -0,0 +1,189 @@
...
+  //  index => (1 << index) MB:
+  //    0 => 1MB, 1 => 2MB, 2 => 4MB, 3 => 8MB, 4 => 16MB, 5 => 32MB, 6 => 64MB
+  std::array<freelist_t, JOB_EXTRA_MEMORY_BUFFER_BUCKETS> free_extra_memory{};
+};
+
+void SharedMemoryManager::init() noexcept {
+  static_assert(sizeof(ControlBlock) % 8 == 0, "check yourself");
+  assert(!control_block_);
+
+  const size_t processes = vk::singleton<job_workers::JobWorkersContext>::get().job_workers_num + std::max(workers_n, 1);
```
DrDet, line 68: why do we need std::max here?

runtime/job-workers/processing-jobs.cpp
```diff
@@ -3,36 +3,44 @@
...
 #include "runtime/instance-copy-processor.h"
 
+#include "server/job-workers/job-message.h"
+#include "server/job-workers/shared-memory-manager.h"
+
 #include "runtime/job-workers/processing-jobs.h"
 
 namespace job_workers {
 
+struct FinishedJob {
```
DrDet, line 15: Can it be type alias?

server/job-workers/shared-memory-manager.cpp
AlexK0: This is a preparation for immutable memory: for the immutable memory `owners_counter` can be > 1.

server/job-workers/shared-memory-manager.cpp
AlexK0: Because the default value of `workers_n` is 0, so it just a trick, if the number of workers is not set. (I have no idea how it would work for 0 workers and N job workers..)

runtime/job-workers/processing-jobs.cpp
AlexK0: I believe, that the simple struct is a bit clearer. And also, I'm using the forward declaration for it, so the definition of `FinishedJob` can be a private.

DrDet (approved):

PR #194, opened 2021-04-16
Abstract in traits

eugene536:
This PR allows `abstract` methods from traits to be overridden by instance methods from other traits.
On top of that, KPHP will check signatures and parameters more precisely.

There are some subtle issues with abstract methods in traits as PHP checks their signatures unpredictably and some errors can occur by changing the order of `uses` in a class. Look at tests e.g. 006 vs 100 and 005 vs 102 the difference is only in order.

Implementation in KPHP stays close as much as possible to PHP7.4. The website compiles well and tests are passed.

### DIFF
--- a/compiler/vertex-desc.json
+++ b/compiler/vertex-desc.json
@@ -440,6 +440,10 @@
         "type": "bool",
         "default": "false"
       },
+      "type_as_part_of_signature": {
+        "type": "const TypeHint *",
+        "default": "nullptr"
+      },
       "is_callable": {
         "type": "bool",
         "default": "false"
--- a/compiler/data/function-data.cpp
+++ b/compiler/data/function-data.cpp
@@ -192,15 +192,25 @@ bool FunctionData::can_override_method(FunctionPtr interface_method) {
   }
 
   auto cur_params = get_params();
-  int32_t pos_after_this = 1;
+  int32_t pos_after_this = interface_method->has_implicit_this_arg();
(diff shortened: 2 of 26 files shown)


quasilyte:
I wonder if traits signatures are checked with the same compatibility rules as normal base/derived methods.

#171

eugene536:
> I wonder if trait signatures are checked with the same compatibility rules as normal base/derived methods.
> 
> #171

It works differently LOL
look at the 101 test in the PR. Here is the result of this test on 3v4l https://3v4l.org/1PNBA

eugene536:
> I wonder if trait signatures are checked with the same compatibility rules as normal base/derived methods.
> 
> #171

It looks like that, I think it would be better to merge #171 before the PR and use functions from it to check signatures.

compiler/vertex-desc.json
```diff
@@ -440,6 +440,10 @@
         "type": "bool",
         "default": "false"
       },
+      "type_as_part_of_signature": {
+        "type": "const TypeHint *",
+        "default": "nullptr"
+      },
```
unserialize, line 446: I really don't like this field added. 
@quasilyte, how do you think: after #171 is merged, this addition becomes useless?

quasilyte:
I would merge this PR before #171 if @eugene536 will have no ability (or desire) to adjust the code after it.
It's easier for me to combine these two afterward, I think.

PR #198, opened 2021-04-19
fix loosing methods visibility when `instanceof` is used

eugene536:
In the PR fix of losing extra information about variables is present. When a variable is checked with `instanceof` about brother class, KPHP forgets about the original type, so now we find a first child of both classes and cast the variable to it.

### DIFF
--- a/compiler/pipes/transform-to-smart-instanceof.cpp
+++ b/compiler/pipes/transform-to-smart-instanceof.cpp
@@ -4,7 +4,26 @@
 
 #include "compiler/pipes/transform-to-smart-instanceof.h"
 
+#include "common/algorithms/contains.h"
+
+#include "compiler/compiler-core.h"
+#include "compiler/data/class-data.h"
 #include "compiler/name-gen.h"
+
+namespace {
+  ClassPtr get_first_common_child(ClassPtr base1, ClassPtr base2) {
+    auto base1_children = base1->get_all_inheritors();
+    std::unordered_set<ClassPtr> derived_base1{base1_children.begin(), base1_children.end()};
+
+    for (auto base2_child : base2->get_all_inheritors()) {
+      if (vk::contains(derived_base1, base2_child)) {
+        return base2_child;
+      }
+    }
+
+    return {};
+  }
+} // namespace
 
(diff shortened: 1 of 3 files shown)


compiler/pipes/transform-to-smart-instanceof.cpp
```diff
@@ -14,14 +33,24 @@ bool TransformToSmartInstanceof::user_recursion(VertexPtr v) {
...
   run_function_pass(if_vertex->cond(), this);
 
-  add_tmp_var_with_instance_cast(instance_var, condition->rhs(), if_vertex->true_cmd_ref());
+  auto name_of_rhs_class_vertex = condition->rhs().as<op_string>();
+  const std::string &name_of_rhs_class = name_of_rhs_class_vertex->str_val;
+  ClassPtr rhs_class = G->get_class(name_of_rhs_class);
+  kphp_error_act(rhs_class, fmt_format("can't find right operand of `instanceof`: {}", name_of_rhs_class), return false);
+
+  if (!rhs_class->is_parent_of(type_of_lhs) && !type_of_lhs->is_parent_of(rhs_class)) {
+    rhs_class = get_first_common_child(rhs_class, type_of_lhs);
```
AlexK0, line 47: The following example works incorrect on KPHP :(
```php
<?php

interface A {
  public function f_a();
}

interface B {
  public function f_b();
}

class AB implements A, B {
  public function f_a() { echo "AB::f_a\n"; }
  public function f_b() { echo "AB::f_b\n"; }
}

class AB2 implements A, B {
    public function f_a() { echo "AB2::f_a\n"; }
    public function f_b() { echo "AB2::f_b\n"; }
}

function demo(A $a) {
  if ($a instanceof B) {
    $a->f_b();
    $a->f_a();
  }
}

demo(new AB());
demo(new AB2());
```

PHP out:
```
AB::f_b
AB::f_a
AB2::f_b
AB2::f_a
```

KPHP out:
```
AB::f_b
AB::f_a
AB::f_b
AB::f_a
```

eugene536: Yeap, you are right this solution doesn't work at all. I'd close this PR and fix it later with a completely different approach.

PR #200, opened 2021-04-21
fix compilation error when typed lambda is declared but not used

eugene536:
In the PR we pass through pipes the new-generated lambda's `_invoke` method as it is created implicitly in resolving assumptions and will be thrown away lately.

### DIFF
--- a/compiler/pipes/preprocess-function.cpp
+++ b/compiler/pipes/preprocess-function.cpp
@@ -531,6 +531,14 @@ private:
 
       kphp_error(!f->modifiers.is_static() || call->extra_type != op_ex_func_call_arrow, "It's not allowed to call static method through instance var");
       call = set_func_id(call, f);
+      if (!f->is_required && f->is_virtual_method && !f->is_template && f->local_name() == ClassData::NAME_OF_INVOKE_METHOD) {
+        /**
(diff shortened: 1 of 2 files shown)


compiler/pipes/preprocess-function.cpp
```diff
@@ -531,6 +531,9 @@ class PreprocessFunctionPass final : public FunctionPassBase {
 
       kphp_error(!f->modifiers.is_static() || call->extra_type != op_ex_func_call_arrow, "It's not allowed to call static method through instance var");
       call = set_func_id(call, f);
+      if (!f->is_required && f->is_virtual_method && !f->is_template) {
```
quasilyte, line 534: Could you add a short explanation in a form of comment what this condition means and why
we push this function along the pipe (what could happen if we'll not do so).

You mentioned this info in the commit message, but it's unclear when we're looking at the code; at least for me.

eugene536: No problem, bro. I assumed that coders can blame this piece of code and look at the test case for clarification.
but the comment is much better, so added it.

On top of that, I added one more check for lambdas inside `if`

quasilyte (approved):

AlexK0 (approved):

PR #202, opened 2021-04-21
remove unused fix_php_bugs flag from the regexp module

quasilyte:
It's always false; it makes the code slightly more complicated.
If we ever want to add conditional fixes to the regular expressions,
it could be a different flag or we can take it back, but the
current fixes are not needed.

### DIFF
--- a/runtime/regexp.cpp
+++ b/runtime/regexp.cpp
@@ -722,7 +722,7 @@ Optional<int64_t> regexp::match(const string &subject, mixed &matches, bool all_
     return false;
   }
   while (offset <= int64_t{subject.size()}) {
-    int64_t count = exec(subject, offset, second_try);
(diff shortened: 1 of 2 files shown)


runtime/regexp.cpp
```diff
@@ -751,7 +751,7 @@ Optional<int64_t> regexp::match(const string &subject, mixed &matches, bool all_
         matches[i].push_back(match_str);
       }
     } else {
-      int64_t size = fix_php_bugs ? subpatterns_count : count;
+      int64_t size = count;
```
AlexK0, line 754: May be we can drop `size` and use `count` directly?
Same for diff below.

quasilyte: Fixed.

AlexK0 (approved):

PR #207, opened 2021-04-22
add trailing zeros for bcmul results

### LINKED ISSUE #206: bcmul ignores scale param
This was the behavior in PHP7.2 and prior.
Starting from PHP7.3 and above, scale is not ignored anymore.
https://bugs.php.net/bug.php?id=66364
Affected test: `phpt/dl/483_bcmath`

PHP7.2:
```php
<?php

var_dump(bcmul('-1', '-1', 10)); // => 1
```
PHP7.3+:
```php
<?php

var_dump(bcmul('-1', '-1', 10)); // => 1.0000000000
```

quasilyte:
This is consistent with how PHP7.3+ `bcmul` operates.

Fixes #206


AlexK0 (approved):

unserialize:
Marked WIP to highlight, that it can't be merged now due to PHP problems

runtime/bcmath.cpp
```diff
@@ -326,7 +326,7 @@ static string bc_mul_positive(const char *lhs, int lint, int ldot, int lfrac, in
 
   dl::deallocate(res, static_cast<size_t>(sizeof(int) * result_size));
 
-  return bc_round(result.buffer(), resint, resdot, resfrac, resscale, scale, sign, 0);
+  return bc_round(result.buffer(), resint, resdot, resfrac, resscale, scale, sign, 1);
```
unserialize, line 329: In `bc_div_positive` this is left 0, as you've changed to 1 only here, in `bc_mul_positive`. Shouldn't div also be modified? (I mean, we need to nest how PHP works)

unserialize (approved):

unserialize (approved):

PR #208, opened 2021-04-22
Job workers: graceful termination

DrDet:
Here is graceful termination logic for job workers. Now master process waits for termination of all job workers as well as all HTTP workers on graceful shutdown / restart.
Note: job workers are always terminated strictly after all HTTP workers.

### DIFF
--- a/server/php-master.cpp
+++ b/server/php-master.cpp
@@ -1617,8 +1617,12 @@ void run_master_off_in_graceful_shutdown() {
   assert(state == master_state::off_in_graceful_shutdown);
   to_kill = me_running_http_workers_n;
   if (all_http_workers_killed()) {
-    to_exit = 1;
-    me->is_alive = false;
+    if (all_job_workers_killed()) {
+      to_exit = 1;
+      me->is_alive = false;
+    } else {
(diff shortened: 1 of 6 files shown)


server/php-master.cpp
```diff
@@ -763,12 +763,13 @@ int run_worker(WorkerType worker_type) {
     switch (worker_type) {
       case WorkerType::http_worker:
         run_mode = RunMode::http_worker;
+        prctl(PR_SET_PDEATHSIG, SIGKILL); // TODO: or SIGTERM
         break;
       case WorkerType::job_worker:
         run_mode = RunMode::job_worker;
+        prctl(PR_SET_PDEATHSIG, SIGTERM);
```
AlexK0, line 770: I'm afraid, but the master process should be terminated last and after him no processes should left.
1) Some stuff are destroyed on master process shutdown 
2) Also, if a job worker is not terminated, it continues working as a zombie processes

I suggest using `prctl(PR_SET_PDEATHSIG, SIGKILL);` for all child and processing termination from master process.

DrDet: Agree, fixed

AlexK0 (approved):

PR #209, opened 2021-04-23
Implement null coalesce via class.

AlexK0:
Hi!
This patch fixes the code generation of null coalesce:
before:
```
$x = null_coalesce($a[$b], $key, $fallback_lambda);
```
after
```
$x = NullCoalesce($a[$b], $key).finalize($fallback_lambda);
```
so now, `$a[$b]` is not held on `$fallback_labmda` call.

Also I've used a perfect forwarding for basic cases.

### DIFF
--- a/runtime/null_coalesce.h
+++ b/runtime/null_coalesce.h
@@ -8,79 +8,131 @@
 #include "runtime/include.h"
 
 namespace impl_ {
+
 template<class ReturnType, class FallbackType>
-auto perform_fallback_impl(const FallbackType &lambda_fallback,
-                                 std::enable_if_t<bool(sizeof((std::declval<FallbackType>()(), 0)))> *) noexcept {
+auto perform_fallback_impl(FallbackType &&lambda_fallback,
+                           std::enable_if_t<bool(sizeof((std::declval<FallbackType>()(), 0)))> *) noexcept {
   return ReturnType(lambda_fallback());
 }
 
 template<class ReturnType, class FallbackType>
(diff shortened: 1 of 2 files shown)


runtime/null_coalesce.h
```diff
@@ -8,79 +8,131 @@
...
-ReturnType null_coalesce(const mixed &v, const string &string_key, int64_t precomuted_hash, const FallbackType &fallback) noexcept {
-  return v.is_string()
-         ? null_coalesce<ReturnType>(v.as_string(), string_key, fallback)
-         : null_coalesce<ReturnType>(v.get_value(string_key, precomuted_hash), fallback);
-}
+template<typename ResultType>
+struct NullCoalesce {
+public:
+  template<class ValueType>
+  NullCoalesce(ValueType &&value) noexcept {
```
unserialize, line 34: explicit?

runtime/null_coalesce.h
```diff
@@ -8,79 +8,131 @@
...
+  }
+
+  template<class FallbackType>
+  ResultType finalize(FallbackType &&fallback) noexcept {
+    return result_ ? std::move(*result_) : impl_::perform_fallback<ResultType>(std::forward<FallbackType>(fallback));
+  }
+
+  ~NullCoalesce() noexcept {
+    if (result_) {
+      result_->~ResultType();
```
unserialize, line 81: Shouldn't we assign `result_ = nullptr` here?

AlexK0: No, this is a destructor, therefore result_ can't be used after it.

unserialize (approved):

PR #212, opened 2021-04-26
parse inherited __construct phpdoc with base method context

### LINKED ISSUE #210: Class name resolving error (when namespace and class names collide?)
Reproducer:

https://github.com/quasilyte/kphp-compile-errors/tree/master/namespace_resolving

```
git clone https://github.com/quasilyte/kphp-compile-errors.git
cd kphp-compile-errors/namespace_resolving
composer install
php -f main.php # works fine
kphp --composer-root $(pwd) main.php # gives a compile-time error
```

Tried without composer as well; it also fails.

Error message:
```
Compilation error at stage: Apply phpdocs, gen by phpdoc.cpp:627
  lib/BinaryExpr/BinaryPlus.php:5  in Ast\BinaryExpr\BinaryPlus::__construct
    use Ast\BinaryExpr;

Could not find class Ast\BinaryExpr\Expr


Compilation error at stage: Apply phpdocs, gen by parse-and-apply-phpdoc.cpp:75
  lib/BinaryExpr/BinaryPlus.php:5  in Ast\BinaryExpr\BinaryPlus::__construct
    use Ast\BinaryExpr;

Failed to parse @param of Ast\BinaryExpr\BinaryPlus::__construct


Compilation error at stage: Apply phpdocs, gen by phpdoc.cpp:627
  lib/BinaryExpr/BinaryPlus.php:5  in Ast\BinaryExpr\BinaryPlus::__construct
    use Ast\BinaryExpr;

Could not find class Ast\BinaryExpr\Expr


Compilation error at stage: Apply phpdocs, gen by parse-and-apply-phpdoc.cpp:75
  lib/BinaryExpr/BinaryPlus.php:5  in Ast\BinaryExpr\BinaryPlus::__construct
    use Ast\BinaryExpr;

Failed to parse @param of Ast\BinaryExpr\BinaryPlus::__construct

```

quasilyte:
Imagine this case:

	// src/Foo/Base.php
	namespace Foo;
	class Base {
		/** @param T */
		public function __construct($x) {}
	}

	// src/Foo/Bar/Derived.php
	namespace Foo/Bar;
	class Derived extends \Foo\Base {}

The `Derived` class will get its constructor from the `Base` class.

The `Base::__construct` phpdoc mentions `T` class, which would
resolve into `Foo\T` if we were resolving it inside `Base::__construct` context.

Until this fix, derived constructor phpdocs were parsed in the derived method
context, so in the example below we'll get `\Foo\Bar\T` type for the `$x` param.

Fixes #210


compiler/data/class-data.h
```diff
@@ -58,6 +58,7 @@ class ClassData : public Lockable {
   bool really_used{false};
   bool is_tl_class{false};
   bool has_custom_constructor{false};
+  bool has_inherited_constructor{false}; // set to true if derived class gets its constructor from the base class
```
unserialize, line 61: To be honest, I suppose, that this field can be avoided.
FunctionData has the `is_auto_inherited` field. Maybe, play around that field for constructors? Instead of adding a new one to ClassData

quasilyte: Good idea. Fixed.

unserialize (approved):

AlexK0 (approved):

PR #219, opened 2021-04-27
enable instanceof smart casts for unions

### LINKED ISSUE #217: Consider supporting a union of related types as instanceof operand
```php
<?php

class Base {}

class Derived1 extends Base {}
class Derived2 extends Base {}

/** @param Derived1|Derived2 $x */
function f($x) {
  if ($x instanceof Derived1) {
    return 'derived1';
  }
  return 'derived2';
}

var_dump(f(new Derived1()));
var_dump(f(new Derived2()));
```
```
Compilation error at stage: Trasform To Smart Instanceof, gen by transform-to-smart-instanceof.cpp:18
  hello.php:10  in f
    if ($x instanceof Derived1) {

left operand of `instanceof` should be instance of class
```

We already infer `f()` param as `Base`; it should be possible to use `instanceof` as well, like it someone wrote `@param Base $x` instead of `Derived1|Derived2`.

quasilyte:
Teach extract_instance_from_type_hint() how to handle union types.
The algorithm used is close to the one that is used during the type inference.

We try to find a common base type for all union variants.
If there are multiple candidates, we'll give up.

This is good enough for most sane class hierarchies.

Fixes #217

### DIFF
--- a/compiler/class-assumptions.cpp
+++ b/compiler/class-assumptions.cpp
@@ -114,6 +114,56 @@ ClassPtr Assumption::extract_instance_from_type_hint(const TypeHint *a) {
     a = as_optional->inner;
   }
 
+  if (const auto *as_pipe = a->try_as<TypeHintPipe>()) {
+    kphp_assert(as_pipe->items.size() >= 2);
+    // to avoid the redundant work, check whether all union variants are classes
+    // and that they have some parent or interface (otherwise they would never
+    // have a common base)
+    for (const auto *item : as_pipe->items) {
+      if (auto as_primitive = item->try_as<TypeHintPrimitive>()) {
+        if (as_primitive->ptype != tp_Null) {
+          return ClassPtr{};
+        }
+        continue; // allow cases like `A|B|null`
+      }
+      const auto *as_instance = item->try_as<TypeHintInstance>();
+      if (!as_instance) {
+        return ClassPtr{};
+      }
+      ClassPtr klass = as_instance->resolve();
+      if (!klass) {
+        return ClassPtr{};
+      }
+      if (!klass->parent_class && klass->implements.empty()) {
+        return ClassPtr{};
+      }
+    }
+
+    // now we know that there is a chance to find something in common
+    // between them, so we can call a more expensive get_common_base_or_interface() method here;
+    // the method below is somewhat conservative, but it's good enough in the usual cases
+    ClassPtr result;
+    ClassPtr first = as_pipe->items[0]->try_as<TypeHintInstance>()->resolve();
+    for (int i = 1; i < as_pipe->items.size(); ++i) {
+      if (as_pipe->items[i]->try_as<TypeHintPrimitive>()) {
+        continue; // we know that it's tp_Null
+      }
+      ClassPtr other = as_pipe->items[i]->try_as<TypeHintInstance>()->resolve();
+      const auto common_bases = first->get_common_base_or_interface(other);
+      if (common_bases.size() != 1) {
+        return ClassPtr{};
+      }
+      if (!result || common_bases[0]->is_parent_of(result)) {
+        result = common_bases[0];
+      } else if (result == common_bases[0] || result->is_parent_of(common_bases[0])) {
+        // OK: still compatible.
+      } else {
+        return ClassPtr{};
+      }
+    }
+    return result;
+  }
+
   if (const auto *as_instance = a->try_as<TypeHintInstance>()) {
     return as_instance->resolve();
   }
--- a/tests/phpt/smart_casts/05_unions.php
+++ b/tests/phpt/smart_casts/05_unions.php
@@ -1,0 +1,123 @@
+@ok
+<?php
+
+class Expr {}
+
+class UnaryOp extends Expr {
+  public $arg = null;
+
+  public function __construct(Expr $arg) {
+    $this->arg = $arg;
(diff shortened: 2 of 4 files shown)


compiler/class-assumptions.cpp
```diff
@@ -114,6 +114,46 @@ ClassPtr Assumption::extract_instance_from_type_hint(const TypeHint *a) {
...
+    // and that they have some parent or interface (otherwise they would never
+    // have a common base)
+    for (const auto *item : as_pipe->items) {
+      const auto *as_instance = item->try_as<TypeHintInstance>();
+      if (!as_instance) {
+        return ClassPtr{};
+      }
+      ClassPtr klass = G->get_class(as_instance->full_class_name);
+      if (!klass) {
+          return ClassPtr{};
```
unserialize, line 129: formatting

compiler/class-assumptions.cpp
```diff
@@ -114,6 +114,46 @@ ClassPtr Assumption::extract_instance_from_type_hint(const TypeHint *a) {
...
+    kphp_assert(as_pipe->items.size() >= 2);
+    // to avoid the redundant work, check whether all union variants are classes
+    // and that they have some parent or interface (otherwise they would never
+    // have a common base)
+    for (const auto *item : as_pipe->items) {
+      const auto *as_instance = item->try_as<TypeHintInstance>();
+      if (!as_instance) {
+        return ClassPtr{};
+      }
+      ClassPtr klass = G->get_class(as_instance->full_class_name);
```
unserialize, line 127: as_instance->resolve() ?

compiler/class-assumptions.cpp
```diff
@@ -114,6 +114,46 @@ ClassPtr Assumption::extract_instance_from_type_hint(const TypeHint *a) {
...
+      }
+      if (!klass->parent_class && klass->implements.empty()) {
+        return ClassPtr{};
+      }
+    }
+    // now we know that there is a chance to find something in common
+    // between them, so we can call a more expensive get_common_base_or_interface() method here;
+    // the method below is somewhat conservative, but it's good enough in the usual cases
+    ClassPtr result;
+    ClassPtr first = G->get_class(as_pipe->items[0]->try_as<TypeHintInstance>()->full_class_name);
```
unserialize, line 139: resolve() also

compiler/class-assumptions.cpp
```diff
@@ -114,6 +114,46 @@ ClassPtr Assumption::extract_instance_from_type_hint(const TypeHint *a) {
...
+    for (int i = 1; i < as_pipe->items.size(); ++i) {
+      ClassPtr other = G->get_class(as_pipe->items[i]->try_as<TypeHintInstance>()->full_class_name);
```
unserialize, line 141: resolve() also

unserialize (dismissed):

tests/phpt/smart_casts/05_unions.php
```diff
@@ -0,0 +1,123 @@
...
+}
+
+/** @param UnaryMinus|Literal|UnaryPlus $x */
+function test6($x) {
+  if ($x instanceof UnaryPlus) {
+    var_dump(['unary plus' => get_class($x->arg)]);
+  } else if ($x instanceof StringLiteral) {
+    var_dump($x->string_value);
+  } else if ($x instanceof IntLiteral) {
+    var_dump($x->int_value);} else if ($x instanceof UnaryMinus) {
```
AlexK0, line 80: formatting

compiler/class-assumptions.cpp
```diff
@@ -114,6 +114,46 @@ ClassPtr Assumption::extract_instance_from_type_hint(const TypeHint *a) {
     a = as_optional->inner;
   }
 
+  if (const auto *as_pipe = a->try_as<TypeHintPipe>()) {
+    kphp_assert(as_pipe->items.size() >= 2);
+    // to avoid the redundant work, check whether all union variants are classes
+    // and that they have some parent or interface (otherwise they would never
+    // have a common base)
+    for (const auto *item : as_pipe->items) {
+      const auto *as_instance = item->try_as<TypeHintInstance>();
```
AlexK0, line 123: may be it's worth to check TypeHintOptional case as well, so the following example would work

```php
<?php

class A {
  public $a = 1;
}

class A1 extends A {
  public $a1 = 1;
}

class A2 extends A {
  public $a2 = 1;
}

/** @param A1|A2|null $x */
function foo($x) {
  if ($x instanceof A1) {
    var_dump($x->a);
  }
}

foo(new A1);
foo(new A2);
```

unserialize: `A1|A2|null` is a pipe. TypeHintOptional is T|null (a single T, not many)

tests/phpt/smart_casts/05_unions.php
quasilyte: Fixed. Thanks for noticing. It's actually not only a formatting issue, it was a redundant branch that made another one unreachable.

compiler/class-assumptions.cpp
quasilyte: I made `A1|A2|null` work by skipping `null` types during the common base type searching.
It doesn't make code much more complicated but looks quite useful. :+1:

unserialize (approved):

AlexK0 (approved):

PR #226, opened 2021-04-29
rewrite ClassData::get_instance_method()

### LINKED ISSUE #215: Interface method is not implemented via multi-level abstract class inheritance
This don't work:
```php
<?php

interface Iface {
  function f();
}

abstract class AbstractBase {
  public function f() {}
}

abstract class AbstractBase2 extends AbstractBase {}

class Concrete extends AbstractBase2 implements Iface {}

$c = new Concrete();
```

This, hovewer, does work:
```php
<?php

interface Iface {
  function f();
}

abstract class AbstractBase {
  public function f() {}
}

class Concrete extends AbstractBase implements Iface {}

$c = new Concrete();
```

Diff:
```diff
- class Concrete extends AbstractBase2 implements Iface {}
+ class Concrete extends AbstractBase implements Iface {}
```
Error message:
```
Compilation error at stage: Generate virtual methods of interfaces/base classes, gen by virtual-method-generator.cpp:122
  hello.php:0  in Iface::f

You should override abstract method: `Iface::f` in class: `Concrete`
```
Could be a regression.
This deserves a dedicated test when this issue is fixed (so we don't break it again).

quasilyte:
Previously, we used this approach:

* Collect all ancestors into `std::vector` (interfaces, base classes)
* Iterate over them, return the first member match

What we get wrong with this approach is that we could get an abstract method
instead of the implementation as the order of `get_all_ancestors()` doesn't
prefer classes to interfaces.

Consider this case:

```php
interface Iface { function f(); }

// Actually implements Iface, but doesn't state that
abstract class AbstractBase {
  public function f() {}
}

abstract class AbstractBase2 extends AbstractBase {}

// Implements Iface via AbstractBase, which is a base class of AbstractBase2
class Concrete extends AbstractBase2 implements Iface {}
```

With the old approach, we'll get a compile-time error.

The new approach prefers non-abstract methods; it'll traverse the classes and
try to find the first non-abstract method. If it's not possible, it will
return the first abstract method (like we did before). The code is slightly
more complicated than that just to avoid redundant work.

As we do more (useful) work now and sometimes need to check more classes,
I tried to write this method so it's not slower than before.

	old method:
	| Instance method lookup | 365669 | 3.478 sec | 6.370 sec | - | +6.3 MB |
	new method:
	| Instance method lookup | 365669 | 3.256 sec | 7.154 sec | - |       - |

Note: locking is applied in the same fashion as in find_by_local_name, per `.members` lookup.

Fixes #215

### DIFF
--- a/compiler/data/class-data.cpp
+++ b/compiler/data/class-data.cpp
@@ -247,8 +247,67 @@ std::vector<ClassPtr> ClassData::get_common_base_or_interface(ClassPtr other) co
   return lcas;
 }
 
+static const ClassMemberInstanceMethod *find_method_in_interface(InterfacePtr interface, vk::string_view local_name) {
+  {
+    AutoLocker<Lockable*> locker(&(*interface));
+    if (const auto *member = interface->members.find_by_local_name<ClassMemberInstanceMethod>(local_name)) {
+      return member;
+    }
+  }
+
+  // interfaces can extend multiple other interfaces; these interfaces are stored in 'implements' member
+  for (InterfacePtr base_interface : interface->implements) {
+    if (const auto *member = find_method_in_interface(base_interface, local_name)) {
+      return member;
+    }
+  }
+  return nullptr;
+}
+
+static const ClassMemberInstanceMethod *find_method_in_interface_list(const std::vector<InterfacePtr> &interfaces, vk::string_view local_name) {
+  for (InterfacePtr interface : interfaces) {
+    if (const auto *member = find_method_in_interface(interface, local_name)) {
+      return member;
+    }
+  }
+  return nullptr;
+}
+
+const ClassMemberInstanceMethod *ClassData::find_instance_method_by_local_name(vk::string_view local_name) const {
+  const ClassMemberInstanceMethod *result{nullptr};
+
+  ClassPtr current = get_self();
+  while (true) {
+    const auto *member = ({
+      AutoLocker<Lockable*> locker(&(*current));
+      current->members.find_by_local_name<ClassMemberInstanceMethod>(local_name);
+    });
+
+    if (member) {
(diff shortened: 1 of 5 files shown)


compiler/data/class-data.cpp
```diff
@@ -246,8 +246,67 @@ std::vector<ClassPtr> ClassData::get_common_base_or_interface(ClassPtr other) co
...
+
+const ClassMemberInstanceMethod *ClassData::find_method_by_local_name(vk::string_view local_name) const {
+  const ClassMemberInstanceMethod *result{nullptr};
+
+  ClassPtr current = get_self();
+  while (true) {
+    const auto *member = ({
+      AutoLocker<Lockable*> locker(&(*current));
+      current->members.find_by_local_name<ClassMemberInstanceMethod>(local_name);
+    });
```
quasilyte, line 283: Could use something simpler here, like:
```c++
    current->lock();
    const auto *member = current->members.find_by_local_name<ClassMemberInstanceMethod>(local_name);
    current->unlock();
```

compiler/data/class-data.h
```diff
@@ -187,6 +187,8 @@ class ClassData : public Lockable {
 private:
   bool has_polymorphic_member_dfs(std::unordered_set<ClassPtr> &checked) const;
 
+  const ClassMemberInstanceMethod *find_method_by_local_name(vk::string_view local_name) const;
```
unserialize, line 190: find_instance_method_by_local_name?

unserialize (dismissed):

compiler/data/class-data.h
quasilyte: Done.

unserialize (approved):

PR #235, opened 2021-05-17
Use default malloc instead of script memory for curl.

AlexK0:
The full curl cleanup after each request leads memory leak via openssl (from curl internals).

This patch is about the default malloc usage instead of script memory allocator for the curl internals. It allows to avoid of the curl full cleanup after each request.


runtime/string-list.h
```diff
@@ -0,0 +1,80 @@
+// Compiler for PHP (aka KPHP)
+// Copyright (c) 2020 LLC «V Kontakte»
+// Distributed under the GPL v3 License, see LICENSE.notice.txt
+
+#pragma once
+
+#include "runtime/allocator.h"
+#include "runtime/kphp_core.h"
+
+class string_list : vk::not_copyable {
```
unserialize, line 10: Is it `string_concat` actually?

unserialize: I slightly suggested renaming, but don't care, let's leave it so

DrDet (approved):

PR #236, opened 2021-05-18
added initial support for the `__toString()` magic method

i582:
# PR introducing support for the magic method `__toString()`

## Principle

To support the `__toString()` method, now, each `strval($expr)` 
expression, where `$expr` has a class type and a class has a 
`__toString()` method is replaced with `$expr->__toString()`.

String and class conversions are now allowed in 
`CheckConversionsPass::forbidden_conversions`. If the class 
does not contain the `__toString()` method, then an error is 
thrown in `OptimizationPass::try_convert_expr_to_call_to_string_method`.

## Concessions

Since the actual replacement occurs after the stage of 
type inference, and the stage of removing unused 
functions go before type inference, it is impossible 
to determine the reachability of the `__toString` function 
at the stage of removing unused functions, so this 
function will **always be generated**.


compiler/pipes/optimization.cpp
```diff
@@ -218,6 +218,50 @@ VertexPtr OptimizationPass::remove_extra_conversions(VertexPtr root) {
...
+  if (type == nullptr) {
+    return {};
+  }
+
+  const auto klass = type->class_type();
+  if (!klass) {
+    return {};
+  }
+
+  if (!klass->has_tostring) {
```
AlexK0, line 236: It seems to me, that we don't need this flag, because we can use a result of `get_instance_method`:
```cpp
const auto *to_string_method = klass->get_instance_method(ClassData::NAME_OF_TO_STRING);
if (!to_string_method) {
  return error;
}
  ```

File: compiler/pipes/filter-only-actually-used.cpp [hunk 2]
```diff
@@ -171,7 +171,8 @@ IdMap<FunctionPtr> calc_actually_used_having_call_edges(std::vector<FunctionAndE
       fun->is_main_function() ||
       fun->type == FunctionData::func_class_holder || // classes should be carried along the pipeline
       (fun->is_extern() && vk::any_of_equal(fun->name, "wait", "make_clone")) ||
-      fun->kphp_lib_export;
+      fun->kphp_lib_export ||
+      vk::ends_with(fun->name, ClassData::NAME_OF_TO_STRING);
```
AlexK0, line 175: Can we use assumptions instead of type inferring? If we use the assumptions we can move conv replacements before this pipe and filter unused `__toString()` methods. Have you thought about it?

compiler/pipes/check-classes.cpp
```diff
@@ -131,6 +132,28 @@ void CheckClassesPass::check_serialized_fields(ClassPtr klass) {
...
+
+  const auto count_args = fun->param_ids.size();
+  kphp_error(count_args == 1, fmt_format("Magic method {} cannot take arguments", fun->get_human_readable_name()));
+
+  const auto *ret_type = tinf::get_type(fun, -1);
+  if (!ret_type) {
+    return;
+  }
+
+  kphp_error(ret_type->ptype() == tp_string, fmt_format("Magic method {} must have string return type", fun->get_human_readable_name()));
```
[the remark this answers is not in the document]
i582: Fixed.

compiler/pipes/check-conversions.cpp
```diff
@@ -12,7 +13,7 @@ const std::multimap<Operation, PrimitiveType> CheckConversionsPass::forbidden_co
   {op_conv_int,      tp_Class},
   {op_conv_int_l,    tp_Class},
   {op_conv_float,    tp_Class},
-  {op_conv_string,   tp_Class},
+  // {op_conv_string,   tp_Class}, // allowed for classes with __toString()
```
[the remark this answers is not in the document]
i582: Removed.

compiler/pipes/check-conversions.cpp
```diff
@@ -3,6 +3,7 @@
 // Distributed under the GPL v3 License, see LICENSE.notice.txt
 
 #include "compiler/pipes/check-conversions.h"
+#include "compiler/data/class-data.h"
```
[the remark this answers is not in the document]
i582: No.
Removed.

compiler/pipes/optimization.cpp
```diff
@@ -253,6 +297,14 @@ VertexPtr OptimizationPass::on_enter_vertex(VertexPtr root) {
         explicit_cast_array_type(args[index], tinf::get_type(params[index]), &current_function->explicit_const_var_ids);
       }
     }
+    if (func->name == "print_r") {
```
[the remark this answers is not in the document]
i582: Removed.

compiler/pipes/optimization.cpp
i582: Removed.

compiler/data/class-members.cpp
```diff
@@ -172,6 +172,10 @@ void ClassMembersContainer::add_instance_method(FunctionPtr function) {
   if (vk::string_view(function->name).ends_with(ClassData::NAME_OF_CONSTRUCT)) {
     klass->construct_function = function;
   }
+
+  if (vk::string_view(function->name).ends_with(ClassData::NAME_OF_TO_STRING)) {
```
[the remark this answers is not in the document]
i582: Since the `has_tostring` flag has been removed, this piece of code no longer exists.

compiler/pipes/optimization.cpp
```diff
@@ -218,6 +218,50 @@ VertexPtr OptimizationPass::remove_extra_conversions(VertexPtr root) {
   return root;
 }
 
+// The function will try to convert the expression to a call to the
+// __toString method if the expression has a class type and the class
```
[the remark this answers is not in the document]
i582: Fixed.

compiler/pipes/check-classes.cpp
```diff
@@ -131,6 +132,28 @@ void CheckClassesPass::check_serialized_fields(ClassPtr klass) {
   });
 }
 
+void CheckClassesPass::check_magic_methods(ClassPtr klass) {
+  klass->members.for_each([&](const ClassMemberInstanceMethod &f) {
+    if (vk::ends_with(f.function->name, ClassData::NAME_OF_TO_STRING)) {
```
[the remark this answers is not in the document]
i582: `And because there are a lot of operations to do for `local_name()`, 
we will not slow down much because of the check for each method? 
Suffix checking should be much faster.

compiler/pipes/check-classes.cpp
```diff
@@ -131,6 +132,28 @@ void CheckClassesPass::check_serialized_fields(ClassPtr klass) {
...
+      check_magic_tostring_method(f.function);
+    }
+  });
+}
+
+void CheckClassesPass::check_magic_tostring_method(FunctionPtr fun) {
+  stage::set_function(fun);
```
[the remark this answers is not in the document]
i582: Moved to FinalCheckPass.

compiler/pipes/check-classes.cpp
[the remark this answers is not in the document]
i582: Since checking magic methods has been moved to 
`FinalCheckPass`, this piece of code no longer exists.

compiler/pipes/filter-only-actually-used.cpp
i582: Since it was decided during the discussion that it was 
too costly for this task, we leave the version as it is.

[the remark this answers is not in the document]
i582: Done.

compiler/pipes/final-check.cpp
```diff
@@ -587,3 +584,37 @@ void FinalCheckPass::raise_error_using_Unknown_type(VertexPtr v) {
...
+
+  if (name == ClassData::NAME_OF_TO_STRING) {
+    check_magic_tostring_method(fun);
+  } else if (name == ClassData::NAME_OF_CLONE) {
+    check_magic_clone_method(fun);
+  }
+}
+
+void FinalCheckPass::check_magic_tostring_method(FunctionPtr fun) {
+  stage::set_function(fun);
```
AlexK0, line 603: No need to do this. `set_function()` for `current_function` performs implicitly.

compiler/pipes/optimization.cpp
```diff
@@ -134,6 +134,10 @@ void OptimizationPass::collect_concat(VertexPtr root, vector<VertexPtr> *collect
       collect_concat(i, collected);
     }
   } else {
+    const auto call = try_convert_expr_to_call_to_string_method(root);
+    if (call) {
```
AlexK0, line 138: nit:
```cpp
if (const auto call = try_convert_expr_to_call_to_string_method(root)) {
  root = call;
}
```
same below.

tests/phpt/cl/190_toString_magic_method_nullable_return_type.php
```diff
@@ -0,0 +1,15 @@
...
+  function __toString() : ?string {
+    return null;
+  }
+}
+
+function demo(A $a) {
+  echo $a;
+}
+
+demo(new A);
```
AlexK0, line 15: new line

i582: Fixed.

compiler/pipes/final-check.cpp
i582: Fixed.

compiler/pipes/optimization.cpp
i582: Fixed.

AlexK0 (dismissed):

compiler/pipes/check-conversions.cpp
```diff
@@ -3,16 +3,12 @@
 // Distributed under the GPL v3 License, see LICENSE.notice.txt
 
 #include "compiler/pipes/check-conversions.h"
-
-#include "common/termformat/termformat.h"
-
```
unserialize, line 8: Please, leave one \n here (we separate including this-file.h from other includes)

compiler/pipes/filter-only-actually-used.cpp
```diff
@@ -171,7 +171,8 @@ IdMap<FunctionPtr> calc_actually_used_having_call_edges(std::vector<FunctionAndE
       fun->is_main_function() ||
       fun->type == FunctionData::func_class_holder || // classes should be carried along the pipeline
       (fun->is_extern() && vk::any_of_equal(fun->name, "wait", "make_clone")) ||
-      fun->kphp_lib_export;
+      fun->kphp_lib_export ||
+      fun->local_name() == ClassData::NAME_OF_TO_STRING;
```
unserialize, line 175: What I don't like here, is that every function will be searched for the "$$" substring (inside `local_name()` call).
Maybe, prevent this at least by doing in such a way:
```
(fun->modifiers.is_instance() && fun->local_name() == ClassData::NAME_OF_TO_STRING)
```

compiler/pipes/final-check.cpp
```diff
@@ -587,3 +584,35 @@ void FinalCheckPass::raise_error_using_Unknown_type(VertexPtr v) {
...
+    return;
+  }
+
+  kphp_error(ret_type->ptype() == tp_string && ret_type->flags() == 0,
+             fmt_format("Magic method {} must have string return type", fun->get_human_readable_name()));
+}
+
+void FinalCheckPass::check_magic_clone_method(FunctionPtr fun) {
+  kphp_error(!fun->is_resumable, fmt_format("{} method has to be not resumable", ClassData::NAME_OF_CLONE));
+  kphp_error(!fun->can_throw(), fmt_format("{} method should not throw exception", ClassData::NAME_OF_CLONE));
```
unserialize, line 617: Let's use `fun->get_human_readable_name()` here instead of `NAME_OF_CLONE`

compiler/pipes/check-conversions.cpp
```diff
@@ -3,16 +3,12 @@
...
 #include "compiler/inferring/public.h"
 
 const std::multimap<Operation, PrimitiveType> CheckConversionsPass::forbidden_conversions = {
   {op_conv_int,      tp_Class},
   {op_conv_int_l,    tp_Class},
   {op_conv_float,    tp_Class},
-  {op_conv_string,   tp_Class},
```
unserialize, line 15: Don't you want to comment it out, leaving a hint that it's commented out intentionally?

tests/phpt/cl/191_toString_magic_method_string_builder.php
```diff
@@ -0,0 +1,14 @@
...
+<?php
+
+class A {
+  function __toString() {
+    return "hello";
+  }
+}
+
+function demo(A $a) {
+  echo "$a and $a";
```
unserialize, line 11: Let's extend the testcase with the following:
```
// class A
public function getCopy() : A { return $this; }
public function getArray: array { return [$this]; }

echo "{$a->getCopy()} {$a->getArray()[0]}";
```
(we want to test, that arbitrary expressions work inside string builds)

compiler/pipes/check-conversions.cpp
i582: Fixed.

compiler/pipes/filter-only-actually-used.cpp
i582: Done.

compiler/pipes/check-conversions.cpp
i582: Added.

tests/phpt/cl/191_toString_magic_method_string_builder.php
i582: Added.

compiler/pipes/final-check.cpp
i582: Fixed.

AlexK0 (approved):

PR #240, opened 2021-05-20
Job workers: enhancements

DrDet:
Here are the following enhancements for job workers:
1. Configuring epoll to wake up only one worker on event, instead of all
2. Fix job shared memory leak, when worker terminates before script is inited
3. Implement `kphp_job_workers_count()` function
4. Other minor improvements

### DIFF
--- a/server/job-workers/job-worker-server.cpp
+++ b/server/job-workers/job-worker-server.cpp
@@ -175,7 +175,7 @@ void JobWorkerServer::init() {
     clear_event(result_pipe[0]);
   }
 
-  read_job_connection = epoll_insert_pipe(pipe_for_read, read_job_fd, &php_jobs_server, nullptr);
+  read_job_connection = epoll_insert_pipe(pipe_for_read, read_job_fd, &php_jobs_server, nullptr, EVT_LEVEL);
   assert(read_job_connection);
   memset(read_job_connection->custom_data, 0, sizeof(read_job_connection->custom_data));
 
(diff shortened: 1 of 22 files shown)


server/job-workers/job-worker-server.cpp
```diff
@@ -175,7 +175,7 @@ void JobWorkerServer::init() {
     clear_event(result_pipe[0]);
   }
 
-  read_job_connection = epoll_insert_pipe(pipe_for_read, read_job_fd, &php_jobs_server, nullptr);
+  read_job_connection = epoll_insert_pipe(pipe_for_read, read_job_fd, &php_jobs_server, nullptr, EVT_LEVEL);
```
AlexK0, line 178: Please, add the following test case:
setup: 2 job workers + N http workers (may be 4, to be sure that job messages would be enough)
use instance_cache for synchronization

Http worker:
1) starts 2 sync jobs
2) reads in while from instance_cache to be sure, that jobs are started 
3) starts N (>=2, 3 is fine) simple jobs
4) writes something into instance_cache (any special key), so sync job are going to finish
5) wait simple jobs reply
6) wait sync jobs reply

Job worker (SYNC):
1) writes something into isntance_cache (any special key)
2) read in while from instance_cache to be sure, that http worker has started simple jobs
3) finish

Job worker (simple): up to you, x^2 would be enough

AlexK0 (dismissed):

AlexK0 (approved):

PR #241, opened 2021-05-21
Fix negative keys bug

AlexK0:
This patch fixes a bug when negative keys lead a crash from runtime.
Also I've done some refactoring.

Here I've also moved shared memory copying process for primitive types into runtime lib, it compiles with -O3 optimization level, so it works faster now!

### DIFF
--- a/runtime/array_decl.inl
+++ b/runtime/array_decl.inl
@@ -433,7 +433,9 @@ private:
   template<class Arg, class...Args>
   void push_back_values(Arg &&arg, Args &&... args);
 
-private:
+  template<class T1>
+  void push_back_iterator(const array_iterator<T1> &it) noexcept;
+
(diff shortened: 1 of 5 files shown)


runtime/array_decl.inl
```diff
@@ -45,25 +45,39 @@ enum class overwrite_element {
...
+  uint32_t int_buf_size;
+  uint32_t string_size;
+  uint32_t string_buf_size;
+};
+
 template<class T>
 class array {
-
 public:
+  // TODO why we need 2 value types?
```
DrDet, line 71: `ValueType` is used in `TlDependentTypesUsings::deduce_params_from_type_tree()` 
`value_type` seems to be unused

DrDet (approved):

PR #242, opened 2021-05-23
add IntHash64 function

i582:
See polyfill here: https://github.com/VKCOM/kphp-polyfills/pull/20


i582:
We made a decision that it makes no sense to add an implementation 
to KPHP since it can be written in PHP and the transpiled version will be just as fast.

PR #247, opened 2021-05-28
added str_ireplace function

i582:
### Changes

- All functions for implementing `str_replace` now accept the `with_case` 
flag as the last parameter, which determines whether this function will 
be case insensitive.

- To search for a substring, a separate function `find_substr` has been 
  created, which, when case sensitive, does a quick search, as is done
  now with `memmem`, and when case insensitive, does `strcasestr`.

### DIFF
--- a/runtime/string_functions.cpp
+++ b/runtime/string_functions.cpp
@@ -1972,12 +1972,8 @@ static string str_replace_char(char c, const string &replace, const string &subj
     if (with_case) {
       pos = static_cast<const char *>(memchr(piece, c, piece_end - piece));
     } else {
-      for (int i = 0; i < piece_end - piece; ++i) {
-        if (std::toupper(piece[i], std::locale()) == std::toupper(c, std::locale())) {
-          pos = piece + i;
-          break;
-        }
-      }
+      const char needle[] = {c, '\0'};
+      pos = strcasestr(piece, needle);
     }
 
     if (pos == nullptr) {
@@ -2005,11 +2001,7 @@ static const char *find_substr(const char *where, const char *where_end, const s
     return static_cast<char *>(memmem(where, where_end - where, what.c_str(), what.size()));
   }
 
-  const auto *it = std::search(where, where_end, what.c_str(), what.c_str() + what.size(), [](char ch1, char ch2) {
-    return std::toupper(ch1, std::locale()) == std::toupper(ch2, std::locale());
-  });
-
-  return it == where_end ? nullptr : it;
+  return strcasestr(where, what.c_str());
 }
 
 void str_replace_inplace(const string &search, const string &replace, string &subject, int64_t &replace_count, bool with_case) {
--- a/runtime/string_functions.cpp
+++ b/runtime/string_functions.cpp
@@ -1957,15 +1957,29 @@ string f$str_repeat(const string &s, int64_t multiplier) {
   return result;
 }
 
-static string str_replace_char(char c, const string &replace, const string &subject, int64_t &replace_count) {
+static string str_replace_char(char c, const string &replace, const string &subject, int64_t &replace_count, bool with_case) {
   int count = 0;
-  const char *piece = subject.c_str(), *piece_end = subject.c_str() + subject.size();
+  const char *piece = subject.c_str();
+  const char *piece_end = subject.c_str() + subject.size();
+
   string result;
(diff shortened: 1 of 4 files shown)


runtime/string_functions.cpp
```diff
@@ -1982,24 +1996,39 @@ static string str_replace_char(char c, const string &replace, const string &subj
...
 }
 
-void str_replace_inplace(const string &search, const string &replace, string &subject, int64_t &replace_count) {
-  if ((int)search.size() == 0) {
+static const char *find_substr(const char *where, const char *where_end, const string &what, bool with_case) {
+  if (with_case) {
+    return static_cast<char *>(memmem(where, where_end - where, what.c_str(), what.size()));
+  }
+
+  const auto *it = std::search(where, where_end, what.c_str(), what.c_str() + what.size(), [](char ch1, char ch2) {
```
AlexK0, line 2008: just strcasestr()?

runtime/string_functions.cpp
```diff
@@ -1957,15 +1957,29 @@ string f$str_repeat(const string &s, int64_t multiplier) {
...
+    const char *pos = nullptr;
+    if (with_case) {
+      pos = static_cast<const char *>(memchr(piece, c, piece_end - piece));
+    } else {
+      for (int i = 0; i < piece_end - piece; ++i) {
+        if (std::toupper(piece[i], std::locale()) == std::toupper(c, std::locale())) {
+          pos = piece + i;
+          break;
+        }
+      }
```
DrDet, line 1980: Consider using `memchr` here too. You can perform it via 2 searches:
1. search for `c` in its original case using `memchr` (you've done this above)
2. search for `c` in inversed case using `memchr`
3. choose the minimal pos from two searches (don't forget to check on nullptr)

runtime/string_functions.cpp
```diff
@@ -1957,15 +1957,29 @@ string f$str_repeat(const string &s, int64_t multiplier) {
...
   string result;
   if (!replace.empty()) {
     result.reserve_at_least(subject.size());
   }
+
   while (true) {
-    const char *pos = (const char *)memchr(piece, c, piece_end - piece);
+    const char *pos = nullptr;
+    if (with_case) {
+      pos = static_cast<const char *>(memchr(piece, c, piece_end - piece));
```
DrDet, line 1973: Why do we need `memchr` here? Can we change it to `strchr`?

runtime/string_functions.cpp
i582: Hmm, yes, for sure, it will be better.

runtime/string_functions.cpp
i582: It seems we can.

runtime/string_functions.cpp
DrDet, line 1980: Consider doing it via `strcasestr` like this:
```suggestion
      const char needle[] = {c, '\0'};
      pos = strcasestr(piece, needle);
```

runtime/string_functions.cpp
DrDet: Then let's do it like this:
```suggestion
      pos = strchr(piece, c);
```

AlexK0: I've thought, that memchr() works a bit faster.

runtime/string_functions.cpp
i582: Hmm, it seems we cannot use `strcasestr`, because we need to search only in `[where, where_end]`.
Perhaps there is a version where we can limit the number of characters, but I did not find one.

i582: Done.

runtime/string_functions.cpp
i582: Done.

runtime/string_functions.cpp
i582: > I've thought, that memchr() works a bit faster.

My small benchmarks showed that up to 10,000,000 cycles, 
`strchr` works 62ms against 74ms, but by 100,000,000 cycles 
`strchr` is slower, 1013ms against 932ms for `memchr`.

It seems to me that this is not a big difference and we can 
take the `strchr` function simply because it looks simpler.

unserialize (approved):

PR #249, opened 2021-05-28
added check for indexing

i582:
For strings and tuples, `int`, `float`, `future`, and `mixed` types are allowed.
For others, `string`, `int`, `float`, `future`, and `mixed` types are allowed.

The `float` type is allowed because it is a fairly popular pattern to 
use the result of a function that returns a `float` and expects an 
implicit cast to an `int`.

The `future` type is also allowed, as it is a fairly common pattern.

Fixed crash if not a number or a string was passed to the tuple index.


compiler/pipes/final-check.cpp
```diff
@@ -408,10 +411,50 @@ VertexPtr FinalCheckPass::on_enter_vertex(VertexPtr vertex) {
...
+
+  const auto key_ptype = key_type->ptype();
+  const auto key_type_is_raw = !key_type->or_false_flag() && !key_type->or_null_flag();
+  bool is_allowed = false;
+  std::string allowed_types;
+  std::string what_indexing;
+
+  switch (array_type->ptype()) {
+    case tp_tuple:
+    case tp_string:
```
AlexK0, line 439: tp_shape?

compiler/pipes/final-check.cpp
```diff
@@ -408,10 +411,50 @@ VertexPtr FinalCheckPass::on_enter_vertex(VertexPtr vertex) {
...
 
+  if (vertex->type() == op_index) {
+    check_indexing(vertex);
+  }
+
   //TODO: may be this should be moved to tinf_check
   return vertex;
 }
 
+void FinalCheckPass::check_indexing(VertexPtr vertex) {
```
AlexK0, line 422: It wold be better to accept `VertexAdaptor<op_index>` here

compiler/pipes/final-check.cpp
```diff
@@ -408,10 +411,50 @@ VertexPtr FinalCheckPass::on_enter_vertex(VertexPtr vertex) {
...
+  const auto *array_type = tinf::get_type(array);
+  if (key_type == nullptr || array_type == nullptr) {
+    return;
+  }
+
+  const auto key_ptype = key_type->ptype();
+  const auto key_type_is_raw = !key_type->or_false_flag() && !key_type->or_null_flag();
+  bool is_allowed = false;
+  std::string allowed_types;
+  std::string what_indexing;
```
AlexK0, line 435: seems to me, that vk::string_view would be enough here (for bought cases).

compiler/pipes/final-check.cpp
```diff
@@ -408,10 +411,50 @@ VertexPtr FinalCheckPass::on_enter_vertex(VertexPtr vertex) {
...
+
+  switch (array_type->ptype()) {
+    case tp_tuple:
+    case tp_string:
+      is_allowed = vk::any_of_equal(key_ptype, tp_int, tp_float, tp_mixed);
+      allowed_types = "int and float";
+      what_indexing = array_type->ptype() == tp_string ? "string " : "tuple ";
```
AlexK0, line 442: ptype_name()?

compiler/pipes/final-check.cpp
```diff
@@ -408,10 +411,50 @@ VertexPtr FinalCheckPass::on_enter_vertex(VertexPtr vertex) {
     check_null_usage_in_binary_operations(binary_vertex);
   }
 
+  if (vertex->type() == op_index) {
+    check_indexing(vertex);
```
AlexK0, line 415: pls, consider more cases:
1) 
```php
<?php

class A {}

function demo() {
  $a = new A;
  $x = [$a => 1];
}

demo();
```

2)
```php
<?php

class A {}

function demo() {
  $x = [1];
  $x[new A] = 2;
}

demo();
```

compiler/pipes/final-check.cpp
i582: Are we not allowing strings when indexing shapes?

AlexK0: you have the special logic for tp_tuple, but you don't have it for tp_shape is it ok?

i582: Since the validation logic for shapes is the same as validation for all other types, does it make sense to put this in a separate case?

AlexK0: ouch, I see, ignore this comment

compiler/pipes/final-check.cpp
i582: Fixed.

compiler/pipes/final-check.cpp
i582: Fixed.

compiler/pipes/final-check.cpp
i582: Added.

compiler/pipes/final-check.cpp
```diff
@@ -408,10 +411,50 @@ VertexPtr FinalCheckPass::on_enter_vertex(VertexPtr vertex) {
...
+      what_indexing = array_type->ptype() == tp_string ? "string " : "tuple ";
+      break;
+
+    default:
+      is_allowed = vk::any_of_equal(key_ptype, tp_string, tp_int, tp_float, tp_mixed);
+      allowed_types = "int, string and float";
+      break;
+  }
+
+  if (!key_type_is_raw) {
```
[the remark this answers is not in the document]
i582: Allowed.

compiler/pipes/final-check.cpp
i582: Fixed.

PR #250, opened 2021-05-29
added optimization for (v)sprintf call with simple format

i582:
`(v)sprintf` calls with a simple format string are now converted to concatenation.

A simple format string is a string that only contains the `%s` and `%d` specifiers and 
any other text. If the string contains something else, then such a call will be left as it is.

Example:
```php
$name = "John";

echo sprintf("Hello %s", $name);
// converted to:
echo "Hello " . $name;
```

Depending on the length of the string and count specifiers, the speed of concatenation 
is several times faster. Even for the given example, the speed of `sprintf` is 4 times slower 
than that of concatenation.

This optimization was taken out of the optimization pipe in order to place it before the pipe, 
where the constants are moved into variables so that the constant strings created in this pipe 
are moved into variables


compiler/pipes/convert-sprintf-calls.cpp
```diff
@@ -0,0 +1,199 @@
...
+    parts.push_back(FormatPart{last_value});
+  }
+
+  return parts;
+}
+
+VertexPtr ConvertSprintfCallsPass::on_exit_vertex(VertexPtr root) {
+  if (auto func_call = root.try_as<op_func_call>()) {
+    const auto func = func_call->func_id;
+    if (vk::any_of_equal(func->name, "sprintf", "vsprintf")) {
```
AlexK0, line 73: What about vprintf, printf, vfprintf, fprintf? The solution this these functions looks the same.

compiler/pipes/convert-sprintf-calls.cpp
```diff
@@ -0,0 +1,199 @@
...
+  const auto args = call->args();
+  auto format_arg_raw = args[0];
+
+  if (format_arg_raw->type() == op_conv_string) {
+    format_arg_raw = format_arg_raw.as<op_conv_string>()->expr();
+  }
+
+  VertexAdaptor<op_string> format_string;
+
+  switch (format_arg_raw->type()) {
```
AlexK0, line 91: Please, consider using GenTree::get_constexpr_string() instead.

compiler/pipes/convert-sprintf-calls.cpp
```diff
@@ -0,0 +1,199 @@
...
+      } else if (symbol == '%') {
+        last_value += "%";
+      } else {
+        return {};
+      }
+      find_percent = false;
+    } else if (symbol == '%') {
+      find_percent = true;
+      if (!last_value.empty()) {
+        parts.push_back(FormatPart{last_value});
```
AlexK0, line 55: std::move(last_value) here and below

i582: Fixed.

compiler/pipes/convert-sprintf-calls.cpp
i582: Fixed.

compiler/pipes/convert-sprintf-calls.cpp
i582: It seems to me that this does not make much sense, since the speed for these functions is no critical, and their support will slightly complicate the code and increase the already large diff between versions.

AlexK0: but..
replace `f$printf($format, ...)` to `print(string_build)` function call
replace `f$fprintf($stream, $format, ...)` to `f$write($stream, string_build)` function call

But ok, it's up to you..

AlexK0: > f$printf($format, ...) can be replaced with print(string_build) function call
ok, this is not truth, `print` always returns 1.. 
anyway, I don't see any problem to add one more simple function, which prints and returns size..

i582: > > f$printf($format, ...) can be replaced with print(string_build) function call
> 
> ok, this is not truth, `print` always returns 1..
> anyway, I don't see any problem to add one more simple function, which prints and returns size..

Yep, this is one of the problems with replacement, because `printf` returns the number of characters printed.

You mean something like this:

```cpp
int f$print_with_size(string str) {
   f$print(str);
   return str.size();
}
```

i582: > > > f$printf($format, ...) can be replaced with print(string_build) function call
> > 
> > 
> > ok, this is not truth, `print` always returns 1..
> > anyway, I don't see any problem to add one more simple function, which prints and returns size..
> 
> Yep, this is one of the problems with replacement, because `printf` returns the number of characters printed.
> 
> Do you mean something like that in runtime?
> 
> ```c++
> int f$print_with_size(string str) {
>    f$print(str);
>    return str.size();
> }
> ```

Although this can be solved more easily, we can only replace those for which `rl_type` is equal to left.

tests/python/tests/job_workers/php/http_worker.php
```diff
@@ -124,7 +124,7 @@ function test_job_script_timeout_error() {
 
 function test_job_errors() {
   $context = json_decode(file_get_contents('php://input'));
-  $ids = send_jobs($context, 5);
+  $ids = send_jobs($context, 15);
```
unserialize, line 127: Is it intentional, or you just forgot to rollback this line?
(this seems very strange as a change in this PR)

compiler/pipes/convert-sprintf-calls.cpp
```diff
@@ -0,0 +1,179 @@
...
+  // flag for the presence of a value in format_args_var
+  bool is_var{false};
+  // stores a set of vertices of the call arguments from var or array
+  VertexRange args;
+};
+
+struct FormatPart {
+  enum class FormatSpecType {
+    Decimal,
+    String
```
unserialize, line 26: We don't name enum values from a capital letter :)

compiler/pipes/convert-sprintf-calls.cpp
```diff
@@ -0,0 +1,179 @@
...
+
+std::vector<FormatPart> try_parse_format_string(const std::string &format) {
+  std::vector<FormatPart> parts;
+  std::string last_value;
+  bool find_percent = false;
+
+  for (const auto &symbol : format) {
+    if (find_percent) {
+      if (symbol == 'd') {
+        parts.push_back(FormatPart{"", FormatPart::FormatSpecType::Decimal});
```
unserialize, line 45: BTW, this can be rewritten as
```
parts.emplace_back("", FormatPart::FormatSpecType::Decimal);
```

compiler/pipes/convert-sprintf-calls.cpp
```diff
@@ -0,0 +1,179 @@
...
+    const auto func = func_call->func_id;
+    if (vk::any_of_equal(func->name, "sprintf", "vsprintf")) {
+      return convert_sprintf_call(func_call);
+    }
+  }
+
+  return root;
+}
+
+VertexPtr ConvertSprintfCallsPass::convert_sprintf_call(VertexAdaptor<op_func_call> &call) {
```
unserialize, line 83: Why a reference here?

compiler/pipes/convert-sprintf-calls.cpp
```diff
@@ -0,0 +1,179 @@
...
+    parts.push_back(FormatPart{std::move(last_value)});
+  }
+
+  return parts;
+}
+
+VertexPtr ConvertSprintfCallsPass::on_exit_vertex(VertexPtr root) {
+  if (auto func_call = root.try_as<op_func_call>()) {
+    const auto func = func_call->func_id;
+    if (vk::any_of_equal(func->name, "sprintf", "vsprintf")) {
```
unserialize, line 75:
```
if(func->is_extern() && name...)
```
(not to compare strings for every func call)

tests/python/tests/job_workers/php/http_worker.php
i582: Without this, functional tests now fail. I will rebase to the current master, which already has this change. See https://github.com/VKCOM/kphp/commit/22ee0ae898c2386e0c148cbcfcc0cdf042f61fb4

compiler/pipes/convert-sprintf-calls.cpp
AlexK0: discussed offline: let's leave as is and may be fix it in the future

compiler/pipes/convert-sprintf-calls.cpp
i582: > BTW, this can be rewritten as
> 
> ```
> parts.emplace_back("", FormatPart::FormatSpecType::Decimal);
> ```

Since `FormatPart` has no constructors, it seems that it won't be possible to rewrite this way. Perhaps replacing `push_back` with `emplace_back` would be okay.

compiler/pipes/convert-sprintf-calls.cpp
i582: Done.

compiler/pipes/convert-sprintf-calls.cpp
i582: Fixed.

compiler/pipes/convert-sprintf-calls.cpp
i582: Fixed.

tests/python/tests/job_workers/php/http_worker.php
i582: Fixed.

compiler/pipes/convert-sprintf-calls.cpp
i582: Fixed.

compiler/pipes/convert-sprintf-calls.cpp
```diff
@@ -0,0 +1,186 @@
...
+  VertexRange args;
+};
+
+struct FormatPart {
+  enum class FormatSpecType {
+    decimal,
+    string
+  };
+
+  explicit FormatPart(string value)
```
unserialize, line 30: std::string

compiler/pipes/convert-sprintf-calls.h
```diff
@@ -0,0 +1,41 @@
...
+// Depending on the length of the string and count specifiers, the speed
+// of concatenation is several times faster. Even for the given example,
+// the speed of sprintf is 4 times slower than that of concatenation.
+//
+// This optimization was taken out of the optimization pipe in order to
+// place it before the pipe, where the constants are moved into variables,
+// so that the constant strings created in this pipe are moved into variables.
+class ConvertSprintfCallsPass final : public FunctionPassBase {
+public:
+  string get_description() final {
```
unserialize, line 32: std::string

i582: Fixed.

compiler/pipes/convert-sprintf-calls.cpp
i582: Fixed.

unserialize (approved):

PR #257, opened 2021-06-02
Job workers: shared immutable messages

DrDet:
relates to `KPHP-1297`

### DIFF
--- a/compiler/code-gen/declarations.cpp
+++ b/compiler/code-gen/declarations.cpp
@@ -495,7 +495,9 @@ void ClassDeclaration::compile(CodeGenerator &W) const {
 
   compile_back_includes(W, std::move(front_includes));
 
+  W << OpenNamespace{};
   compile_job_worker_shared_memory_piece_methods(W);
+  W << CloseNamespace{};
 
   W << CloseFile();
 }
@@ -715,9 +717,9 @@ void ClassDeclaration::compile_job_worker_shared_memory_piece_methods(CodeGenera
   FunctionSignatureGenerator(W).set_inline() << "void " << klass->src_name
                                                      << "::set_shared_memory_piece(const class_instance<C$KphpJobWorkerSharedMemoryPiece> &instance) " << BEGIN;
   if (field) {
-    W << "const auto &casted = instance.cast_to<" << field->get_inferred_type()->class_type()->src_name << ">();" << NL;
+    W << "auto casted = instance.cast_to<" << field->get_inferred_type()->class_type()->src_name << ">();" << NL;
     W << "php_assert(instance.is_null() || !casted.is_null());" << NL;
-    W << field->get_hash_name() << " = casted;" << NL;
+    W << field->get_hash_name() << " = std::move(casted);" << NL;
   } else {
     W << "(void)instance;" << NL;
   }
--- a/compiler/code-gen/declarations.cpp
+++ b/compiler/code-gen/declarations.cpp
@@ -702,7 +702,10 @@ void ClassDeclaration::compile_job_worker_shared_memory_piece_methods(CodeGenera
     FunctionSignatureGenerator(W).set_overridden() << "void set_shared_memory_piece(const class_instance<C$KphpJobWorkerSharedMemoryPiece> &instance)" << SemicolonAndNL{};
     return;
   }
-  const auto *field = G->get_job_worker_shared_memory_piece(klass->name);
+  const ClassMemberInstanceField *field = nullptr;
+  if (klass->has_job_shared_memory_piece) {
+    field = klass->get_job_shared_memory_pieces().front();
+  }
   W << NL;
   FunctionSignatureGenerator(W).set_const_this().set_inline()
     << "class_instance<C$KphpJobWorkerSharedMemoryPiece> " << klass->src_name << "::get_shared_memory_piece() " << BEGIN;
--- a/compiler/code-gen/declarations.cpp
+++ b/compiler/code-gen/declarations.cpp
@@ -488,10 +488,14 @@ void ClassDeclaration::compile(CodeGenerator &W) const {
     W << END << NL;
   }
 
+  compile_job_worker_shared_memory_piece_methods(W, true);
+
   W << END << ";" << NL;
   W << CloseNamespace();
 
   compile_back_includes(W, std::move(front_includes));
+
+  compile_job_worker_shared_memory_piece_methods(W);
 
   W << CloseFile();
 }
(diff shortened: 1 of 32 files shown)


compiler/compiler-core.h
```diff
@@ -38,6 +41,7 @@ class CompilerCore {
   std::vector<std::string> kphp_runtime_opts;
   bool is_untyped_rpc_tl_used{false};
   function_palette::Palette function_palette;
+  std::map<std::string, const ClassMemberInstanceField *> job_workers_shared_memory_pieces;
```
unserialize, line 44: This is not to fixed in this PR, just a bit to discuss. This field seems overcomplicated for me.
I suppose, that adding `bool has_job_shared_memory_piece` into `ClassData`, setting it to true once (where you  add a value to this map) and a getter method in `ClassData` that just uses full lookup through instance fields, would be enough and more clear that storing a global map here

runtime/job-workers/client-functions.cpp
```diff
@@ -86,65 +105,127 @@ Optional<int64_t> f$kphp_job_worker_start(const class_instance<C$KphpJobWorkerRe
...
+                    requests.begin().get_key().to_string().c_str(), it.get_key().to_string().c_str());
+        return {};
+      }
     }
   }
 
-  int64_t job_resumable_id = register_forked_resumable(new job_resumable{job_id});
+  array<Optional<int64_t>> res{requests.size()};
+
+  job_workers::JobSharedMemoryPiece *common_job_request = nullptr;
```
unserialize, line 178: common_job_memory_piece?

compiler/code-gen/declarations.cpp
```diff
@@ -687,6 +691,36 @@ void ClassDeclaration::compile_back_includes(CodeGenerator &W, IncludesCollector
...
+  W << NL;
+  FunctionSignatureGenerator(W).set_const_this().set_inline()
+    << "class_instance<C$KphpJobWorkerSharedMemoryPiece> " << klass->src_name << "::get_shared_memory_piece() " << BEGIN;
+  W << "return " << (field ? field->get_hash_name() : "{}") << SemicolonAndNL{};
+  W << END << NL << NL;
+
+  FunctionSignatureGenerator(W).set_inline() << "void " << klass->src_name
+                                                     << "::set_shared_memory_piece(const class_instance<C$KphpJobWorkerSharedMemoryPiece> &instance) " << BEGIN;
+  if (field) {
+    W << "const auto &casted = instance.cast_to<" << field->get_inferred_type()->class_type()->src_name << ">();" << NL;
```
AlexK0, line 715: I suggest saving it by value, and using `std::move(casted)` bellow (so we use only move semantic)

compiler/compiler-core.h
DrDet: Hmm. Probably yes, there won't be too many classes containing `KphpJobWorkerSharedMemoryPiece` field, and so we simply can perform full lookup twice for them. I didn't think about that. I agree it would be more clear

compiler/pipes/final-check.cpp
```diff
@@ -42,11 +42,37 @@ void check_class_immutableness(ClassPtr klass) {
...
+    kphp_error(klass->is_immutable, fmt_format("Class {} must be immutable (@kphp-immutable-class) as it implements KphpJobWorkerSharedMemoryPiece", klass->name));
+  }
+  if (implements_request) {
+    size_t shared_memory_piece_members_count = 0;
+    for (ClassPtr ancestor : klass->get_all_ancestors()) {
+      ancestor->members.for_each([&](const ClassMemberInstanceField &field) {
+        for (ClassPtr field_class : field.get_inferred_type()->class_types()) {
+          if (shared_memory_piece_interface->is_parent_of(field_class)) {
+            ++shared_memory_piece_members_count;
+            G->add_job_worker_shared_memory_piece(klass->name, &field);
```
AlexK0, line 68: usage std::map in `add_job_worker_shared_memory_piece()` looks like a race condition :(
May be put this field directly into class ClassData (ClassPtr) object? It will be a bit simpler and without races :)

compiler/code-gen/declarations.cpp
```diff
@@ -488,11 +488,15 @@ void ClassDeclaration::compile(CodeGenerator &W) const {
...
   }
 
+  compile_job_worker_shared_memory_piece_methods(W, true);
+
   W << END << ";" << NL;
   W << CloseNamespace();
 
   compile_back_includes(W, std::move(front_includes));
 
+  compile_job_worker_shared_memory_piece_methods(W);
```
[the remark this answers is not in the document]
DrDet: done

runtime/job-workers/client-functions.cpp
```diff
@@ -86,65 +105,127 @@ Optional<int64_t> f$kphp_job_worker_start(const class_instance<C$KphpJobWorkerRe
...
+      continue;
+    }
 
-class_instance<C$KphpJobWorkerResponse> f$kphp_job_worker_wait(int64_t job_resumable_id, double timeout) noexcept {
-  if (!f$is_kphp_job_workers_enabled()) {
-    php_warning("Can't wait job: job workers disabled");
-    return {};
+    if (common_job_request) {
+      const auto &job_instance = job_request->instance.cast_to<C$KphpJobWorkerRequest>();
+      const auto &common_job_instance = common_job_request->instance.cast_to<C$KphpJobWorkerSharedMemoryPiece>();
```
AlexK0, line 210: let's save `common_job_instance` by value and move it later
`job_instance.get()->set_shared_memory_piece(std::move(common_job_instance));`

compiler/code-gen/declarations.cpp
DrDet: done

compiler/compiler-core.h
DrDet: done

compiler/pipes/final-check.cpp
DrDet: fixed

runtime/job-workers/client-functions.cpp
```diff
@@ -86,65 +105,127 @@ Optional<int64_t> f$kphp_job_worker_start(const class_instance<C$KphpJobWorkerRe
...
+
+  job_workers::JobSharedMemoryPiece *common_job_request = nullptr;
+  if (!common_shared_memory_piece.is_null()) {
+    common_job_request = make_job_request_message<job_workers::JobSharedMemoryPiece>(common_shared_memory_piece);
+    /**
+     * common_job_request lifetime:
+     * 1. attaches to client process on creating in `acquire_shared_message()`
+     * 2. attaches to job worker process on job receiving in `job_parse_execute()`
+     * 3. detaches from job worker process on terminating in `release_shared_message()` recursively
+     * 4. detaches from client process at the end of this function in `release_shared_message()` recursively
```
AlexK0, line 186: It seems to me that at the end of this function `release_shared_message()` detaches common_job_request without recursion

runtime/job-workers/client-functions.cpp
```diff
@@ -86,65 +105,127 @@ Optional<int64_t> f$kphp_job_worker_start(const class_instance<C$KphpJobWorkerRe
...
 
-  update_precise_now();
-  kphp_event_timer *timer = allocate_event_timer(get_precise_now() + timeout, get_job_timeout_wakeup_id(), job_id);
+  for (const auto &it : requests) {
+    const auto &req = it.get_value();
 
-  vk::singleton<job_workers::ProcessingJobs>::get().start_job_processing(job_id, job_workers::JobRequestInfo{job_resumable_id, timer});
-  return job_resumable_id;
-}
+    req.get()->set_shared_memory_piece({});                                                 // prepare for copying to shared memory
```
AlexK0: something wrong with comments block :(

runtime/job-workers/client-functions.cpp
DrDet: yep, fixed comment

runtime/job-workers/client-functions.cpp
DrDet: fixed

AlexK0 (approved):

PR #258, opened 2021-06-07
Optimize array_diff() for array<int64_t> arguments

drdzyk:


quasilyte:
Could you provide some rationale behind it and, if possible, some benchmark results?

runtime/array_functions.h
```diff
@@ -789,23 +789,31 @@ array<T> f$array_diff_key(const array<T> &a1, const array<T1> &a2) {
   return result;
 }
 
-template<class T, class T1>
-array<T> f$array_diff(const array<T> &a1, const array<T1> &a2) {
+template<typename T, typename T1, typename Proj>
+array<T> array_diff_impl(const array<T> &a1, const array<T1> &a2, const Proj &projector) {
   array<T> result(a1.size());
 
   array<int64_t> values(array_size(0, a2.count(), false));
```
AlexK0, line 796: The second param here is about of the expected number of string keys in array (for the reserving). 
For int specialization it is better to use the first parameter for reserving space for the integer keys.

runtime/array_functions.h
```diff
@@ -789,23 +789,31 @@ array<T> f$array_diff_key(const array<T> &a1, const array<T1> &a2) {
   return result;
 }
 
-template<class T, class T1>
-array<T> f$array_diff(const array<T> &a1, const array<T1> &a2) {
+template<typename T, typename T1, typename Proj>
```
AlexK0, line 792: pls, use class instead of typename.

drdzyk:
Sure. With this simple benchmark:
```php
<?php

function run_bench() {
    $a = ["a" => 1, 2, 3, 500, 7000, 324234, 3242, 45656, 23324234, "ad" => 432, 5634, 2234345, 56435, 234234];
    $b = ["b" => 1, 2, 4, 34, 265264, 5435, 564, 87657, 43, 34534, "ds" => 324, 46435, 23423, 456345, 23423];
    $c = [];
    $start = hrtime(true);
    for ($i = 0; $i < 1000000; $i++) {
        $c = array_diff($a, $b);
    }
    $elapsed = (hrtime(true) - $start) / 1000000000;
    echo "elapsed: ", $elapsed, ", s\n";
}

run_bench();
```
at my machine master shows ~2.5 sec, while PR is about 1.6 sec.

runtime/array_functions.h
drdzyk: done

runtime/array_functions.h
drdzyk: Yea, nice idea. It allow to speed up my synthetic benchmark from ~1.6 to ~1.5 sec, but the code became a bit more complicated. I'm not sure worth this changes to be merged?

AlexK0: this is fine, let's leave as is

AlexK0 (approved):

PR #264, opened 2021-06-16
Precompute hashes for isset() and unset()

drdzyk:
Currently, isset() and unset() functions of array calculate hash from compile time known strings at runtime. This patch modify such behavior and move calculation of hash at code generation phase.

### DIFF
--- a/runtime/array.inl
+++ b/runtime/array.inl
@@ -1307,14 +1307,14 @@ void array<T>::assign_raw(const char *s) {
 }
 
 template<class T>
-const T *array<T>::find_value(int64_t int_key) const {
+const T *array<T>::find_value(int64_t int_key) const noexcept {
   return p->is_vector()
          ? p->find_vector_value(int_key)
          : p->find_map_value(int_key);
 }
 
 template<class T>
-const T *array<T>::find_value(const string &string_key) const {
+const T *array<T>::find_value(const string &string_key) const noexcept {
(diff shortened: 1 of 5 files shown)


AlexK0 (approved):

unserialize:
If you don't mind, could you write richer descriptions for PRs? :)

runtime/array.inl
```diff
@@ -1488,8 +1494,17 @@ void array<T>::unset(const string &string_key) {
     return;
   }
```
unserialize, line 1495: This if() can be deleted?

unserialize (approved):

drdzyk:
> If you don't mind, could you write richer descriptions for PRs? :)
done)

runtime/array.inl
drdzyk: Sure, but in this case we get small performance issue - hash will be calculated even for array that is actually vector. Given that is_vector() much cheaper then string_key.hash() I think better leave as is.

unserialize: Yes, you're right, but partially. Because unsetting from a vector by a string key is a very strange operation, that will immediately lead to converting this vector-array to a hashmap-array :)
But ok, let's leave it

unserialize (approved):

AlexK0 (approved):

PR #265, opened 2021-06-17
Enable smartcasts for opset inside condition

drdzyk:
Enable smartcasts for opset inside condition; this also involves smartcasts in shorthand ternary operator.

### DIFF
--- a/compiler/gentree.cpp
+++ b/compiler/gentree.cpp
@@ -701,7 +701,7 @@ VertexAdaptor<op_ternary> GenTree::create_ternary_op_vertex(VertexPtr condition,
     return VertexAdaptor<op_ternary>::create(condition, true_expr, false_expr);
   }
 
-  auto cond_var = create_superlocal_var("shorthand_ternary_cond");
+  auto cond_var = create_superlocal_var("shorthand_ternary_cond").set_location(condition);
 
   auto cond = conv_to<tp_bool>(VertexAdaptor<op_set>::create(cond_var, condition));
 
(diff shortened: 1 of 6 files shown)


unserialize:
Did you examine the codegenerated diff for vkcom after this change?

drdzyk:
> Did you examine the codegenerated diff for vkcom after this change?

After receiving access to vkcom and vacation I finally did this =)
At first glance everything looks good. Here is few main types of changes in code base:

1)
```
/** @return ?int[] */
function get_optional_array() { return []; }

function demo() {
  $elements = get_optional_array() ?: [];
}
```
Here $elements inferred as array<int64_t>, but not Optional<array<int64_t>> as in master.

2)
```
/** @return ?int */
function get_optional_int() { return 0; }

function demo() {
  $elements = [];
  if ($element = get_optional_int()) {
    $elements[] = $element;
  }
}
```
Here $elements inferred as array<int64_t>, but not array<Optional<int64_t>> as in master.

3)
```
/** @return ?int[] */
function get_optional_array() { return []; }

function demo() {
  if ($elements = get_optional_array()) {
    $element = $elements[0];
  }
}
```
Here $element inferred as int64_t in both branches, but in patch additional 'check_not_null()' aux function will be called.

4)
```
/** @return ?int[] */
function get_optional_array() { return []; }

/** @return mixed */
function demo() {
  return get_optional_array() ?: [];
}
```
Here we fixate return type as mixed, but in patch mixed constructed from array<int64_t>, in master - from Optional < array<int64_t>>.

unserialize (approved):

PR #269, opened 2021-06-21
add support of spread operator `(...$a)`

i582:
Example of spread operator:

```php
$a = [1, 2];
$b = [1, ...$a]; // will be [1, 1, 2]
```

https://wiki.php.net/rfc/spread_operator_for_array

### Implementation
- All arrays with the spread operator are converted to `array_merge_spread` call;
  For example:
  ```php
  $a = [1, 2];
  $b = [1, ...$a, 1, 2, ...$a];
  // converted to
  $b = array_merge_spread([1], $a, [1, 2], $a);
  ```
- Added a new function `array_merge_spread`, which actually calls `array_merge`, but throws 
  an error at runtime if arrays with keys are passed to it;

> PHP 8.1 is planned to allow arrays with string keys, this will need to be taken 
into account if we are going to make features from PHP 8.
https://wiki.php.net/rfc/array_unpacking_string_keys


runtime/array_functions.h
```diff
@@ -669,6 +680,34 @@ R f$array_reduce(const array<T> &a, const CallbackT &callback, InitialT initial)
   return result;
 }
 
+template<class T>
+T f$array_merge_spread(const T &a1) {
+  if (!a1.is_vector()) {
+    php_error("Cannot unpack array with string keys");
```
AlexK0, line 686: usually we use php_warning

compiler/pipes/convert-spread-operator.cpp
```diff
@@ -0,0 +1,82 @@
...
+#include "compiler/compiler-core.h"
+#include "compiler/inferring/public.h"
+
+VertexAdaptor<op_array> array_vertex_from_slice(const VertexConstRange &args, size_t start, size_t end) {
+  return VertexAdaptor<op_array>::create(
+    std::vector<VertexPtr>{args.begin() + start, args.begin() + end}
+  );
+}
+
+VertexPtr ConvertSpreadOperatorPass::on_exit_vertex(VertexPtr root) {
```
AlexK0, line 16: common comment:
Creating array for non array elements is not a simple operation and it works slow. 
I believe, that we can do better and avoid of creating useless arrays.

compiler/gentree.cpp
```diff
@@ -477,6 +477,21 @@ VertexPtr GenTree::get_expr_top(bool was_arrow) {
       break;
     }
     case tok_varg: {
+      // cases:
+      // [...$a]:        prev: tok_opbrk, next: tok_var_name
+      // [1, ...$a]:     prev: tok_comma, next: tok_var_name
+      // [1, ...$a, 1]:  prev: tok_comma, next: tok_var_name
+      // [1, ...f(), 1]: prev: tok_comma, next: tok_func_name
```
unserialize, line 484: I suppose, that `array(...$a)` is also a valid syntax?

compiler/vertex-desc.json
```diff
@@ -1779,6 +1779,22 @@
       }
     }
   },
+
```
unserialize, line 1782: We don't use empty lines between blocks in vertex-desc.json

compiler/pipes/convert-spread-operator.cpp
```diff
@@ -0,0 +1,82 @@
...
+
+  const auto array_merge_spread = G->get_function("array_merge_spread");
+  if (!array_merge_spread) {
+    return array_vertex;
+  }
+
+  auto call = VertexAdaptor<op_func_call>::create(parts);
+  call->func_id = array_merge_spread;
+  call->set_string(array_merge_spread->name);
+  call.set_location(root->location);
```
unserialize, line 79: Above: `auto call = ...::create(...).set_location(root->location);`

compiler/compiler.cpp
```diff
@@ -237,6 +238,7 @@ bool compiler_execute(CompilerSettings *settings) {
     >> PipeC<SplitSwitchF>{}
     >> PipeC<CollectRequiredAndClassesF>{} >> use_nth_output_tag<0>{}
     >> SyncC<CheckRequires>{}
+    >> PassC<ConvertSpreadOperatorPass>{}
```
unserialize, line 241: Do we really need a separate pass for this? Can't this be done in GenTreePostprocess or right in gentree itself?

runtime/array_functions.h
i582: Fixed.

compiler/gentree.cpp
i582: Yep, added.

compiler/vertex-desc.json
i582: Fixed.

compiler/pipes/convert-spread-operator.cpp
i582: Fixed.

compiler/compiler.cpp
i582: Yep, moved to `GenTreePostprocessPass`.

i582: It doesn't seem to work, since we need to get the function `array_merge_spread`, so we need to do the conversion after the functions are collected.

tests/phpt/spread/001_spread_ok.php
```diff
@@ -0,0 +1,49 @@
+@ok php7_4
```
unserialize, line 1: Do we need `php7_4` tag?

i582: Until all tests run on 7.4, we need this tag here.

compiler/pipes/gen-tree-postprocess.cpp
```diff
@@ -252,5 +252,71 @@ VertexPtr GenTreePostprocessPass::on_exit_vertex(VertexPtr root) {
     }
   }
 
+  if (auto array = root.try_as<op_array>()) {
+    return convert_array_with_spread_operators(array);
```
unserialize, line 256: You launch this function (and a loop inside it) for every array in code. While it's acceptable, it's not good, as it leads to excessive computations even when there is no spread operator (99.99% cases).
To avoid this, you can add a field `was_spread_operator_` to this pass, and then use `on_exit_vertex` checking for `op_array` both with this flag set. Be careful with nesting.

quasilyte:
We decided to merge this as is and address the things later.

quasilyte (approved):

PR #270, opened 2021-06-22
Flex refactoring

DrDet:
Here is refactoring of flex:
1. Extracted copypasted flex logic from kphp and vkext to separate function in flex lib
2. Added wide set of tests
3. Refactoring of main flex logic

relates to `KPHP-1310`


flex/flex.cpp
```diff
@@ -0,0 +1,164 @@
...
+}
+
+bool lang::has_symbol(char c) const noexcept {
+  return strchr(flexible_symbols, c) != nullptr;
+}
+
+const char *flex(vk::string_view name, vk::string_view case_name, bool is_female, vk::string_view type, int lang_id, char * const dst_buf, char * const err_buf) {
+  if (name.size() > (1 << 10)) {
+    sprintf(err_buf, "Name has length %zu and is too long in function vk_flex", name.size());
+    return name.data();
```
AlexK0, line 44: nit

roughly speaking it is unsafe to return data() from string_view, because in general it's not a null terminated string, but const char * may assume it. I suggest using vk::string_view as return value and use size() on calling place.

flex/flex.h
```diff
@@ -0,0 +1,40 @@
...
+// Copyright (c) 2021 LLC «V Kontakte»
+// Distributed under the GPL v3 License, see LICENSE.notice.txt
+
+#pragma once
+
+#include "common/wrappers/string_view.h"
+
+struct lang;
+
+struct node {
```
unserialize, line 11: What do you think of renaming `node` to `flex_node` and `lang` to `flex_lang`? It's a bit weird that such general names exist in a global scope

DrDet: I agree with you, but the problem is It will break backward compatibility. In that case we will need to update flex, kphp and vkext simultaneously somehow
Current version preserves backward compatibility and we can simply update flex at first, and then kphp and vkext

flex/flex.cpp
DrDet: fixed

AlexK0 (approved):

PR #282, opened 2021-07-14
[tl-compiler] allow use constructor as argument type before constructor declaration

Kurpilyansky:
If type of the argument is undeclared constructor, then its type can not be polymorphic.
If type can not be polymorphic, then constructor name and type name are differ in register of first letter.

This commit allows to concatenate tl-scheme files really in any order.

For example, now you can use primitive type int before its declaration:

```
....
@readwrite friends.deleteUser uid:int = Bool;
...
int#a8509bda ? = Int;
```

### DIFF
--- a/common/tl/compiler/tl-parser-new.cpp
+++ b/common/tl/compiler/tl-parser-new.cpp
@@ -944,11 +944,11 @@ std::pair<bool, std::string> convert_type_to_bare_constructor(std::string id) {
       break;
     }
   }
-  if ('a' <= id[pos] && id[pos] <= 'z') {
+  if (std::islower(static_cast<unsigned char>(id[pos]))) {
     return {false, std::move(id)};
   } else {
-    assert('A' <= id[pos] && id[pos] <= 'Z');
-    id[pos] += 'a' - 'A';
+    assert(std::isupper(static_cast<unsigned char>(id[pos])));
+    id[pos] = std::tolower(static_cast<unsigned char>(id[pos]));
     return {true, std::move(id)};
   }
 }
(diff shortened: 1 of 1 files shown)


Kurpilyansky:
Couple bugs fixed. Now resulting combined.tlo does not differ with old compiler result.

Kurpilyansky:
I found out that there are undocumented type flags in resulting tlo.

Some statistics about flags from tls.Type based on VKCOM tl scheme
118 types with 1<<0 = FLAG_BARE
1 type with 1<<25 = FLAG_DEFAULT_CONSTRUCTOR

492 types with 1<<3 = FLAG_USED_AS_BARE???
133 types (most of them enum-like types) with 1<<4 = FLAG_HAVE_EQUAL_CONSTRUCTORS???

Is appearance of unknown flags a bug or a feature?

common/tl/compiler/tl-parser-new.cpp
```diff
@@ -932,18 +933,44 @@ tree_tl_var_t *vars[10];
...
+      pos = i + 1;
+      break;
     }
   }
-  return ok;
+  if ('A' <= id[pos] && id[pos] <= 'Z') {
+    return {false, std::move(id)};
+  } else {
+    assert('a' <= id[pos] && id[pos] <= 'z');
+    id[pos] += 'A' - 'a';
```
DrDet, line 971: std::toupper?

common/tl/compiler/tl-parser-new.cpp
```diff
@@ -932,18 +933,44 @@ tree_tl_var_t *vars[10];
...
+  size_t pos = 0;
+  for (size_t i = 0; i < id.size() - 1; i++) {
     if (id[i] == '.') {
-      ok = id[i + 1] >= 'A' && id[i + 1] <= 'Z';
+      pos = i + 1;
+      break;
     }
   }
-  return ok;
+  if ('A' <= id[pos] && id[pos] <= 'Z') {
```
DrDet, line 967: std::isupper?

common/tl/compiler/tl-parser-new.cpp
```diff
@@ -932,18 +933,44 @@ tree_tl_var_t *vars[10];
...
+    return {false, std::move(id)};
+  } else {
+    assert('a' <= id[pos] && id[pos] <= 'z');
```
DrDet, line 970: std::islower?

common/tl/compiler/tl-parser-new.cpp
Kurpilyansky: done

common/tl/compiler/tl-parser-new.cpp
Kurpilyansky: done

common/tl/compiler/tl-parser-new.cpp
Kurpilyansky: done

unserialize (approved):

PR #286, opened 2021-07-29
added support for separator in numeric literal

i582:
RFC: https://wiki.php.net/rfc/numeric_literal_separator

Example:
```php
echo 100_000;
echo 0b1111_0011;
```

Added two types of tokens to represent delimited numbers:
- `tok_int_const_sep`
- `tok_float_const_sep`


compiler/gentree.cpp
```diff
@@ -436,6 +436,13 @@ VertexPtr GenTree::get_expr_top(bool was_arrow) {
       next_cur();
       break;
     }
+    case tok_int_const_sep: {
+      std::string val = static_cast<string>(cur->str_val);
+      remove_underscores(val);
```
quasilyte, line 441: Since token is "_sep" (separators?), maybe call remove_underscores as remove_separators?

i582: Done.

quasilyte (approved):

PR #288, opened 2021-07-31
mysql with user, password, port and remote host, utf8 header parameter

ivashkevitch:


net/net-mysql-client.cpp
```diff
@@ -271,10 +271,12 @@ static int sqlc_inner_authorise (struct connection *c) {
 
   T = (struct mysql_auth_packet_end *)p;
 
-  const char *sql_username = SQLC_FUNC(c)->sql_get_username(c);
   const char *sql_database = SQLC_FUNC(c)->sql_get_database(c);
+  const char *sql_user = SQLC_FUNC(c)->sql_get_user(c);
   const char *sql_password = SQLC_FUNC(c)->sql_get_password(c);
 
+  fprintf(stdout, "mysql db: %s; user: %s; password: *\n", sql_database, sql_user);
```
quasilyte, line 278: Is this intended to stay here or it's a debug leftover?

runtime/interface.cpp
```diff
@@ -72,6 +72,8 @@ static string_buffer oub[OB_MAX_BUFFERS];
 string_buffer *coub;
 static int http_need_gzip;
 
+static int is_utf8_enabled = false;
```
DrDet, line 75:
```suggestion
static bool is_utf8_enabled = false;
```

server/php-engine.cpp
```diff
@@ -121,6 +121,7 @@ tcp_rpc_client_functions tcp_rpc_client_outbound = [] {
  ***/
 
 static int db_port = 3306;
+const char *db_host = "localhost";
```
DrDet, line 124:
```suggestion
static const char *db_host = "localhost";
```

server/php-engine.cpp
```diff
@@ -1508,7 +1509,8 @@ static void generic_event_loop(WorkerType worker_type, bool init_and_listen_rpc_
   if (no_sql) {
     sql_target_id = -1;
   } else {
-    sql_target_id = get_target("localhost", db_port, &db_ct);
+    fprintf(stdout, "mysql host: %s; port: %d\n", db_host, db_port);
```
DrDet, line 1512: If you want to log something here, use `kprintf` for this. It adds some useful output like pid, time and so on

DrDet (changes requested):

server/php-engine.cpp
```diff
@@ -2071,6 +2107,11 @@ void parse_main_args(int argc, char *argv[]) {
   parse_option("job-workers-shared-memory-size", required_argument, 2017, "the total size of shared memory used for job workers related communication");
   parse_option("job-workers-shared-messages", required_argument, 2018, "the total count of the shared messages for job workers related communication");
   parse_option("lease-stop-ready-timeout", required_argument, 2019, "timeout for RPC_STOP_READY acknowledgement waiting in seconds (default: 0)");
+  parse_option("mysql-user", required_argument, 2020, "MySQL user");
+  parse_option("mysql-password", required_argument, 2021, "MySQL password");
+  parse_option("mysql-host", required_argument, 2022, "MySQL host");
```
DrDet, line 2112: I think it's a bit strange to have customizable mysql host and not to have option for setting mysql port. What do you think about this?

ivashkevitch: I agree with you, I just forgot about port customization, updated it, thanks

server/php-engine.cpp
ivashkevitch: Ok, fixed

net/net-mysql-client.cpp
ivashkevitch: I think it's useful information when use custom db settings, but of course, it's not required

server/php-engine.cpp
DrDet: Oh, sorry for that, It turns out such an option already exists. It's `-Q` option. So new option isn't needed actually :(

server/php-engine.cpp
```diff
@@ -2000,6 +2012,47 @@ int main_args_handler(int i, const char *long_option) {
...
+      }
+      kprintf("--%s option: couldn't set mysql password '%s'\n", long_option, optarg);
+      return -1;
+    }
+    case 2022: {
+      if (set_mysql_host(optarg)) {
+        return 0;
+      }
+      kprintf("--%s option: couldn't set mysql host '%s'\n", long_option, optarg);
+      return -1;
```
DrDet, line 2034:
```suggestion
      db_host = optarg;
      return 0;
```

server/php-engine.cpp
```diff
@@ -1729,6 +1731,16 @@ int read_option_to(const char *option_name, T min_value, T max_value, T &out) no
...
 
+bool set_mysql_host(const char *host) {
+  db_host = host;
+  return true;
+}
+
+bool set_mysql_port(const int port) {
+  db_port = port;
+  return true;
+}
```
DrDet, line 1742: I think these functions are redundant

net/net-mysql-client.cpp
```diff
@@ -271,10 +271,12 @@ static int sqlc_inner_authorise (struct connection *c) {
 
   T = (struct mysql_auth_packet_end *)p;
 
-  const char *sql_username = SQLC_FUNC(c)->sql_get_username(c);
   const char *sql_database = SQLC_FUNC(c)->sql_get_database(c);
+  const char *sql_user = SQLC_FUNC(c)->sql_get_user(c);
   const char *sql_password = SQLC_FUNC(c)->sql_get_password(c);
 
+  kprintf("mysql db: %s; user: %s; password: *\n", sql_database, sql_user);
```
DrDet, line 278:
```suggestion
  vkprintf(1, "mysql db: %s; user: %s; password: *\n", sql_database, sql_user);
```
It's logging with verbosity set to >= 1

server/php-engine.cpp
ivashkevitch: Kek, exacltly) No problems, fixed

server/php-engine.cpp
ivashkevitch: +

unserialize (approved):

Issue #290, opened 2021-08-09
PHP 8


i582:
# PHP 8

### New syntax

* [ ] Attributes                              ([RFC](https://wiki.php.net/rfc/attributes_v2))
* [ ] Union types in type hints               ([RFC](https://wiki.php.net/rfc/union_types_v2))
* [ ] `nullsafe` operator (`?->`)             ([RFC](https://wiki.php.net/rfc/nullsafe_operator))
* [ ] `static` typehint                       (RFC not found)
* [ ] `match` expression                      ([RFC](https://wiki.php.net/rfc/match_expression_v2))
* [ ] `throw` can be used as an expression    ([RFC](https://wiki.php.net/rfc/throw_expression))
* [ ] Block `catch` without variable          ([RFC](https://wiki.php.net/rfc/non-capturing_catches))
* [ ] Named arguments                         ([RFC](https://wiki.php.net/rfc/named_params))
* [ ] Declaring properties in the constructor ([RFC](https://wiki.php.net/rfc/constructor_promotion))
* [ ] Trailing comma in parameter lists       ([RFC](https://wiki.php.net/rfc/trailing_comma_in_parameter_list))
* [ ] Trailing comma in closure use list      ([RFC](https://wiki.php.net/rfc/trailing_comma_in_closure_use_list))

### Other

* [ ] `Stringable` interface ([RFC](https://wiki.php.net/rfc/stringable))

## Changed

* [ ] Locale-independent float to string cast                              ([RFC](https://wiki.php.net/rfc/locale_independent_float_to_string))
* [ ] Sorting now stable                                                   ([RFC](https://wiki.php.net/rfc/stable_sorting))
* [ ] Names in the namespace are treated as a single token                 ([RFC](https://wiki.php.net/rfc/namespaced_names_as_token))
* [ ] Deferencable changes and arbitrary expressions in `new`/`instanceof` ([RFC](https://wiki.php.net/rfc/variable_syntax_tweaks))
* [ ] Access `::class` on object                                           ([RFC](https://wiki.php.net/rfc/class_name_literal_on_object))
* [ ] Concatenation precedence                                             ([RFC](https://wiki.php.net/rfc/concatenation_precedence))

## Removed

* [ ] Remove `(real)` cast  (without separate RFC)
* [ ] Remove `(unset)` cast (without separate RFC)
* [ ] Remove `{}` access    (without separate RFC)

# PHP 8.1

### New syntax

[...]

PR #292, opened 2021-08-18
added switchable warning for comparison numbers and strings

i582:
In PHP 8, the logic for comparing numbers and strings has changed, so we need a way to find all such places, while not including warning at all, as this will lead to a huge number of warnings.

In order to enable the warning, the `set_migration_php8_warning` function has been added, with which you can enable or disable the warning. This function accepts a bitset, in which only the first right bit is used so far, but the rest will be used in the future.

```php
<?php
$b = 0;

set_migration_php8_warning(0b001);

if ($b == "") {} // warning

set_migration_php8_warning(0b000);

if ($b == "") {} // no warning
```

At the end of each request, the warning is automatically disabled.

### Changes in

The operators `<=>`, `==`, `!=`, `>`, `>=`, `<`, and `<=`.
The functions `in_array()`, `array_search()` and `array_keys()` with `$strict` set to false (which is the default).
The sorting functions `sort()`, `rsort()`, `asort()`, `arsort()` and `array_multisort()` with `$sort_flags` set to `SORT_REGULAR` (which is the default).

### Links

RFC: https://wiki.php.net/rfc/string_to_number_comparison

In documentation: [String to Number Comparison](https://www.php.net/manual/en/migration80.incompatible.php#migration80.incompatible.core.string-number-comparision)

The following table shows how the result of some simple comparisons changes (or doesn't change) under this RFC:

Comparison    | Before | After
---------------|---------|----
 `0 == "0"`     | `true`   | `true`
 `0 == "0.0"`   | `true` | `true`
 `0 == "foo"`   | `true` | `false`
 `0 == ""`      | `true` | `false`
`42 == "   42"` | `true` | `true`
`42 == "42foo"` | `true` | `false`

See full table in tests: https://github.com/VKCOM/kphp/blob/d5a76ad890e46244e94a5e3003212383e351ca14/tests/cpp/runtime/number-string-comparison.cpp#L21

### Note
We are already comparing infinities as in PHP 8.
See [Special values](https://wiki.php.net/rfc/string_to_number_comparison#special_values) part of RFC.

### Related changes (not implemented in this PR, see #298)

In https://wiki.php.net/rfc/saner-numeric-strings (implemented in PHP 8):
```php
42 == "42"; // true
42 == "   42"; // true
42 == "42   "; // true
```

Before

```php
42 == "42"; // true
42 == "   42"; // true
42 == "42   "; // false
```

See https://github.com/VKCOM/kphp-polyfills/pull/25
#290


runtime/mixed.inl
```diff
@@ -1673,8 +1675,113 @@ inline int64_t operator>>(const mixed &lhs, const mixed &rhs) {
...
 }
 
+template <typename Arg1, typename Arg2>
+struct conversion_php_warning_string {
+  static constexpr const char *text = "";
+};
+
+template<>
+struct conversion_php_warning_string<int64_t, const string &> {
+  static constexpr const char *text
```
AlexK0, line 1685: It may not work on some builds. It would be better to use functions instead.

runtime/mixed.inl
```diff
@@ -1673,8 +1675,113 @@ inline int64_t operator>>(const mixed &lhs, const mixed &rhs) {
   return lhs.to_int() >> rhs.to_int();
 }
 
+template <typename Arg1, typename Arg2>
+struct conversion_php_warning_string {
+  static constexpr const char *text = "";
+};
+
+template<>
+struct conversion_php_warning_string<int64_t, const string &> {
```
AlexK0, line 1684: the const reference overloading doesn't look good, it is better to use a [decayed](https://en.cppreference.com/w/cpp/types/decay) (clear) type without cv and refs.

tests/python/tests/migration_php8/test_comparison.py
```diff
@@ -0,0 +1,81 @@
+import os
+from typing import List
+
+from python.lib.testcase import KphpCompilerAutoTestCase
+
+from python.lib.colors import green, red, yellow
+
+
+class TestPHP8NumberStringComparison(KphpCompilerAutoTestCase):
```
AlexK0, line 9: Take a look at _**kphp_should_fail**_ and _**kphp_should_warn**_ tags. I strongly recommend add the same for runtime warnings and use it instead of the function python tests.

runtime/mixed.inl
i582: Done.

runtime/mixed.inl
i582: Done.

tests/python/tests/migration_php8/test_comparison.py
i582: Done, added `@kphp_runtime_should_warn` tag.

PR #293, opened 2021-08-20
Job workers improvements

AlexK0:
This patch
1) Saves references when copying instances to the instance cache and between workers.
2) Fixes a bug, which spoiled multiple inheritance for job workers requests/responses classes.

### DIFF
--- a/runtime/instance-cache.cpp
+++ b/runtime/instance-cache.cpp
@@ -638,9 +638,6 @@ private:
   }
 
   void fire_warning(const InstanceDeepCopyVisitor &detach_processor, const char *class_name) noexcept {
-    if (detach_processor.is_depth_limit_exceeded()) {
-      php_warning("Depth limit exceeded on cloning instance of class '%s' into cache", class_name);
-    }
     if (detach_processor.is_memory_limit_exceeded()) {
       php_warning("Memory limit exceeded on saving instance of class '%s' into cache", class_name);
       context_->memory_swap_required = true;
--- a/runtime/refcountable_php_classes.h
+++ b/runtime/refcountable_php_classes.h
@@ -17,36 +17,43 @@ public:
   virtual void release() noexcept = 0;
   virtual uint32_t get_refcnt() noexcept = 0;
   virtual void set_refcnt(uint32_t new_refcnt) noexcept = 0;
+
+  virtual uint32_t &get_unique_index_ref() noexcept = 0;
 };
 
 template<class ...Bases>
 class refcountable_polymorphic_php_classes : public Bases... {
 public:
   void add_ref() noexcept final {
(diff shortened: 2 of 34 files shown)


tests/python/tests/job_workers/php/ReferenceInvariant/job_worker.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+use ReferenceInvariant\ReferenceInvariantRequest;
+use ReferenceInvariant\ReferenceInvariantResponse;
+
+function run_job_reference_invariant(ReferenceInvariantRequest $request) {
+  $request->verify_data();
+
```
DrDet, line 8: Don't we need to verify refcnt here as well as at http worker?

runtime/instance-cache.cpp
```diff
@@ -638,9 +638,6 @@ class InstanceCache {
   }
 
   void fire_warning(const InstanceDeepCopyVisitor &detach_processor, const char *class_name) noexcept {
-    if (detach_processor.is_depth_limit_exceeded()) {
-      php_warning("Depth limit exceeded on cloning instance of class '%s' into cache", class_name);
-    }
```
DrDet, line 643: Why was it deleted? Was it only for cycle refs?

runtime/refcountable_php_classes.h
```diff
@@ -117,11 +139,16 @@ class refcountable_php_classes  : public ManagedThroughDlAllocator {
...
+  }
+
+  uint32_t &get_unique_index_ref() noexcept {
+    return unique_index_;
   }
 
 private:
-  uint32_t refcnt{0};
+  uint32_t refcnt_{0};
+  uint32_t unique_index_{0};
```
DrDet, line 151: Why does it not change size of every php class? As far as I can see, alignment of `refcountable_php_classes` is 4

runtime/instance-cache.cpp
AlexK0: Yep, the main users of this feature were cycles. 
Now, when cycles are supported, the memory limit is enough as one limit for everything.

runtime/refcountable_php_classes.h
AlexK0: You are right, but most of our data types:
int64_t - 8 bytes
double - 8 bytes
class_instance - 8 bytes
mixed - align 8 bytes , size 16
string - 8 bytes
array - 8 bytes

and only bool - 1 byte, and Optional<bool> - 2 bytes

so it matters only for the instances with bool or Optional<bool> (or tuple<bool, bool> :)) ) as the first fields, and for the other cases it doesn't matter at all.

DrDet: ok, I got it)

DrDet (approved):

PR #295, opened 2021-08-25
WIP: initial FFI implementation for KPHP

quasilyte:
Based on PHP FFI model:

* RFC: https://wiki.php.net/rfc/ffi
* Docs: https://www.php.net/manual/ru/book.ffi.php

TODO: more information.


quasilyte:
How to use FFI from this branch.

## Simplest case, no shared libraries

1. Prepare a test script.

```php
<?PHP

$int = FFI::new('int8_t');
var_dump(FFI::sizeof($int));
var_dump(FFI::sizeof(FFI::addr($int));
```

2. Compile and run normally.

```bash
$ kphp2cpp --mode cli main.php
$ ./kphp_out/cli
int(1)
int(8)
```

3. Compare with PHP

```bash
$ php8.0 -f main.php
int(1)
int(8)
```

Note: you may need to enable FFI in your `php.ini` or via the command line option:

```bash
php7.4 -d ffi.enable_cli='true' -f main.php
```

## Shared library example

1. Prepare some test script.

```php
<?php

$cdef = FFI::cdef('
  const char *getenv(const char *name);
', 'libc.so.6');
$home = $cdef->getenv("HOME");
var_dump($home);
```

2. Compile with extra linker flags.

```bash
$ kphp2cpp --mode cli --extra-linker-flags='-ldl -L${KPHP_PATH}/objs/flex -ggdb' main.php
```

Note that we only need an extra `-ldl` option, but our `--extra-linker-flags` are not empty by default, so we have to include the original defaults to actually add some flags instead of replacing the default set. (Maybe we should fix that?)

3. Run normally.

```bash
$ ./kphp_out/cli
string(15) "/home/isharipov"
```

Same with PHP:

```bash
$ php8.0 -f main.php
string(15) "/home/isharipov"
```

PR #303, opened 2021-09-06
WIP: `match` expression

i582:
#290

### DIFF
--- a/compiler/debug.cpp
+++ b/compiler/debug.cpp
@@ -57,6 +57,7 @@ std::string debugTokenName(TokenType t) {
     {tok_as, "tok_as"},
     {tok_case, "tok_case"},
     {tok_switch, "tok_switch"},
+    {tok_match, "tok_match"},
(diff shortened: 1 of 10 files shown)


compiler/debug.cpp
```diff
@@ -57,6 +57,7 @@ std::string debugTokenName(TokenType t) {
     {tok_as, "tok_as"},
     {tok_case, "tok_case"},
     {tok_switch, "tok_switch"},
+    {tok_match, "tok_match"},
```
quasilyte, line 60: Add a lexer test case for a new token?

i582: I haven't started making tests yet, this is an early draft.

i582: Added some tests for lexer.

vkaverin:
Is it possible to add here some kind of annotation to force exhaustion check at compile time instead of runtime `warning`?

i582:
@Danil42Russia review please

PR #313, opened 2021-09-15
fixed a bug with bin and hex literals

i582:
Previously, using uppercase letters in binary and hexadecimal (`0B10` or `0XAA`) literals resulted in a parsing error.

### DIFF
--- a/compiler/lexer.cpp
+++ b/compiler/lexer.cpp
@@ -367,10 +367,12 @@ bool TokenLexerNum::parse(LexerData *lexer_data) const {
     binary,
   } state = before_dot;
 
-  if (s[0] == '0' && s[1] == 'x') {
+  const auto hex_symbol = vk::any_of_equal(s[1], 'x', 'X');
+  const auto bin_symbol = vk::any_of_equal(s[1], 'b', 'B');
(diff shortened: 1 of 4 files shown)


compiler/lexer.cpp
```diff
@@ -367,10 +367,12 @@ bool TokenLexerNum::parse(LexerData *lexer_data) const {
     binary,
   } state = before_dot;
 
-  if (s[0] == '0' && s[1] == 'x') {
+  const auto hex_symbol = vk::any_of_equal(s[1], 'x', 'X');
+  const auto bin_symbol = vk::any_of_equal(s[1], 'b', 'B');
```
unserialize, line 371: I'd prefer not to perform these calculations, because these comparisons are done every time, for every number — for decimal integers also.
I'd change the code like
```
if (s[0] == '0') {
  if (s[1] == 'x' || s[1] == 'X') {
    t += 2;
    state = hex;
  } else if (s[1] == 'b' || s[1] == 'B') {
    t += 2;
    state = binary;
  }
}
```
What about the last line, nothing bad is to enumerate 4 cases via `||`

quasilyte (approved):

PR #317, opened 2021-09-22
WIP: named arguments

i582:
RFC: https://wiki.php.net/rfc/named_params

#290

### DIFF
--- a/compiler/gentree.cpp
+++ b/compiler/gentree.cpp
@@ -229,6 +229,10 @@ VertexAdaptor<Op> GenTree::get_func_call() {
   string name{cur->str_val};
   next_cur();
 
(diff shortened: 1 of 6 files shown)


compiler/gentree.cpp
```diff
@@ -775,6 +775,16 @@ VertexPtr GenTree::get_binary_op(int op_priority_cur, bool till_ternary) {
     return left;
   }
 
+  // in function named argument call:
+  //   foo(name: <expr>)
+  //           | color here
```
quasilyte, line 780: s/color/colon/?

i582: yes, typo

PR #318, opened 2021-09-24
drop debian jessie support

drdzyk:


CMakeLists.txt
```diff
@@ -1,6 +1,7 @@
 cmake_minimum_required(VERSION 3.14..3.18)
 project(kphp
         VERSION 1.0.1
+        LANGUAGES CXX
```
drdzyk, line 4: Previously here was a room for misconfiguration between c and cpp compilers like in our clang pipeline:
image
Since we don't use c compiler I just remove  it.

quasilyte (approved):

eugene536 (approved):

PR #320, opened 2021-09-27
Prevent job workers from waking up on new job while another one is running

DrDet:
The main idea is to enable `EPOLLONESHOT` flag for fd we read new jobs from and rearm it only when job is already served.

relates to `KPHP-1353`

### DIFF
--- a/net/net-reactor.cpp
+++ b/net/net-reactor.cpp
@@ -245,6 +245,9 @@ static int epoll_conv_flags(int flags) {
   if (!(flags & EVT_LEVEL)) {
     r |= EPOLLET;
   }
+  if (flags & EPOLLONESHOT) {
+    r |= EPOLLONESHOT;
+  }
   return r;
 }
 
@@ -323,7 +326,7 @@ int net_reactor_insert(net_reactor_ctx_t *ctx, int fd, int flags) {
     return 0;
   }
 
-  flags &= EVT_NEW | EVT_NOHUP | EVT_LEVEL | EVT_RWX;
+  flags &= EVT_NEW | EVT_NOHUP | EVT_LEVEL | EVT_RWX | EPOLLONESHOT;
   ev->ready = 0; // !!! this bugfix led to some AIO-related bugs, now fixed with the aid of C_REPARSE flag
   if ((ev->state & (EVT_LEVEL | EVT_RWX | EVT_IN_EPOLL)) == flags + EVT_IN_EPOLL) {
     return 0;
--- a/server/job-workers/job-worker-server.cpp
+++ b/server/job-workers/job-worker-server.cpp
@@ -110,7 +110,7 @@ conn_type_t php_jobs_server = [] {
 
 } // namespace
 
-int JobWorkerServer::job_parse_execute(connection *c) {
+int JobWorkerServer::job_parse_execute(connection *c) noexcept {
   assert(c == read_job_connection);
 
   if (sigterm_on) {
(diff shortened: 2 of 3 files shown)


server/job-workers/job-worker-server.h
```diff
@@ -47,6 +49,8 @@ class JobWorkerServer : vk::not_copyable {
   bool reply_is_expected() const noexcept;
 
 private:
+  static const int EPOLL_FLAGS;
```
drdzyk, line 52: constexpr static? Maybe place this var directly in source file?

server/job-workers/job-worker-server.cpp
```diff
@@ -129,10 +133,13 @@ int JobWorkerServer::job_parse_execute(connection *c) {
     assert(errno == EWOULDBLOCK);
     // another job worker has already taken the job (all job workers are readers for this fd)
     // or there are no more jobs in pipe
+    tvkprintf(job_workers, 3, "No jobs in pipe after wakeup\n");
     ++vk::singleton<SharedMemoryManager>::get().get_stats().job_worker_skip_job_due_steal;
+    rearm_read_job_fd(); // because > 1 workers can wake up on single job
```
drdzyk, line 138: I think it worth to wrap rearm_read_job_fd() call in some RAII class. Currently it is possible forget to made such call and fd will hang.

net/net-reactor.cpp
```diff
@@ -323,7 +326,7 @@ int net_reactor_insert(net_reactor_ctx_t *ctx, int fd, int flags) {
     return 0;
   }
 
-  flags &= EVT_NEW | EVT_NOHUP | EVT_LEVEL | EVT_RWX;
+  flags &= EVT_NEW | EVT_NOHUP | EVT_LEVEL | EVT_RWX | EPOLLONESHOT;
```
drdzyk, line 329: Maybe add own flag for EPOLLONESHOT as it done for e.g. EVT_READ <-> EPOLLIN?

drdzyk (changes requested):

server/job-workers/job-worker-server.h
DrDet: ok, done

net/net-reactor.cpp
DrDet: I think it's strange that we have own flags which map to library flags (EVT_READ <-> EPOLLIN and so on). So, I think it's better not to create new own flags. For example, EPOLLEXCLUSIVE doesn't have other own flag too

server/job-workers/job-worker-server.cpp
DrDet: done

net/net-reactor.cpp
drdzyk: Sure

drdzyk (dismissed):

server/job-workers/job-worker-server.h
```diff
@@ -36,6 +36,8 @@ class JobWorkerServer : vk::not_copyable {
 
   void init();
 
+  void rearm_read_job_fd();
```
unserialize, line 39: noexcept?

server/job-workers/job-worker-server.cpp
```diff
@@ -188,6 +203,12 @@ void JobWorkerServer::init() {
   job_reader = PipeJobReader{read_job_fd};
 }
 
+
+void JobWorkerServer::rearm_read_job_fd() {
+  // We need to rearm fd because we use EPOLLONESHOT
+  epoll_insert(read_job_fd, EPOLL_FLAGS);
+}
+
```
unserialize, line 211: Seems like an extra newline

DrDet: fixed whitespaces

server/job-workers/job-worker-server.h
DrDet: added for all of the JobWorkerServer functions

unserialize (approved):

PR #325, opened 2021-09-30
forbid bcmath to return negative zero numbers

drdzyk:
bcmath currently able to return numbers like -0, -0.0 etc. This behavior was fixed in php 7.4.23, 8.0.10, see https://bugs.php.net/bug.php?id=78238 for more details.

### DIFF
--- a/tests/phpt/dl/483_bcmath.php
+++ b/tests/phpt/dl/483_bcmath.php
@@ -108,7 +108,7 @@
   var_dump (bcdiv("-645345345.88", "0.1", 6));
 
   $a = array ("0.0", "-1.111111", "1.12312312", "-1.111111", "1.12312312", "1.111111", "-1.12312312", "1000101010101", "9999999999999", "0000.1111111", "-123123213213.1", "-1");
-  $b = array (0, 1, 2, 3, 4, 5, 10, 20, 100);
+  $b = array (2, 3, 4, 5, 10, 20, 100);
   foreach ($a as $u)
     foreach ($a as $v) 
       foreach ($b as $scale) {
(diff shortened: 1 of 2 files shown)


DrDet (changes requested):
I think it would be much better to keep it in phpt tests, and not to mix it up with end-to-end tests for HTTP server.
Let's do it in 483_bcmath.php. For those tests which return 0 or -0 in different php versions it will look like this:
```
  #ifndef KPHP
  if (version_compare(PHP_VERSION, "7.4.23") >= 0) {
  #endif
    bcadd("0", "-0.1"); // 0
    bcmul("0.999999999", "-0.111111111111"); // 0
  #ifndef KPHP
  } else {
    echo "0\n";
    echo "0\n";
  }
  #endif
```

drdzyk:
> I think it would be much better to keep it in phpt tests, and not to mix it up with end-to-end tests for HTTP server. Let's do it in 483_bcmath.php. For those tests which return 0 or -0 in different php versions it will look like this:
> 
> ```
>   #ifndef KPHP
>   if (version_compare(PHP_VERSION, "7.4.23") >= 0) {
>   #endif
>     bcadd("0", "-0.1"); // 0
>     bcmul("0.999999999", "-0.111111111111"); // 0
>   #ifndef KPHP
>   } else {
>     echo "0\n";
>     echo "0\n";
>   }
>   #endif
> ```

Cool! fixed

tests/phpt/dl/483_bcmath.php
```diff
@@ -121,3 +121,43 @@
...
+  In new php versions and in kphp bcmath never returns numbers like -0, -0.0 (but returns 0, 0.0).
+  See https://bugs.php.net/bug.php?id=78238 for more details.
+*/
+  #ifndef KPHP
+  if ((version_compare(PHP_VERSION, "7.4.23") >= 0) &&
+       (version_compare(PHP_VERSION, "8.0.0") < 0) ||
+       (version_compare(PHP_VERSION, "8.0.10") >= 0)) {
+  #endif
+    var_dump(bcadd("0", "-0.1"));    // 0
+    var_dump(bcadd("0", "-0.1", 1)); // -0.1
```
DrDet, line 136: Why do we need to keep tests returning non-zero here? If I'm not mistaken, their behavior must be the same in different php versions and kphp

drdzyk: In order to test that code near is not broken.

DrDet (approved):

PR #328, opened 2021-10-05
Increasing the request body limit to 2 MB

troy4eg:
Also fixed mac os test build


DrDet (commented):
Let's add an end-to-end test for http server (here `tests/python/tests/http_server`) and ensure that:
1. before these changes the test failed
2. after these changes the test passed

PR #338, opened 2021-10-16
added compile-time error about undefined constant

### LINKED ISSUE #238: Confusing error message for undefined constants
```php
<?php

var_dump(UNDEFINED_CONST);
```
KPHP error:
```
Compilation error at stage: Preprocess function, gen by preprocess-function.cpp:426
  hello.php:3
    var_dump(UNDEFINED_CONST);

Unexpected function pointer: UNDEFINED_CONST
```
Maybe we need to mention that it could also be a result of undefined const:
```
Unexpected function pointer or undefined const: UNDEFINED_CONST
```
Or rethink this error message completely (do we need to support this kind of unquoted func names at all?).

i582:
In `functions.txt`:
- Added missing `PHP_INT_MIN`
- Added `TODO`, `TODO_OVERLOAD` placeholder constants
- Renamed `INT_MIN` and `INT_MAX` with `PHP_INT_MIN` and `PHP_INT_MAX`

Example:
```
Compilation error at stage: Inline defines pass, gen by inline-defines-usages.cpp:37
  test.php:18  in foo
    echo UNDEFINED_CONSTANT;

Undefined constant 'UNDEFINED_CONSTANT'
```

Fixes #238

### DIFF
--- a/functions.txt
+++ b/functions.txt
@@ -1,6 +1,11 @@
 <?php
 
+// placeholder constants
+define('TODO', -1);
+define('TODO_OVERLOAD', -1);
+
 define('PHP_INT_MAX', 9223372036854775807);
+define('PHP_INT_MIN', -9223372036854775808);
 define('PHP_INT_SIZE', 8);
 define('PHP_EOL', "\n");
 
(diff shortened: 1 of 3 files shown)


functions.txt
```diff
@@ -1,6 +1,11 @@
 <?php
 
+// placeholder constants
+define('TODO', -1);
+define('TODO_OVERLOAD', -1);
```
DrDet, line 5: Why do we need these constants?

DrDet (approved):

functions.txt
i582: There are places in `functions.txt` where we use them. And their use there seems to be necessary in order to organize a kind of overloading for the function.

i582: > Why do we need these constants?

For example:

```php
function mt_rand ($l ::: int = TODO_OVERLOAD, $r ::: int = TODO_OVERLOAD) ::: int;
```

DrDet: ok, nice

DrDet (approved):

PR #344, opened 2021-10-25
wait_queue_size function

vajexal:
This function could be useful for job buffering

For example
```php
<?php

function foo(): int {
    sched_yield_sleep(0.001);

    return mt_rand(1, 10);
}

$queue = wait_queue_create([]);

function process(): void {
    global $queue;

    $ready_fork_id = wait_queue_next($queue);
    $result = wait($ready_fork_id);

    echo $result, PHP_EOL;
}

for ($i = 0; $i < 100; $i++) {
    $fork_id = fork(foo());
    wait_queue_push($queue, $fork_id);

    while (wait_queue_size($queue) >= 10) {
        process();
    }
}

while (!wait_queue_empty($queue)) {
    process();
}
```

### DIFF
--- a/docs/kphp-language/best-practices/async-programming-forks.md
+++ b/docs/kphp-language/best-practices/async-programming-forks.md
@@ -344,6 +344,10 @@ Adds a fork to an existing queue; if *T2!=T*, leads to re-inferring *T=lca(T,T2)
 
 Returns if this queue is empty (no forks to wait).
 
+<aside>wait_queue_size( future_queue&lt;any&gt; ): int</aside>
+
+Returns queue size.
+
 <aside>wait_queue_next( future_queue&lt;T&gt; ): future&lt;T&gt;|false (resumable)</aside>
 
 Stops the current fork until any result of the queue is ready and lets other forks to run.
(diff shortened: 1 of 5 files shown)


DrDet (commented):
Hello! Nice to see such an involved contributor here!

Let's try to define what wait queue size is. I think it is a number of forks that were added to wait queue and not waited (popped) still.
Consider the following actions:
1. fork is called.
2. fork is added to wait queue
3. fork is finished
4. result of fork is taken by `wait`

At the end of step 3 fork is still in wait queue, until someone waits it at step 4. 
But your implementation doesn't fit this. `q->left_functions` is number of forks that weren't finished yet (every finished fork can be not waited yet)

So it's possible when `wait_queue_size($q) === 0` but `wait_queue_empty($q) === false`. Here is an example:
```php
<?php

function foo(int $i): int {
    fwrite(STDERR, "~~~ foo($i) before\n");
    if ($i == -1) {
      sched_yield_sleep(0.1);
    } else {
      sched_yield_sleep(0.001);
    }
    fwrite(STDERR, "~~~ foo($i) after\n");
    return mt_rand(1, 10);
}

$queue = wait_queue_create([]);

function process(): void {
    global $queue;

    $ready_fork_id = wait_queue_next($queue);
    $result = wait($ready_fork_id);

//     echo $result, PHP_EOL;
}

$long_fork = fork(foo(-1));

for ($i = 0; $i < 5; $i++) {
    $fork_id = fork(foo($i));
    wait_queue_push($queue, $fork_id);
}

usleep(2000);
wait($long_fork);

var_dump(wait_queue_size($queue));
var_dump(wait_queue_empty($queue));
```

vajexal:
sorry, i don't think i have such kphp internals knowledge so i'll close this pr

PR #350, opened 2021-10-29
fixate minimum required cpu microarchitecture to sandybridge

drdzyk:


compiler/compiler-settings.cpp
```diff
@@ -273,6 +273,7 @@ void CompilerSettings::init() {
      << " -iquote " << kphp_src_path.get() << "objs/generated/auto/runtime";
   ss << " -Wall -fwrapv -Wno-parentheses -Wno-trigraphs";
   ss << " -fno-strict-aliasing -fno-omit-frame-pointer";
+  ss << " -march=sandybridge";
```
AlexK0, line 276: nit: It looks too tough, roughly speaking you are forcing this options for all builds in any architectures. I believe, that **get_default_extra_cxxflags()** is more sutable plase for this option (because it can be overridden).

drdzyk: Yea, its true. There are some reasons for it. First, there was a story that on production occasionally used -march=core2 that we don't expect. Second, sandybridge is not modern arch actually and we expect that it is ok require it by default, in same way we expect c++ compiler for kphp to be able in c++17. And lastly, fixating arch allow us to use it's new instructions set explicitly in code.

PR #351, opened 2021-10-29
Fixed deprecated 'vfork()' function call for mac os 12

troy4eg:


compiler/make/make-runner.cpp
```diff
@@ -96,7 +96,12 @@ static int run_cmd(const string &cmd) {
   }
   argv.back() = nullptr;
 
+  #ifdef __APPLE__
+  int pid = fork();
```
AlexK0, line 100: nit: it seems to me, that you can use **fork()** fo linux and other platforms as well

unserialize (approved):

PR #361, opened 2021-11-09
Move functions*.txt files to separate folder

DrDet:
Now you need to manually modify install rule in CMake file on adding new functions*.txt files. It's too easy to forget about it and get problem with deb package distribution (It's already happened twice).
Moreover, the single folder is much more elegant way than several files with name pattern like "functions_*.txt".


unserialize (approved):

i582 (approved):

quasilyte:
Cool!

PR #371, opened 2021-11-25
Use inheritance and virtual methods instead of C-style manual emulation in php queries

DrDet:
The main change is:

image

### DIFF
--- a/server/php-queries-types.cpp
+++ b/server/php-queries-types.cpp
@@ -12,7 +12,7 @@
 #include "server/php-mc-connections.h"
 #include "server/php-sql-connections.h"
 
-void php_query_x2_t::run(php_worker *worker __attribute__((unused))) noexcept {
+void php_query_x2_t::run([[maybe_unused]] php_worker *worker) noexcept {
   query_stats.desc = "PHPQX2";
 
   php_script_query_readed(php_script);
@@ -43,7 +43,7 @@ void php_query_rpc_answer::run(php_worker *worker) noexcept {
   php_script_query_answered(php_script);
 }
 
-void php_query_connect_t::run(php_worker *worker __attribute__((unused))) noexcept {
+void php_query_connect_t::run([[maybe_unused]] php_worker *worker) noexcept {
   query_stats.desc = "CONNECT";
 
   php_script_query_readed(php_script);
@@ -79,8 +79,6 @@ void php_query_wait_t::run(php_worker *worker) noexcept {
 
 
 int php_worker_http_load_post_impl(php_worker *worker, char *buf, int min_len, int max_len) {
-  assert (worker != nullptr);
-
   connection *c = worker->conn;
   double precise_now = get_utime_monotonic();
 
--- a/server/php-queries-types.cpp
+++ b/server/php-queries-types.cpp
@@ -1,0 +1,197 @@
+// Compiler for PHP (aka KPHP)
+// Copyright (c) 2021 LLC «V Kontakte»
+// Distributed under the GPL v3 License, see LICENSE.notice.txt
+
+#include <poll.h>
+
+#include "common/kprintf.h"
+#include "server/php-queries-types.h"
+#include "server/php-worker.h"
+#include "server/php-runner.h"
+#include "server/php-engine.h"
+#include "server/php-mc-connections.h"
+#include "server/php-sql-connections.h"
+
+void php_query_x2_t::run(php_worker *worker __attribute__((unused))) noexcept {
+  query_stats.desc = "PHPQX2";
+
+  php_script_query_readed(php_script);
+
+  //  worker->conn->status = conn_wait_net;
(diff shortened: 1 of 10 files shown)


server/php-queries-types.h
```diff
@@ -0,0 +1,69 @@
...
+};
+
+struct php_query_rpc_answer : php_query_base_t {
+  const char *data;
+  int data_len;
+
+  void run(php_worker *worker) noexcept final;
+};
+
+enum class protocol_type {
```
unserialize, line 30: Move enum above structs? Seems strange that it's in the middle

server/server.cmake
```diff
@@ -21,7 +21,9 @@ prepend(KPHP_SERVER_SOURCES ${BASE_DIR}/server/
         server-log.cpp
         server-stats.cpp
         slot-ids-factory.cpp
-        workers-control.cpp)
+        workers-control.cpp
+        php-queries-types.cpp
```
unserialize, line 25: I suppose, this list should remain sorted

server/php-queries-types.cpp
```diff
@@ -0,0 +1,197 @@
...
+  query_stats.desc = "WAIT_NET";
+  php_script_query_readed(php_script);
+  php_script_query_answered(php_script);
+
+  php_worker_wait(worker, timeout_ms);
+}
+
+
+int php_worker_http_load_post_impl(php_worker *worker, char *buf, int min_len, int max_len) {
+  assert (worker != nullptr);
```
unserialize, line 82: This assert is not excessive? Or at least, should be moved to `php_worker_run_query`

server/php-queries-types.cpp
```diff
@@ -0,0 +1,197 @@
...
+
+void php_query_x2_t::run(php_worker *worker __attribute__((unused))) noexcept {
+  query_stats.desc = "PHPQX2";
+
+  php_script_query_readed(php_script);
+
+  //  worker->conn->status = conn_wait_net;
+  //  worker->conn->pending_queries = 1;
+
+  static php_query_x2_answer_t res;
```
unserialize, line 23: Why static?

server/php-queries-types.cpp
```diff
@@ -0,0 +1,197 @@
...
+
+#include "common/kprintf.h"
+#include "server/php-queries-types.h"
+#include "server/php-worker.h"
+#include "server/php-runner.h"
+#include "server/php-engine.h"
+#include "server/php-mc-connections.h"
+#include "server/php-sql-connections.h"
+
+void php_query_x2_t::run(php_worker *worker __attribute__((unused))) noexcept {
```
drdzyk, line 15: c++17 [[maybe_unused]] ?

server/php-queries-types.cpp
drdzyk: would be even better to pass php_worker type by reference.

server/php-queries-types.h
```diff
@@ -0,0 +1,69 @@
+// Compiler for PHP (aka KPHP)
+// Copyright (c) 2021 LLC «V Kontakte»
+// Distributed under the GPL v3 License, see LICENSE.notice.txt
+
+#pragma once
+
+struct php_worker;
+
+struct php_query_base_t {
+  void *ans;
```
drdzyk, line 10: void *ans{nullptr} please, here and below

server/php-queries.cpp
```diff
@@ -46,14 +46,13 @@ php_query_http_load_post_answer_t *php_query_http_load(char *buf, int min_len, i
...
-  q.base.type = PHPQ_HTTP_LOAD_POST;
   q.buf = buf;
   q.min_len = min_len;
   q.max_len = max_len;
 
-  PHPScriptBase::current_script->ask_query((void *)&q);
+  PHPScriptBase::current_script->ask_query(&q);
 
-  return (php_query_http_load_post_answer_t *)q.base.ans;
+  return (php_query_http_load_post_answer_t *)q.ans;
```
drdzyk, line 55: can we avoid c-style cast? Check other places too please

drdzyk (changes requested):

server/php-queries-types.cpp
DrDet: > 
moved assert to `php_worker_main`

DrDet: > 
php_worker * actually should be `this`, and all these functions should be methods of `php_worker` class. But let's leave it for separate PR

server/php-queries-types.cpp
DrDet: Because pointer to `res` is stored to `ans` below and later used after returning from this function and swapping context back

server/php-queries-types.cpp
DrDet: done

unserialize (approved):

drdzyk (approved):

PR #375, opened 2021-11-30
revert xgboost lib integration

drdzyk:
xgboost didn't work in production: it caused some troubles. Probably, they are related to interprocess bugs inside xgboost kernel. Probably, shared memory is corrupted. After all, we couldn't invesigate the reason.


server/php-engine.cpp
```diff
@@ -2027,10 +2026,6 @@ int main_args_handler(int i, const char *long_option) {
       use_utf8();
       return 0;
     }
-    case 2025: {
-      vk::singleton<vk::xgboost::ModelPath>::get().path = optarg;
```
unserialize, line 2031: Let's leave an option for future experiments

unserialize (approved):

server/php-engine.cpp
drdzyk: done

cmake/init-compilation-flags.cmake
```diff
@@ -17,6 +17,8 @@ cmake_print_variables(CMAKE_CXX_STANDARD)
 
 if(APPLE)
     include_directories(/usr/local/include)
+    add_definitions(-D_XOPEN_SOURCE)
```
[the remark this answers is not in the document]
drdzyk: without this macos build fails. So need to leave it as is.

unserialize (approved):

PR #377, opened 2021-11-30
support compilation for Apple M1 Silicon

### LINKED ISSUE #330: Support Apple M1 chipset
Now, compiling kphp from sources fails on M1. `homebrew` has other paths, M1 doesn't support SIMD and several similar problems.
Finally, I got it compiled (basically by commenting cpp code), but couldn't get it work. It just hangs around `swap_context()`, even in cli mode.
I found [this library](https://github.com/kaniini/libucontext) on Github, but couldn't get in compiled also.

unserialize:
This PR brings an ability for KPHP to be **compiled and run** for Mac OS with M1 chipset (fixes #330).

M1 is an arm-based CPU, it doesn't have simd/sse instrictions equal to x86. So, from the first side, we should hide (via `#ifndef`) x86-based implementations of some algorithms and provide generic ones.

Another problem is that `makecontext()` and similar that are used internally can be compiled, but don't work (they just hang). To overcome this, we use a third-party library [libucontext](https://github.com/kaniini/libucontext). From now, we call `makecontext_portable` and others, that are macro replaced with native `makecontext` on x86 and `libucontext_makecontext` on M1. 

I couldn't find a way to intergrate libucontext to cmake build, that's why I provide an instruction, how to patch and build it, see the docs. Some paths related to libucontext are hardcoded, commented by the same link to documentation.

Crashdump description is not supported. It's not a big deal, since M1 is used for development purposes only.

Also, now we still support aarch64 in aes and crc32. M1 and aarch64 are different in fact (M1 is `arm64` more precisely). Partial support of aarch64 seems strange (these are historical sources), I'll probably drop it in the near future.


quasilyte (approved):

common/algorithms/simd-int-to-string.h
```diff
@@ -259,3 +268,30 @@ inline char *simd_int64_to_string(int64_t value, char *out_buffer) {
   }
   return simd_uint64_to_string(u, out_buffer);
 }
+
+#else
+// as written above, the same functions for M1 are just implemented without simd
```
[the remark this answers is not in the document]
unserialize: done )

server/ucontext-portable.h
```diff
@@ -0,0 +1,37 @@
+// Compiler for PHP (aka KPHP)
+// Copyright (c) 2021 LLC «V Kontakte»
+// Distributed under the GPL v3 License, see LICENSE.notice.txt
+
+#ifndef UCONTEXT_PORTABLE_H
+#define UCONTEXT_PORTABLE_H
```
DrDet, line 6: Why not `#pragma once` as usual? I think it would be more clear

server/php-runner.h
```diff
@@ -4,7 +4,7 @@
 
 #pragma once
 
-#include <ucontext.h>
+#include "ucontext-portable.h"
```
DrDet, line 7:
```suggestion
#include "server/ucontext-portable.h"
```

server/ucontext-portable.h
unserialize: done

server/php-runner.h
unserialize: done

DrDet (approved):

PR #378, opened 2021-12-02
implement array_merge_recursive()

drdzyk:


unserialize:
Can you please rebase this MR onto current master (to make it compilable for M1)?

drdzyk:
> Can you please rebase this MR onto current master (to make it compilable for M1)?

done

DrDet (commented):
It's strange, but this code
```
<?php

function demo() {
    $result = array_merge_recursive([0,1,2], [3,4]);
    print_r($result);
}

demo();
```
fails on gcc like this:
image

The same code with `array_merge` works. After extracting the const arrays to variables it works too

drdzyk:
Yea, nice catch. It turns out that @kphp-pure-function tag solve this problem. Fixed.

DrDet (approved):

PR #384, opened 2021-12-08
fixed `string::try_to_float_as_php8` function

i582:
Fixed behavior when an empty string was considered a valid number, which led
to incorrect runtime warnings about the difference in behavior between PHP 7
and 8.

Added a new type of test `@kphp_runtime_should_not_warn` for tests that need to
make sure that the code in it does not generate warnings at runtime.

### DIFF
--- a/tests/phpt/migration_php8/009_blank_no_warning.php
+++ b/tests/phpt/migration_php8/009_blank_no_warning.php
@@ -1,4 +1,4 @@
-@kphp_runtime_should_not_warn
+@ok kphp_runtime_should_not_warn
 <?php
 
 require_once 'kphp_tester_include.php';
--- a/tests/phpt/migration_php8/009_blank_no_warning.php
+++ b/tests/phpt/migration_php8/009_blank_no_warning.php
@@ -1,0 +1,20 @@
+@kphp_runtime_should_not_warn
+<?php
+
+require_once 'kphp_tester_include.php';
+
+set_migration_php8_warning(0b011);
+
+$blank = " ";
+$blank_more = "   \t";
+
+var_dump(is_numeric($blank));
+var_dump(is_numeric($blank_more));
+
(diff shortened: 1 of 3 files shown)


tests/phpt/migration_php8/009_blank_no_warning.php
```diff
@@ -0,0 +1,20 @@
+@kphp_runtime_should_not_warn
```
unserialize, line 1: I think, it has both tags?  
```
@ok kphp_runtime_should_not_warn
```
(need to test whether such an annotation would fit your expectations)

tests/phpt/migration_php8/009_blank_no_warning.php
```diff
@@ -0,0 +1,20 @@
...
+<?php
+
+require_once 'kphp_tester_include.php';
+
+set_migration_php8_warning(0b011);
+
+$blank = " ";
+$blank_more = "   \t";
```
unserialize, line 9: A string with leading/trailing spaces like " 42" works and is tested in other tests already, right?

i582: Yep, in `006_is_numeric.php`

tests/phpt/migration_php8/009_blank_no_warning.php
i582: It seems `@ok` will be superfluous here, because if there are any compilation errors, then `@kphp_runtime_should_not_warn` will catch them and the test will fail.

unserialize: It's confusing that an ok test is not marked with @ok, imho :)

i582: Added

unserialize (approved):

PR #386, opened 2021-12-08
fix caused by rule of five

drdzyk:
Minor fix caused by c++ rule of five, and further code cleanup.

### DIFF
--- a/compiler/data/vertex-adaptor.h
+++ b/compiler/data/vertex-adaptor.h
@@ -49,13 +49,11 @@ public:
   }
 
   template<Operation FromOp>
-  VertexAdaptor &operator=(const VertexAdaptor<FromOp> &from) & {
+  VertexAdaptor &operator=(const VertexAdaptor<FromOp> &from) {
     static_assert(op_type_is_base_of(Op, FromOp), "Strange assignment to not base vertex");
     impl = static_cast<vertex_inner<Op> *> (from.impl);
     return *this;
   }
-
-  VertexAdaptor &operator=(const VertexAdaptor<Op> &) & = default;
 
   explicit operator bool() const {
     return impl != nullptr;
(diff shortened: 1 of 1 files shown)


compiler/data/vertex-adaptor.h
```diff
@@ -7,170 +7,159 @@
...
-class Location;
-
 using VertexPtr = VertexAdaptor<meta_op_base>;
 
 VertexPtr clone_vertex(VertexPtr);
 
 template<Operation Op>
 class VertexAdaptor {
-  vertex_inner<Op> *impl;
+  vertex_inner<Op> *impl_{nullptr};
```
unserialize, line 25: If you have renamed this field, please update lldb_addons.py to remain pretty printers work

compiler/data/vertex-adaptor.h
```diff
@@ -7,170 +7,159 @@
...
   }
 
   bool operator!=(const VertexAdaptor<Op> &other) const noexcept {
-    return impl != other.impl;
+    return impl_ != other.impl_;
   }
 
-  VertexAdaptor &set_rl_type(RLValueType valueType) {
-    impl->rl_type = valueType;
+  VertexAdaptor &set_rl_type(RLValueType valueType) noexcept {
```
unserialize, line 130: It seems inconsistent, that here and below we write just `VertexAdaptor` (no template instantiation), whereas above we write `VertexAdaptor<Op>`

compiler/data/vertex-adaptor.h
drdzyk: done

compiler/data/vertex-adaptor.h
drdzyk: done

PR #390, opened 2021-12-13
forbid bad types in indexing

drdzyk:
Forbid bad types in indexing arrays, strings and tuples

### DIFF
--- a/compiler/pipes/final-check.cpp
+++ b/compiler/pipes/final-check.cpp
@@ -2,10 +2,13 @@
 // Copyright (c) 2020 LLC «V Kontakte»
 // Distributed under the GPL v3 License, see LICENSE.notice.txt
 
+#include <string_view>
+
 #include "compiler/pipes/final-check.h"
 
 #include "common/termformat/termformat.h"
 #include "common/algorithms/string-algorithms.h"
+#include "common/algorithms/contains.h"
 
 #include "compiler/compiler-core.h"
 #include "compiler/data/src-file.h"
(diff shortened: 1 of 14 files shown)


compiler/pipes/final-check.cpp
```diff
@@ -527,6 +543,46 @@ void FinalCheckPass::check_instanceof(VertexAdaptor<op_instanceof> instanceof_ve
...
+                        what_indexing.empty() ? "" : " ", key_type.as_human_readable()));
+}
+
+static void check_indexing_type(PrimitiveType indexed_type, const TypeData &key_type) noexcept {
+  if (key_type.get_real_ptype() == tp_bool) {
+    return;
+  }
+
+  if (indexed_type == tp_tuple || indexed_type == tp_string) {
+    static const std::vector<PrimitiveType> allowed_types{tp_int, tp_float, tp_mixed, tp_future};
```
unserialize, line 559: I suppose, that for strings and tuples, we'd better allow only tp_int and tp_mixed: $s[1.0] seems unreasonable

compiler/pipes/final-check.cpp
```diff
@@ -527,6 +543,46 @@ void FinalCheckPass::check_instanceof(VertexAdaptor<op_instanceof> instanceof_ve
...
+    check_indexing_violation("int, float and future", allowed_types, ptype_name(indexed_type), key_type);
+  } else {
+    static const std::vector<PrimitiveType> allowed_types{tp_int, tp_float, tp_string, tp_mixed, tp_future};
```
unserialize, line 562: Is tp_float reasonable for arrays indexing?

compiler/pipes/final-check.cpp
drdzyk: Done

compiler/pipes/final-check.cpp
drdzyk: To me not, but there are bunch of compilation issues in production. After discussion we decided that better leave as is.

compiler/pipes/final-check.cpp
```diff
@@ -509,6 +515,16 @@ VertexPtr FinalCheckPass::on_enter_vertex(VertexPtr vertex) {
     check_instanceof(vertex.try_as<op_instanceof>());
   }
 
+  if (vertex->type() == op_index) {
+    const auto index_vertex = vertex.as<op_index>();
+    check_indexing(index_vertex->array(), index_vertex->key());
```
unserialize, line 520: Don't you want to use .try_as? Instread of checking type and casting:
```
if (const auto v_index = vertex.try_as<op_index>()) {
  check_indexing(v_index->array(), v_index->key());
} 
// and below
```

drdzyk: Done

DrDet (approved):

unserialize (approved):

PR #391, opened 2021-12-16
Use `std::variant` in `net_events` and `net_queries` instead of self-made union based struct

DrDet:
Before this PR, `std::variant` behavior was emulated via non-standard gcc extension -- [anonymous aggregate](https://gcc.gnu.org/onlinedocs/gcc/Unnamed-Fields.html). It's not allowed to have field of type with user-defined constructor in anonymous aggregates, which prevents making code be more like C++, than C.
This PR fixes the issue.

### DIFF
--- a/server/php-queries.cpp
+++ b/server/php-queries.cpp
@@ -4,6 +4,7 @@
 
 #include "server/php-queries.h"
 
+#include <array>
 #include <cassert>
 #include <cstdarg>
 #include <cstdio>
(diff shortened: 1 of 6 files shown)


server/php-queries.h
```diff
@@ -18,51 +19,46 @@ struct FinishedJob;
...
 } // namespace job_workers
 
-enum class net_event_type_t {
-  rpc_answer,
-  rpc_error,
-  job_worker_answer,
+namespace net_events_data {
+
+struct rpc_answer {
+  int result_len;
```
drdzyk, line 25: Please add member initializers here and below.

server/php-queries.h
```diff
@@ -18,51 +19,46 @@ struct FinishedJob;
...
-  };
+  slot_id_t slot_id;
+  std::variant<net_events_data::rpc_answer, net_events_data::rpc_error, net_events_data::job_worker_answer> data;
 
   const char *get_description() const noexcept;
 };
 
-enum net_query_type_t {
-  nq_rpc_send
+namespace net_queries_data {
```
drdzyk, line 48: namespace with only one type looks a bit strange, maybe combine it with another one above? Or remove it all?

server/php-queries.h
```diff
@@ -18,51 +19,46 @@ struct FinishedJob;
...
   slot_id_t slot_id;
-  union {
-    struct { //nq_rpc_send
-      int host_num;
-      char *request;
-      int request_size;
-      int timeout_ms;
-    };
-  };
+  std::variant<net_queries_data::rpc_send> data;
```
drdzyk, line 61: variant with one type? Looks like vairant is excessive here.

server/php-queries.cpp
```diff
@@ -1089,23 +1086,20 @@ void php_queries_finish() {
 
 const char *net_event_t::get_description() const noexcept {
   static char BUF[10000];
```
drdzyk, line 1088: can u use std::array here pls?

server/php-queries.cpp
```diff
@@ -1089,23 +1086,20 @@ void php_queries_finish() {
...
-    }
-    case net_event_type_t::rpc_error: {
-      sprintf(BUF, "RPC ERROR: error code = %d, error message = %s", error_code, error_message);
-      break;
-    }
-    case net_event_type_t::job_worker_answer: {
-      if (job_result) {
-        sprintf(BUF, "JOB RESPONSE: class name = %s", job_result->response.get_class());
+  std::visit(overloaded{
+    [](const net_events_data::rpc_answer &e) {
```
drdzyk, line 1090: clang-tidy complains like:
"Parameter name 'e' is too short, expected at least 3 characters"
can we fix that?

drdzyk (changes requested):

server/php-queries.h
DrDet: Second type will be added in this PR: https://github.com/VKCOM/kphp/pull/393

server/php-queries.h
DrDet: Second type will be added in this PR: https://github.com/VKCOM/kphp/pull/393

server/php-queries.h
DrDet: done

server/php-queries.cpp
DrDet: done

server/php-queries.cpp
DrDet: fixed

server/php-queries.h
drdzyk: got it

server/php-queries.h
drdzyk: got it

drdzyk (approved):

PR #393, opened 2021-12-19
Resumable-friendly external database drivers embedding API & PDO::MySQL resumable basic version

DrDet:
The main goal of the PR is to introduce internal _resumable-friendly_ API for embedding asynchronous database drivers (MySQL, PostgreSQL, Redis etc.)

_Resumable-friendly_ means that any network communication based on this API will be considered as a suspendable point by KPHP resumable functions (a.k.a. coroutines). To achieve this the API deeply interacts with KPHP web server network internals and epoll reactor. 

The API provides the following abstractions:
- `database_drivers::Connector` -- class representing connection to database.
Specific database driver is supposed to extend this class and implement all pure virtual methods.
The most important of them:
  - Obtaining file descriptor
      ```c++
      /**
      * @brief Gets file descriptor number of underlying connection.
      * @return File descriptor number.
      */
      virtual int get_fd() const noexcept = 0;
      ```
  - Performing connection itself
      ```c++
      /**
       * @brief Attempts to establish underlying connection asynchronously.
       * @return Status of operation: in progress, completed or error.
       */
      virtual AsyncOperationStatus connect_async() noexcept = 0;
      ```
  - Methods for handling epoll events of different types:
      ```c++
      virtual void handle_write() noexcept = 0;
      virtual void handle_read() noexcept = 0;
      virtual void handle_special() noexcept = 0;
      ```
   See all such methods at `server/database-drivers/connector.h` 
  > **Note:** It's only for not persistent connections, e.g. bound to script memory. All such connections will be closed at the end of script execution.

- `database_drivers::Request` -- class representing request to database. It has the only method for sending itself asynchronously:
    ```c++
    /**
     * @brief Sends this requests asynchronously.
     * @return Status of operation: in progress, completed or error.
     */
    virtual AsyncOperationStatus send_async() noexcept = 0;
    ```
  Should be called from `database_drivers::Connector::handle_write()`
- `database_drivers::Response` -- class representing response from database. It has the only method for fetching itself asynchronously:
    ```c++
    /**
     * @brief Fetches this response asynchronously.
     * @return Status of operation: in progress, completed or error.
     */
    virtual AsyncOperationStatus fetch_async() noexcept = 0;
    ```
  Should be called from `database_drivers::Connector::handle_read()`

And special singleton class to manipulate these abstractions - `database_drivers::Adaptor`. With it you can:
- Connect to database:
    ```c++
    /**
     * @brief Registers @a connector and embeds it into event loop.
     * @param connector
     * @return ID of registered connector.
     */
    int initiate_connect(std::unique_ptr<Connector> &&connector) noexcept;
    ```
- Send request asynchronously and get waitable reumable id:
    ```c++
    /**
     * @brief Launches @a request resumable (coroutine), that will send @a request asynchronously.
     * @param request
     * @return ID of launched resumable (coroutine), which can be awaited with wait_request_resumable().
     *
     * Creates forked resumable as if we call `fork(launch_request_resumable())`, @see fork_resumable().
     * Moves @a request to Connector::push_async_request().
     */
    int launch_request_resumable(std::unique_ptr<Request> &&request) noexcept
    ```
- Wait sent request by resumable id:
    ```c++
    /**
     * @brief Waits request resumable with @a resumable_id.
     * @param resumable_id
     * @return Response for Request associated with @a resumable_id.
     */
    std::unique_ptr<Response> wait_request_resumable(int resumable_id) const noexcept;
    ```
- And more. See the full methods list at `server/database-drivers/adaptor.h`
----
As an example of usage this API, basic resumable **PDO::MySQL** version is implemented in this PR as well.
It supports only basic read and modify operations with default behaviour. But these operations are truly resumable thanks to new API.

> **Note:** This API can be used only with drivers that are more or less asynchronous and expose the way to get file descriptor of connection. Basically, It can be said that if some database driver is compatible with [`libevent`](https://libevent.org/) or similar lib, then it can be embedded to KPHP preserving asynchrony via this API.


.github/workflows/Dockerfile.buster
```diff
@@ -2,19 +2,23 @@ FROM debian:buster
...
+    apt-get install -y --no-install-recommends apt-utils ca-certificates gnupg wget lsb-release && \
     echo "deb https://deb.debian.org/debian buster-backports main" >> /etc/apt/sources.list && \ 
     wget -qO - https://repo.vkpartner.ru/GPG-KEY.pub | apt-key add - && \
     echo "deb https://repo.vkpartner.ru/kphp-buster/ buster main" >> /etc/apt/sources.list && \
     wget -qO - https://packages.sury.org/php/apt.gpg | apt-key add - && \
     echo "deb https://packages.sury.org/php/ buster main" >> /etc/apt/sources.list.d/php.list && \
+    TEMP_DEB="$(mktemp)" && \
+    wget -O "$TEMP_DEB" 'https://dev.mysql.com/get/mysql-apt-config_0.8.22-1_all.deb' && \
+    dpkg -i "$TEMP_DEB" && \
+    rm -f "$TEMP_DEB" && \
```
drdzyk, line 14: just in case - do we need this? If yes, there is no other way to obtain package?

runtime/pdo/mysql/mysql_pdo_driver.cpp
```diff
@@ -0,0 +1,100 @@
...
+#include "runtime/pdo/mysql/mysql_pdo_emulated_statement.h"
+#include "runtime/pdo/pdo.h"
+#include "runtime/pdo/pdo_statement.h"
+#include "server/database-drivers/mysql/mysql.h"
+#include "server/database-drivers/mysql/mysql-connector.h"
+#include "server/database-drivers/mysql/mysql-response.h"
+#include "server/database-drivers/adaptor.h"
+#include "server/php-queries.h"
+
+namespace pdo::mysql {
```
drdzyk, line 19: Cool idea about using namespaces, what about to use it in pdo.h/cpp and pdo_statement.h/cpp too for consistency?

drdzyk (dismissed):

drdzyk (dismissed):

troy4eg (approved):

PR #395, opened 2021-12-21
improved statshouse client

troy4eg:
added almost all statsd metrics

### DIFF
--- a/common/server/stats.cpp
+++ b/common/server/stats.cpp
@@ -13,7 +13,7 @@ int (*tl_stat_function)(const std::optional<std::vector<std::string>> &sorted_fi
 
 static char buf[STATS_BUFFER_LEN];
 
-char* get_engine_default_prepare_stats_buffer() {
+char *get_engine_default_prepare_stats_buffer() {
   return buf;
 }
 
@@ -26,8 +26,8 @@ std::pair<char *, int> engine_default_prepare_stats_with_tag_mask(stats_t &&stat
   return {buf, stats.sb.pos};
 }
 
-char *engine_default_prepare_stats(stats_t &&stats, const char *statsd_prefix) {
-  return engine_default_prepare_stats_with_tag_mask(std::move(stats), statsd_prefix, STATS_TAG_MASK_FULL).first;
+char *engine_default_prepare_stats(stats_t &&stats, const char *stats_prefix) {
+  return engine_default_prepare_stats_with_tag_mask(std::move(stats), stats_prefix, STATS_TAG_MASK_FULL).first;
 }
 
 void engine_default_tl_stat_function(const std::optional<std::vector<std::string>> &sorted_filter_keys) {
--- a/common/server/stats.cpp
+++ b/common/server/stats.cpp
@@ -7,40 +7,38 @@
 #include "common/server/engine-settings.h"
 #include "common/server/tl-stats-t.h"
 #include "common/stats/buffer.h"
-
 #include "common/tl/parse.h"
 
 int (*tl_stat_function)(const std::optional<std::vector<std::string>> &sorted_filter_keys);
 
-char *engine_default_prepare_stats_with_tag_mask(stats_t &&stats, int *len, const char *statsd_prefix, unsigned int tag_mask) {
-  static char buf[1 << 20];
-  stats.statsd_prefix = statsd_prefix;
+static char buf[STATS_BUFFER_LEN];
 
-  sb_init(&stats.sb, buf, 1 << 20);
-  prepare_common_stats_with_tag_mask(&stats, tag_mask);
-
-  if (len) {
-    *len = stats.sb.pos;
-  }
+char* get_engine_default_prepare_stats_buffer() {
   return buf;
 }
 
-char *engine_default_prepare_stats(stats_t &&stats, int *len, const char *statsd_prefix) {
-  return engine_default_prepare_stats_with_tag_mask(std::move(stats), len, statsd_prefix, STATS_TAG_MASK_FULL);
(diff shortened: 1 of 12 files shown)


server/statshouse/statshouse-client.cpp
```diff
@@ -8,20 +8,71 @@
...
 #include "common/resolver.h"
+#include "common/server/stats.h"
+#include "common/stats/provider.h"
 #include "common/tl/constants/statshouse.h"
 #include "common/tl/methods/string.h"
 #include "net/net-connections.h"
 #include "runtime/critical_section.h"
 #include "server/server-log.h"
 
+class statshouse_stats_t : public stats_t {
```
DrDet, line 19: Let's put this class to anonymous namespace to emphasis that it`s private class for this cpp file

server/statshouse/statshouse-client.cpp
```diff
@@ -8,20 +8,71 @@
...
+};
+
+namespace {
+std::pair<char *, int> prepare_statshouse_stats(statshouse_stats_t &&stats, const char *statsd_prefix, unsigned int tag_mask) {
+  stats.stats_prefix = statsd_prefix;
+
+  char *buf = get_engine_default_prepare_stats_buffer();
+
+  sb_init(&stats.sb, buf, STATS_BUFFER_LEN);
+  constexpr int offset = 3 * 4; // for magic, fields_mask and vector size
```
DrDet, line 59:
```suggestion
  constexpr int offset = 3 * sizeof(int); // for magic, fields_mask and vector size
```

common/server/stats.h
```diff
@@ -10,8 +10,11 @@
 
 #include "common/stats/provider.h"
 
-char *engine_default_prepare_stats_with_tag_mask(stats_t &&stats, int *len, const char *statsd_prefix, unsigned int tag_mask);
-char *engine_default_prepare_stats(stats_t &&stats, int *len, const char *statsd_prefix);
+constexpr int STATS_BUFFER_LEN = 1 << 20;
+
+char* get_engine_default_prepare_stats_buffer();
+std::pair<char *, int> engine_default_prepare_stats_with_tag_mask(stats_t &&stats, const char *stats_prefix, unsigned int tag_mask);
+char *engine_default_prepare_stats(stats_t &&stats, const char *statsd_prefix);
```
unserialize, line 17: `char *` here vs `char*` above — not in a style. We prefer `char *`. It's not the only place, btw.

server/statshouse/statshouse-client.cpp
```diff
@@ -8,20 +8,71 @@
...
+    return counter;
+  }
+
+private:
+  int counter{0};
+  const std::vector<std::pair<std::string, std::string>> &tags;
+};
+
+namespace {
+std::pair<char *, int> prepare_statshouse_stats(statshouse_stats_t &&stats, const char *statsd_prefix, unsigned int tag_mask) {
```
unserialize, line 54: You've renamed the field `statsd_prefix` to `stats_prefix`, but here is still `statsd_prefix`. In a function preparing statshouse).
Btw, I'd suggest to remove this parameter at all and just to hardcode "kphp" below

server/statshouse/statshouse-client.cpp
```diff
@@ -8,20 +8,71 @@
...
+} // namespace
+
 void StatsHouseClient::set_port(int value) {
   this->port = value;
 }
 
 void StatsHouseClient::set_host(std::string value) {
   this->host = std::move(value);
 }
-
```
unserialize, line 24: Line removal seems a misprint

server/statshouse/statshouse-client.cpp
```diff
@@ -50,18 +101,26 @@ bool StatsHouseClient::init_connection() {
...
+
+  ssize_t slen = send(sock_fd, result, len, 0);
+  if (slen < 0) {
+    log_server_error("Can't send metrics to statshouse (len = %i): %s", len, strerror(errno));
+  }
+}
+
+StatsHouseClient::StatsHouseClient() {
+  char hostname[128];
+  int res = gethostname(hostname, 128);
```
unserialize, line 120: Is it really good practice to call this functions in a constructor of a global singleton? @DrDet

DrDet: It seems quite reasonable here. What confuses you?

DrDet (approved):

server/statshouse/statshouse-client.cpp
unserialize: I don't understand, when this constructor will be called. Is there a chance, that `gethostname()` hangs, and master process will hang before `main()`?

DrDet: It will be called on the first `vk::singleton<...>::get()` call which happens on parsing host:port option. I think it never hangs because it doesn't use DNS, I think it just reads some file

DrDet (approved):

common/server/stats.cpp
```diff
@@ -7,40 +7,38 @@
...
-  if (len) {
-    *len = stats.sb.pos;
-  }
-  return buf;
+  return {buf, stats.sb.pos};
 }
 
-char *engine_default_prepare_stats(stats_t &&stats, int *len, const char *statsd_prefix) {
-  return engine_default_prepare_stats_with_tag_mask(std::move(stats), len, statsd_prefix, STATS_TAG_MASK_FULL);
+char *engine_default_prepare_stats(stats_t &&stats, const char *statsd_prefix) {
```
unserialize, line 29: here `statsd_prefix` is still left

unserialize (approved):

PR #398, opened 2021-12-25
speed up tagger.cpp compilation

drdzyk:
Currently autogenerated _tagger.cpp file consists of 3 quite big parts:
```

template<>
int Storage::tagger<Foo>::get_tag()  noexcept {
  return -1717182174;
}
// tons of similar get_tag()

template<typename T>
typename Storage::loader<T>::loader_fun Storage::loader<T>::get_function(int tag) noexcept {
  switch(tag){
    case -2043716967: return Storage::load_implementation_helper<Foo, T>::load;
    // tons of similar cases
  }
  php_assert(0);
}

auto unused_loaders_list __attribute__((unused)) = {
  (void*)(Storage::loader<class_instance<C$A>>::get_function),
  // tons of similar instantiations
};
```
So, it can be compiled as only one translation unit. This PR split _tagger.cpp into _tagger.cpp, but now containing only fist part of code (with get_tag() specializations), _loader.h with second part of code (definition of big template get_function()) and banch of _loader_instantiation{n}.cpp files with third part (namely instantiations of get_function() template). That split allow to compile such code in parallel and speed up final build.
I also changed 3rd part of code with explicit template instantiation syntax.

### DIFF
--- a/compiler/code-gen/files/type-tagger.cpp
+++ b/compiler/code-gen/files/type-tagger.cpp
@@ -77,8 +77,8 @@ void TypeTagger::compile_tagger(CodeGenerator &W, const IncludesCollector &inclu
   W << CloseFile{};
 }
 
-void TypeTagger::compile_loader(CodeGenerator &W, const IncludesCollector &includes, const std::map<int, std::string> &hash_of_types) const noexcept {
-  W << OpenFile{"_loader.cpp"};
+void TypeTagger::compile_loader_header(CodeGenerator &W, const IncludesCollector &includes, const std::map<int, std::string> &hash_of_types) const noexcept {
+  W << OpenFile{loader_file_};
   W << ExternInclude{G->settings().runtime_headers.get()};
   W << includes << NL;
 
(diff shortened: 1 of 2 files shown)


compiler/code-gen/files/type-tagger.cpp
```diff
@@ -41,51 +47,95 @@ void TypeTagger::compile(CodeGenerator &W) const {
...
+    FunctionSignatureGenerator{W} << "int Storage::tagger<" << type << ">::get_tag() " << BEGIN;
     W << "return " << hash << ";" << NL;
     W << END << NL << NL;
   }
 
+  W << CloseFile{};
+}
+
+void TypeTagger::compile_loader_header(CodeGenerator &W, const IncludesCollector &includes, const std::map<int, std::string> &hash_of_types) const noexcept {
+  W << OpenFile{loader_file_};
```
unserialize, line 81: Here you use `_loader_file`, whereas above and before "_tagger.cpp" explicitly, without any constant. I'd suggest to use "_loader.h" directly here also.

BTW, what do you think of renaming `_loader.h` to `_tagger_loader.h` and renaming `_loader_instantiations{i}.cpp` to `_tagger_instantiations{i}.cpp`?

drdzyk: > Here you use _loader_file, whereas above and before "_tagger.cpp" explicitly, without any constant. I'd suggest to use "_loader.h" directly here also.


This is on purpose. "_loader.h" constant used in 2 places so to keep consistency I used variable. "_tagger.cpp" used in one place, so just KISS here.


> BTW, what do you think of renaming _loader.h to _tagger_loader.h and renaming _loader_instantiations{i}.cpp to _tagger_instantiations{i}.cpp?

`
  template<typename T>
  struct tagger {
    static int get_tag() noexcept;
  };

  template<typename T>
  struct loader {
    static loader_fun get_function(int tag) noexcept;
  };`
Struct loader and tagger has no coupling actually, so I think better to leave as is.

unserialize (approved):

PR #402, opened 2022-01-11
use default linker for partial linking

drdzyk:
Prior this patch partial linking performed with hardcoded "ld", which works quite slowly.
After patch, linking will performed indirectly through compiler, and compiler may choose linker from special directory. Clang and gcc allow to specify such directory through "-B" option. At kphp level such dir may be specified through new "--compiler-toolchain-dir" cli option.

### DIFF
--- a/compiler/compiler-settings.cpp
+++ b/compiler/compiler-settings.cpp
@@ -301,7 +301,7 @@ void CompilerSettings::init() {
 
   std::string cxx_default_flags = ss.str();
 
-  incremental_linker.value_ = cxx.get();
+  cxx_toolchain_option.value_ = !cxx_toolchain_dir.value_.empty() ? ("-B" + cxx_toolchain_dir.value_) : "";
   incremental_linker_flags.value_ = dynamic_incremental_linkage.get() ? "-shared" : "-r -nostdlib";
 
   remove_extra_spaces(extra_ld_flags.value_);
--- a/compiler/compiler-settings.cpp
+++ b/compiler/compiler-settings.cpp
@@ -301,8 +301,8 @@ void CompilerSettings::init() {
 
   std::string cxx_default_flags = ss.str();
 
-  incremental_linker.value_ = dynamic_incremental_linkage.get() ? cxx.get() : "ld";
-  incremental_linker_flags.value_ = dynamic_incremental_linkage.get() ? "-shared" : "-r";
+  incremental_linker.value_ = cxx.get();
+  incremental_linker_flags.value_ = dynamic_incremental_linkage.get() ? "-shared" : "-r -nostdlib";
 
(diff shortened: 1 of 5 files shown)


compiler/compiler-settings.cpp
```diff
@@ -301,8 +301,8 @@ void CompilerSettings::init() {
 
   std::string cxx_default_flags = ss.str();
 
-  incremental_linker.value_ = dynamic_incremental_linkage.get() ? cxx.get() : "ld";
-  incremental_linker_flags.value_ = dynamic_incremental_linkage.get() ? "-shared" : "-r";
+  compiler_toolchain_option.value_ = !compiler_toolchain_dir.value_.empty() ? ("-B" + compiler_toolchain_dir.value_) : "";
```
unserialize, line 304: If toolchain_dir is empty, then "" will be passed to compiler cmd line — and it will work as expected (will just invoke some default linker, probably ld), right?

compiler/kphp2cpp.cpp
```diff
@@ -261,6 +261,8 @@ int main(int argc, char *argv[]) {
              "php-code-version", "KPHP_PHP_CODE_VERSION", "unknown");
   parser.add("C++ compiler for building the output binary", settings->cxx,
              "cxx", "KPHP_CXX", get_default_cxx());
+  parser.add("Directory for compiler toolchain. Will be passed to compiler through -B option", settings->compiler_toolchain_dir,
```
unserialize, line 264: "for C++ compiler toolchain"? Not to be confused with kphp compiler itself.
BTW, maybe to name this option `cxx-toolchain-dir`?

compiler/compiler-settings.cpp
drdzyk: If toolchain_dir is empty, then "" will be appended to command line that invoke compiler. So yes it safe.

compiler/kphp2cpp.cpp
drdzyk: > "for C++ compiler toolchain"? Not to be confused with kphp compiler itself.
Done.
> BTW, maybe to name this option cxx-toolchain-dir?
I'm doubt in. cxx is not compiler, I guess it bad naming.

unserialize: But from the other point of view, `cxx` means "C++ compiler" (g++, or clang, or /usr/bin/c++, etc). We use `KPHP_CXX`, `KPHP_EXTRA_CXXFLAGS` and others. CMake also uses such naming (example: `-DCMAKE_CXX_COMPILER`, `-DCMAKE_CXX_COMPILER_LAUNCHER`). I think, "cxx" would be exactly what in means here

drdzyk: Ok, done.

unserialize (approved):

PR #403, opened 2022-01-12
refactored common and server stats

troy4eg:


server/server-stats.cpp
```diff
@@ -418,6 +418,36 @@ struct WorkerStatsBundle : EnumTable<E, std::array<std::atomic<typename E::StatT
...
 
+template<typename T>
+struct WorkerSamples : vk::not_copyable {
+public:
+  void recalc(size_t worker_index) {
+    percentiles.update_percentiles(samples.begin(), samples.begin() + worker_index);
+  }
+
+  Percentiles<T> percentiles;
+  std::array<T, WorkersControl::max_workers_count> samples{};
```
DrDet, line 429: What exactly will `samples` store and why is it needed? One last stat value for each worker process?

server/server-stats.cpp
```diff
@@ -418,6 +418,36 @@ struct WorkerStatsBundle : EnumTable<E, std::array<std::atomic<typename E::StatT
...
+    if (const uint16_t len = last_worker_id - first_worker_id) {
+      for (size_t i = 0; i != this->size(); ++i) {
+        const auto &stat = stats[i];
+        size_t worker_index = first_worker_id;
+        while (worker_index != last_worker_id) {
+          const auto sample = stat[worker_index].load(std::memory_order_relaxed);
+          (*this)[i].samples[worker_index] = sample;
+          ++worker_index;
+        }
+        (*this)[i].recalc(worker_index);
```
DrDet, line 445: At the old version of this, percentiles were recalced in range `[first_worker_id; last_worker_id]`, but now they recalced in range `[0; last_worker_id]`. That's incorrect when we reaclc stats for job workers, e.g. when `first_worker_id != 0` , called from here:
https://github.com/VKCOM/kphp/blob/ce36584a4b77c5ea410a8cc47c7e9c333b8637c1/server/server-stats.cpp#L624

DrDet (changes requested):

DrDet (approved):

PR #405, opened 2022-01-14
unification of vk_win_to_utf8 function from vkext and runtime

troy4eg:
* fixed vkext vk_win_to_utf8 escape parameter

### DIFF
--- a/vkext/vkext.cpp
+++ b/vkext/vkext.cpp
@@ -421,7 +421,48 @@ void write_char_utf8(int c) {
   write_buff_char(';');
 }
 
-int win_to_utf8(const char *s, int len) {
+void write_char_utf8_no_escape(int c) {
+  if (!c) {
+    return;
+  }
+  if (c < 128) {
+    write_buff_char((char)c);
+    return;
+  }
+  // 2 bytes(11): 110x xxxx 10xx xxxx
+  if (c < 0x800) {
+    write_buff_char_2((char)(0xC0 + (c >> 6)), (char)(0x80 + (c & 63)));
+    return;
+  }
+
+  // 3 bytes(16): 1110 xxxx 10xx xxxx 10xx xxxx
+  if (c < 0x10000) {
+    write_buff_char_3((char)(0xE0 + (c >> 12)), (char)(0x80 + ((c >> 6) & 63)), (char)(0x80 + (c & 63)));
+    return;
+  }
+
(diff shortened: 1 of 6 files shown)


vkext/vkext.cpp
```diff
@@ -532,6 +571,47 @@ int win_to_utf8(const char *s, int len) {
   return cur_buff_len;
 }
 
+static int json_escape(const char *string, int len) {
```
DrDet, line 574: What do you think about extracting this logic as well as `write_buff_char_*` and related buffers, somewhere to common and then use the single code in kphp and vkext?

vkext/vkext.cpp
```diff
@@ -614,11 +653,12 @@ PHP_FUNCTION (vk_utf8_to_win) {
 PHP_FUNCTION (vk_win_to_utf8) {
   char *text;
   long text_len = 0;
-  if (zend_parse_parameters(ZEND_NUM_ARGS() TSRMLS_CC, "s", &text, &text_len) == FAILURE) {
+  bool escape = false;
+  if (zend_parse_parameters(ZEND_NUM_ARGS() TSRMLS_CC, "s", &text, &text_len, &escape) == FAILURE) {
```
DrDet, line 657: Now it still accepts only one parameter and will fail with error if you pass the second one. Like this:
```
Warning: vk_win_to_utf8() expects exactly 1 parameter, 2 given in /home/denis/kphp/dev/index.php on line 3
```
You need to add some type specifiers to fix this. You can find them [here](http://php.adamharvey.name/manual/ru/internals2.funcs.php). I think it should look smth like this:
```suggestion
  if (zend_parse_parameters(ZEND_NUM_ARGS() TSRMLS_CC, "s|b", &text, &text_len, &escape) == FAILURE) {
```

DrDet (changes requested):
Also let's add some phpt tests to avoid similar bugs

vkext/vkext.stub.php
```diff
@@ -19,7 +19,7 @@ function vk_utf8_to_win(string $text, int $max_len = 0, bool $exit_on_error = fa
   return "";
 }
 
-function vk_win_to_utf8(string $text) : string {
+function vk_win_to_utf8(string $text, bool $escape) : string {
```
DrDet, line 22: as it's optional, it should be like this:
```suggestion
function vk_win_to_utf8(string $text, bool $escape = false) : string {
```

DrDet (dismissed):

DrDet (approved):

PR #415, opened 2022-01-21
fix race condition in class assumptions

unserialize:
A race condition could occur in such scenario:
```
function demo() {
$log = function() { ... };

(function() use($log) { $log(); })();
(function() use($log) { $log(); })();
(function() use($log) { $log(); })();
}
```

Multiple lambdas (processed in parallel) want to know assumption for `$log` inside `demo`.

### DIFF
--- a/compiler/class-assumptions.cpp
+++ b/compiler/class-assumptions.cpp
@@ -262,6 +262,13 @@ void assumption_add_for_var(FunctionPtr f, const std::string &var_name, const As
     return;
   }
 
+  // in case assumptions_for_vars for f accessed from different threads, use a single int-mutex guard
+  int expected = 0;
+  while (!f->assumption_vars_mu.compare_exchange_strong(expected, 1)) {
+    std::this_thread::sleep_for(std::chrono::nanoseconds{100});
+    expected = 0;
+  }
+
   for (auto &name_and_a : f->assumptions_for_vars) {
     if (name_and_a.first == var_name) {
       if (Assumption merged = assumption_merge(name_and_a.second, assumption)) {
@@ -270,11 +277,13 @@ void assumption_add_for_var(FunctionPtr f, const std::string &var_name, const As
         stage::set_location(v_location->location);
         print_error_assumptions_arent_merged_for_var(var_name, name_and_a.second, assumption);
       }
+      f->assumption_vars_mu = 0;
       return;
     }
   }
 
   f->assumptions_for_vars.emplace_back(var_name, assumption);
+  f->assumption_vars_mu = 0;
 //  printf("%s() $%s %s\n", f->name.c_str(), var_name.c_str(), assumption.as_human_readable().c_str());
 }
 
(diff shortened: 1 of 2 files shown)


unserialize:
I don't like this approach of fixing.
I've replaced `std::vector` to `std::forward_list` in #438, which makes things a bit better: instead of reallocating memory (which caused a crash), a duplicate is inserted into `assumptions_for_vars`.

The problem is not fixed yet, because emplacing an element into forward_list is not atomic. Here is a working example of a problem:
```
<?php

function demo1() {
  $log = function() { echo 1, "\n"; };
  $log();
  (function() use($log) { $log(); })();
  (function() use($log) { $log(); })();
  (function() use($log) { $log(); })();
  (function() use($log) { $log(); })();
  (function() use($log) { $log(); })();
  (function() use($log) { $log(); })();
  (function() use($log) { $log(); })();
  (function() use($log) { $log(); })();
  (function() use($log) { $log(); })();
  (function() use($log) { $log(); })();
  (function() use($log) { $log(); })();
  (function() use($log) { $log(); })();
  (function() use($log) { $log(); })();
  (function() use($log) { $log(); })();
  (function() use($log) { $log(); })();
}

function demo2() {
  $log = function() { echo 1, "\n"; };
  $log();
  (function() use($log) { $log(); })();
  (function() use($log) { $log(); })();
  (function() use($log) { $log(); })();
  (function() use($log) { $log(); })();
  (function() use($log) { $log(); })();
}

function demo3() {
  $log = function() { echo 1, "\n"; };
  $log();
  (function() use($log) { $log(); })();
  (function() use($log) { $log(); })();
  (function() use($log) { $log(); })();
  (function() use($log) { $log(); })();
  (function() use($log) { $log(); })();
}

function demo4() {
  $log = function() { echo 1, "\n"; };
  $log();
  (function() use($log) { $log(); })();
  (function() use($log) { $log(); })();
  (function() use($log) { $log(); })();
  (function() use($log) { $log(); })();
  (function() use($log) { $log(); })();
}

function demo5() {
  $log = function() { echo 1, "\n"; };
  $log();
  (function() use($log) { $log(); })();
  (function() use($log) { $log(); })();
  (function() use($log) { $log(); })();
  (function() use($log) { $log(); })();
  (function() use($log) { $log(); })();
}

demo1();
demo2();
demo3();
demo4();
demo5();
```

Running
```
KPHP_NO_MAKE=1 KPHP_THREADS_COUNT=24 ../objs/bin/kphp2cpp -M cli ../dev/dev.php
```
many times, you can face a compilation error.

The right way to fix this is not to rely on var names at all. If we build cfg before analyzing code, assigning VarPtr to every $variable, then assumptions would work on vars, not on var names. `use` of a lambda would point to the same var as held by the outer function, and `assumptions_for_vars` would probably go away or at least completely rewritten.

Probably, some time I'll investigate to moving cfg above. Until that point, this race condition exists, though is very-very rare in practice.

PR #416, opened 2022-01-24
add to_array_debug() function

drdzyk:
In this PR I add an ability to convert standalone shapes and tuples in arrays via old instance_to_array() and new to_array_debug() functions. instance_to_array() was preserved to keep backward compatibility.


unserialize:
BTW, should this function work with arrays? If I have
```
$arr = [new A];
```
Should `to_array_debug($arr)` work? (for now, it doesn't)

drdzyk:
> BTW, should this function work with arrays? If I have
> 
> ```
> $arr = [new A];
> ```
> 
> Should `to_array_debug($arr)` work? (for now, it doesn't)

I have no strong opinion about that. Maybe it shouldn't because we are about strict typing code, so user should know what he converts and conversion array to array looks excessive and a bit strange thus.

unserialize (approved):

PR #420, opened 2022-01-26
to_array_debug() now able to print shape keys

drdzyk:
Currently, to_array_debug() print just index of key for 'shape' instead of real string name. After this PR it will be real string keys.

### DIFF
--- a/compiler/code-gen/files/init-scripts.cpp
+++ b/compiler/code-gen/files/init-scripts.cpp
@@ -6,6 +6,7 @@
 
 #include "compiler/code-gen/common.h"
 #include "compiler/code-gen/declarations.h"
+#include "compiler/code-gen/files/shape-keys.h"
 #include "compiler/code-gen/includes.h"
 #include "compiler/code-gen/namespace.h"
 #include "compiler/code-gen/naming.h"
@@ -208,6 +209,9 @@ void InitScriptsCpp::compile(CodeGenerator &W) const {
   W << GlobalResetFunction(main_file_id->main_function) << NL;
 
   FunctionSignatureGenerator(W) << "void init_php_scripts() " << BEGIN;
+
+  W << ShapeKeys::get_function_declaration() << ";" << NL;
+  W << ShapeKeys::get_function_name() << "();" << NL << NL;
 
   W << FunctionName(main_file_id->main_function) << "$global_reset();" << NL;
 
--- a/compiler/phpdoc.cpp
+++ b/compiler/phpdoc.cpp
@@ -8,6 +8,7 @@
 #include <utility>
 
 #include "common/termformat/termformat.h"
+#include "common/php-functions.h"
 
 #include "compiler/compiler-core.h"
 #include "compiler/data/function-data.h"
@@ -491,7 +492,7 @@ const TypeHint *PhpDocTypeRuleParser::parse_shape_type() {
   // elements are [ op_double_arrow ( op_func_name "x", TypeHint "int" ), ... ]
   // may end with tok_varg: shape(x:int, ...)
   while (true) {
-    std::string elem_name = static_cast<std::string>(cur_tok->str_val);
+    const auto elem_name = static_cast<std::string>(cur_tok->str_val);
     cur_tok++;
 
     bool is_nullable = false;
@@ -507,9 +508,13 @@ const TypeHint *PhpDocTypeRuleParser::parse_shape_type() {
     const TypeHint *elem_type_hint = parse_type_expression();
     if (is_nullable) {
       elem_type_hint = TypeHintOptional::create(elem_type_hint, true, false);
+      const std::int64_t hash = string_hash(elem_name.data(), elem_name.size());
+      // nullable shape's key that is never used in code will not be present in shapes key storage,
+      // so we will get an error while trying to print it via to_array_debug(), thus it added here
+      TypeHintShape::register_known_key(hash, elem_name);
     }
 
-    shape_items.emplace_back(std::make_pair(elem_name, elem_type_hint));
+    shape_items.emplace_back(elem_name, elem_type_hint);
 
     if (vk::any_of_equal(cur_tok->type(), tok_gt, tok_clpar)) {
       cur_tok++;
--- a/compiler/type-hint.cpp
+++ b/compiler/type-hint.cpp
@@ -4,6 +4,8 @@
 
 #include "compiler/type-hint.h"
 
+#include <mutex>
+
 #include "common/php-functions.h"
 
 #include "compiler/data/class-data.h"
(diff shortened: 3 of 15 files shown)


compiler/code-gen/files/init-scripts.cpp
```diff
@@ -234,8 +235,11 @@ void CppMainFile::compile(CodeGenerator &W) const {
   W << OpenFile("main.cpp");
   W << ExternInclude("server/php-engine.h") << NL;
 
-  W << "int main(int argc, char *argv[]) " << BEGIN
-    << "return run_main(argc, argv, php_mode::" << G->settings().mode.get() << ")" << SemicolonAndNL{}
+  W << "int main(int argc, char *argv[]) " << BEGIN;
+  W << ShapeKeys::get_function_declaration() << ";" << NL;
+  W << ShapeKeys::get_function_name() << "();" << NL << NL;
+
+  W  << "return run_main(argc, argv, php_mode::" << G->settings().mode.get() << ")" << SemicolonAndNL{}
```
unserialize, line 242: It's very strange, that main() function is changed. I suppose, there is a more convenient way to propagate data from compilation to runtime

compiler/phpdoc.cpp
```diff
@@ -9,6 +9,7 @@
 
 #include "common/termformat/termformat.h"
 
+#include "common/php-functions.h"
```
unserialize, line 12: One live above (near another common/ include)

compiler/type-hint.cpp
```diff
@@ -337,6 +337,19 @@ const TypeHint *TypeHintShape::create(std::vector<std::pair<std::string, const T
...
+void TypeHintShape::set_shape_key(std::int64_t key_hash, std::string_view key) noexcept {
+  std::lock_guard lock{shape_keys_storage_lock_};
+  shape_keys_storage_.emplace(key_hash, key);
+}
+
+const std::map<std::int64_t, std::string> &TypeHintShape::get_shape_keys_storage() noexcept {
+  std::lock_guard lock{shape_keys_storage_lock_};
+  return shape_keys_storage_;
+}
+
```
unserialize, line 352: Please, move these lines somewhere to the top of the file. This place is a group of create() methods definitions for all type hints. The top of the file is exactly the place for custom methods that are not grouped together

compiler/type-hint.cpp
```diff
@@ -337,6 +337,19 @@ const TypeHint *TypeHintShape::create(std::vector<std::pair<std::string, const T
...
+std::map<std::int64_t, std::string> TypeHintShape::shape_keys_storage_;
+std::mutex TypeHintShape::shape_keys_storage_lock_;
+
+void TypeHintShape::set_shape_key(std::int64_t key_hash, std::string_view key) noexcept {
+  std::lock_guard lock{shape_keys_storage_lock_};
+  shape_keys_storage_.emplace(key_hash, key);
+}
+
+const std::map<std::int64_t, std::string> &TypeHintShape::get_shape_keys_storage() noexcept {
+  std::lock_guard lock{shape_keys_storage_lock_};
```
unserialize, line 349: We don't need any lock here:
1) accessing the filled storage is done much later that filling it
2) even if (1) is not true, this lock wouldn't actually lock reading access to it from the calling side

compiler/type-hint.cpp
```diff
@@ -337,6 +337,19 @@ const TypeHint *TypeHintShape::create(std::vector<std::pair<std::string, const T
   );
 }
 
+std::map<std::int64_t, std::string> TypeHintShape::shape_keys_storage_;
+std::mutex TypeHintShape::shape_keys_storage_lock_;
+
+void TypeHintShape::set_shape_key(std::int64_t key_hash, std::string_view key) noexcept {
+  std::lock_guard lock{shape_keys_storage_lock_};
```
unserialize, line 344: Can we move std::mutex here, as a static variable? 
`#include<mutex>` seems quite weird in type-hint.h

compiler/type-hint.h
```diff
@@ -454,12 +457,18 @@ class TypeHintShape : public TypeHint {
...
+  static std::mutex shape_keys_storage_lock_;
+
 public:
   std::vector<std::pair<std::string, const TypeHint *>> items;
-  bool is_vararg;
+  bool is_vararg{false};
 
   static const TypeHint *create(std::vector<std::pair<std::string, const TypeHint *>> &&items, bool is_vararg);
 
+  static void set_shape_key(std::int64_t key_hash, std::string_view key) noexcept;
```
unserialize, line 469: 1) There is some better naming for `set_shape_key`, since it doesn't set any shape key, it adds a known key to storage. Maybe, something like `register_known_key`? And not `get_shape_keys_storage`, but `get_all_registered_keys`?
2) We store std::string in a map, let's take std::string here also, not std::string_view

tests/phpt/dl/1036_to_array_debug_shape_nullable.php
```diff
@@ -0,0 +1,16 @@
...
+function to_array_debug_shape_nullable() {
+  /** @var shape(foo:string, bar?:int) */
+  $shape = shape(["foo" => "baz"]);
+  $dump = to_array_debug($shape);
+
+  #ifndef KPHP
+  $dump = ['foo' => 'baz', 'bar' => null];
+  #endif
+  var_dump($dump);
+}
```
unserialize, line 14: Please, add more tests:
1) an instance having two shape fields 
2) an instance as a value of one of the keys
3) a shape within a shape

compiler/code-gen/files/init-scripts.cpp
drdzyk: Yea, true. I did so because of otherwise get weird error related with redefining malloc. Problem fixed with help of Denis.

compiler/phpdoc.cpp
drdzyk: Done

compiler/type-hint.cpp
drdzyk: Done

compiler/type-hint.cpp
drdzyk: Done

compiler/type-hint.cpp
drdzyk: Done

compiler/type-hint.h
drdzyk: 1) Done
2) disagree, string_view intentionally used here to make code simple and clean

tests/phpt/dl/1036_to_array_debug_shape_nullable.php
drdzyk: We discussed that such cases more or less covered in other tests

unserialize (approved):

PR #433, opened 2022-02-12
Add str_contains() function

Danil42Russia:
## Information
- PHP RFC name: str_contains
- PHP RFC link: https://wiki.php.net/rfc/str_contains
- Issue: #290
- Added a function from a low priority block :)
- Tests written

## Example

```php
str_contains("test string", "test"); // true
str_contains("test string", "hello"); // false
```

### DIFF
--- a/runtime/string.inl
+++ b/runtime/string.inl
@@ -345,12 +345,14 @@ bool string::ends_with(const string &other) const noexcept {
 }
 
 bool string::contains(const string &other) const noexcept {
-  if (other.size() > size())
+  if (other.size() > size()) {
     return false;
+  }
 
   for (size_type i = 0; i < (size() - other.size() + 1); i++) {
-    if (memcmp(c_str() + i, other.c_str(), other.size()) == 0)
+    if (memcmp(c_str() + i, other.c_str(), other.size()) == 0) {
       return true;
+    }
   }
 
   return false;


Danil42Russia:
Error text:
```
Starting php to cpp transpiling...

Compilation error at stage: Parse file, gen by compiler-core.cpp:277
  lib/polyfills/file2.php:10
    }

Redeclaration of function str_contains(), the previous declaration was in .../builtin-functions/_functions.txt

```

Danil42Russia:
Tests don't break anymore!

runtime/string.inl
```diff
@@ -344,6 +344,18 @@ bool string::ends_with(const string &other) const noexcept {
   return other.size() > size() ? false : !memcmp(c_str() + (size() - other.size()), other.c_str(), other.size());
 }
 
+bool string::contains(const string &other) const noexcept {
+  if (other.size() > size())
+    return false;
```
quasilyte, line 349: It's a good practice to always use `{}` for statements like `if` and `for`.
Please use it here and in other places.

Danil42Russia: Fixed :)

runtime/string_functions.cpp
```diff
@@ -2327,6 +2327,10 @@ bool f$str_ends_with(const string &haystack, const string &needle) {
   return haystack.ends_with(needle);
 }
 
+bool f$str_contains(const string &haystack, const string &needle) {
+  return haystack.contains(needle);
```
quasilyte, line 2331: What are the advantages of implementing it as a new functions instead of using already existing strpos?

PHP uses similar approach:
https://github.com/php/php-src/blob/0b6f58d907e7169f7876bbea199967262197fbaa/ext/standard/string.c#L1794

As seen in the code below, str_contains could be implemented in terms of strpos like so:
```php
return '' === $needle || false !== strpos($haystack, $needle);
```

I would check what `php_memnstr` from PHP is. Maybe we have compatible function. If not, we could add one and use it in places where PHP uses it.

Maybe `php_memnstr` is identical to the newly introduced `contains` method, but I suggest re-checking that.

PR #438, opened 2022-02-15
speedup phpdoc parsing and store PhpDocComment instead of phpdoc_str

unserialize:
There are three main purposes of this PR.

## `PhpDocComment *` instead of `vk::string_view`

For functions, classes, fields, vars, etc. we now store `const PhpDocComment *phpdoc` instead of `vk::string_view phpdoc_str`.

Storing a meaningful object instead of a string is much more reliable, since we use methods like `phpdoc->has_tag(type)` instead of `phpdoc_tag_exists(phpdoc_str, type)`. 

Phpdoc comment is now parsed in gentree — it's split in tags, these tags are kept in `forward_list`. But the values of tags are analyzed later in the pipeline on demand.

## Speed up phpdoc parsing and accessing

When we stored `vk::string_view`, every time we had to access it like `phpdoc_tag_exists` / `phpdoc_parse_var_and_type_name` / etc. — at first we parsed `phpdoc_str`, when we did an actual query. Some doc comments were parsed more than once. Now they all are parsed exactly once in gentree, and no double parsing is done, never.

The algorithm that parses phpdoc comment to tags was completely rewritten, now it's much faster (but a bit more complicated, especially by introducing hashes).

All in all, we have **about 1 second boost for vkcom** and use slightly less memory.

## `@var` parsing is compatible with `@kphp-template`

Prior to this PR, `@var` inside a function (not above a class field) was parsed right in gentree. That's completely wrong for future generics, since
```
/**
 * @kphp-template T
 * @kphp-param T $obj
 */
function f($obj) {
  /** @var T[] $arr */
  $arr = [$obj];
}
```
We can parse `@var T[]` correctly only when we know that `f` is actually `f<T>`, which means that `T` is a generics type, not a `class T`.

This PR does `@var` parsing in a right place to make it compatible with my future branches.

### DIFF
--- a/compiler/class-assumptions.cpp
+++ b/compiler/class-assumptions.cpp
@@ -517,7 +517,7 @@ static Assumption calc_assumption_for_var(FunctionPtr f, const std::string &var_
 
   if (f->type == FunctionData::func_main || f->type == FunctionData::func_switch) {
     if (var_name == "MC" || var_name == "PMC") {
-      assumption_add_for_var(f, var_name, Assumption(G->get_memcache_class()), f->root);
+      return Assumption(G->get_memcache_class());
     }
   }
 
--- a/compiler/gentree.cpp
+++ b/compiler/gentree.cpp
@@ -400,7 +400,7 @@ VertexPtr GenTree::get_op_num_const() {
   return VertexPtr{};
 }
 
-VertexPtr GenTree::get_expr_top(bool was_arrow, vk::string_view phpdoc_str) {
+VertexPtr GenTree::get_expr_top(bool was_arrow, const PhpDocComment *phpdoc) {
   auto op = cur;
 
   VertexPtr res, first_node;
@@ -611,16 +611,16 @@ VertexPtr GenTree::get_expr_top(bool was_arrow, vk::string_view phpdoc_str) {
     }
     case tok_static:
       next_cur();
-      res = get_lambda_function(phpdoc_str, FunctionModifiers::static_lambda());
+      res = get_lambda_function(phpdoc, FunctionModifiers::static_lambda());
       break;
     case tok_function:
     case tok_fn:
-      res = get_lambda_function(phpdoc_str, FunctionModifiers::nonmember());
+      res = get_lambda_function(phpdoc, FunctionModifiers::nonmember());
       break;
     case tok_phpdoc: {      // /** ... */ before expression (not before a statement)
       vk::string_view phpdoc_str = cur->str_val;
       next_cur();
-      return get_expr_top(was_arrow, phpdoc_str);
+      return get_expr_top(was_arrow, new PhpDocComment(phpdoc_str));
     }
     case tok_isset: {
       auto temp = get_multi_call<op_isset, op_none>(&GenTree::get_expression, true);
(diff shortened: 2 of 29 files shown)


compiler/gentree.cpp
```diff
@@ -2051,16 +2056,16 @@ VertexPtr GenTree::get_statement(vk::string_view phpdoc_str) {
           || (next == tok_static && cur_function->type == FunctionData::func_class_holder)) {
         vk::string_view phpdoc_str = cur->str_val;
         next_cur();
-        return get_statement(phpdoc_str);
+        return get_statement(new PhpDocComment(phpdoc_str));
```
drdzyk, line 2059: Why not to use reliable semantic with std::unique_ptr or std::shared_ptr? (It's about a rule - don't use new/delete in new c++ code)

compiler/phpdoc.h
```diff
@@ -4,81 +4,116 @@
...
+};
+
+// one phpdoc contains of several doc tags
+// every doc tag has a type (enum mapping @tag-name) and a value
+struct PhpDocTag {
+  // if a tag value is like "int|false $var_name" (for @param / @var / etc.),
+  // it's represented as this structure
+  struct TypeAndVarName {
+    const TypeHint *type_hint{nullptr};
+    vk::string_view var_name;     // stored without leading "$"; could be empty if omitted in phpdoc
```
drdzyk, line 58: std::string_view here?

compiler/phpdoc.h
```diff
@@ -4,81 +4,116 @@
...
 
-public:
-  static doc_type get_doc_type(const std::string &str) {
-    auto it = str2doc_type.find(str);
-    bool found = (it != str2doc_type.end());
 
-    return found ? it->second : unknown;
-  }
+  PhpDocType type;
+  vk::string_view value;
```
drdzyk, line 65: and here

compiler/phpdoc.h
```diff
@@ -4,81 +4,116 @@
...
+// one phpdoc contains of several doc tags
+// every doc tag has a type (enum mapping @tag-name) and a value
+struct PhpDocTag {
+  // if a tag value is like "int|false $var_name" (for @param / @var / etc.),
+  // it's represented as this structure
+  struct TypeAndVarName {
+    const TypeHint *type_hint{nullptr};
+    vk::string_view var_name;     // stored without leading "$"; could be empty if omitted in phpdoc
+
+    operator bool() const { return static_cast<bool>(type_hint); }
```
drdzyk, line 60: noexcept specifiers in all new code looks reasonable

compiler/phpdoc.cpp
```diff
@@ -21,115 +21,195 @@
...
+    for (const KnownPhpDocTag &tag: ALL_TAGS) {
+      if (tag.hash == hash) {
+        return tag.type;
       }
-      if (c == '*') {
-        have_star = true;
-        continue;
+    }
+    return PhpDocType::unknown;
+  }
```
drdzyk, line 54: why not just std::unordered_map?
If your concern will be about significant performance difference, can we at least avoid code duplication?
What the magic constant? Can we use std::hash for this?

compiler/phpdoc.cpp
```diff
@@ -21,115 +21,195 @@
...
+  static const KnownPhpDocTag ALL_TAGS[N_TAGS];
 
-  vector<string> lines(1);
-  bool have_star = false;
-  for (char c : phpdoc) {
-    if (!have_star) {
-      if (c == ' ' || c == '\t') {
-        continue;
+public:
+  static PhpDocType name2type(const char *start, const char *end) __attribute__((always_inline)) {
```
drdzyk, line 42: Please use c++17 attributes in new code. Here it will be [[gnu::always_inline]] prior function definition.

compiler/class-assumptions.cpp
```diff
@@ -492,7 +492,7 @@ static Assumption calc_assumption_for_var(FunctionPtr f, const std::string &var_
 
   if (f->type == FunctionData::func_main || f->type == FunctionData::func_switch) {
     if (var_name == "MC" || var_name == "PMC") {
-      assumption_add_for_var(f, var_name, Assumption(G->get_memcache_class()), f->root);
+      return Assumption(G->get_memcache_class());
```
drdzyk, line 495: Don't understand this change. Could you explain please?

drdzyk (changes requested):

compiler/gentree.cpp
unserialize: It's not unique_ptr, because a single phpdoc could be kept in multiple places.
More semantically, it's shared, but like in any other place in our code, I don't see any reason for complification. We don't need to handle refcounts, we don't need to call destructors, all our objects live till compilation end and never freed. Keeping `const T *` in `FunctionData`, `ClassData`, `vertex_inner`, etc. is all over the code.

compiler/phpdoc.h
unserialize: Maybe. Could you please advise, how to replace `value.starts_with("something")` if value is std::string_view?

compiler/phpdoc.cpp
unserialize: At first, I've used unordered_map, but finally tested that linear search is better here, because @param/@var/@return are the move frequest tags in real code (90%), and they are placed at the top of `ALL_TAGS`.
A "magic number" is from http://www.cse.yorku.ca/~oz/hash.html. It's very simple but good enough for short strings (like names). 
About code duplication — personally I don't see any problem in duplicating 3 lines kept together and never changed (even if occasionally changed, lots of tests would fail). The first should be constexpr, the latter should not. How would you advice to remove duplication?

compiler/class-assumptions.cpp
unserialize: This is not a meaningful change, just a minor style fix I've accidently notices (to make this line be in style of above and below, that also use "return")

compiler/phpdoc.h
unserialize: Done, added in several places

compiler/phpdoc.cpp
unserialize: Done. Didn't know about that!

compiler/phpdoc.cpp
drdzyk: Ok, got it about performance. 
About hash, i've found `string_hash_5381()` in our code with same logic. If you make it constexpr, noexcept, accepted std::string_view then it will be possible to use it in both places. E.g. constructor will looks like:
```
  constexpr KnownPhpDocTag(const char *tag_name, PhpDocType type) noexcept
    : hash(string_hash_5381(tag_name))
    , type(type)
    , tag_name(tag_name) {}
```

compiler/class-assumptions.cpp
drdzyk: ok, got it

compiler/phpdoc.h
drdzyk: Ah, in this case nothing. It will be possible after debian buster, or if we upgrade compiler there.

DrDet (dismissed):

drdzyk (approved):

PR #444, opened 2022-02-17
implement array_unset() api function

drdzyk:
array_unset($arr, $key) function just return value corresponding to $key, and remove it from $arr.

### DIFF
--- a/builtin-functions/_functions.txt
+++ b/builtin-functions/_functions.txt
@@ -267,6 +267,7 @@ function array_slice ($a ::: array, $offset ::: int, $length = null, $preserve_k
 /** @kphp-extern-func-info cpp_template_call */
 function array_pad ($a ::: array, $size ::: int, $value ::: any) ::: array< ^1[*] | ^3 >;
 function array_column ($a ::: array, $column_key, $index_key = null) ::: array< ^1[*][*] > | false;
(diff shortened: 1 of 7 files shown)


builtin-functions/_functions.txt
```diff
@@ -267,6 +267,7 @@ function array_slice ($a ::: array, $offset ::: int, $length = null, $preserve_k
 /** @kphp-extern-func-info cpp_template_call */
 function array_pad ($a ::: array, $size ::: int, $value ::: any) ::: array< ^1[*] | ^3 >;
 function array_column ($a ::: array, $column_key, $index_key = null) ::: array< ^1[*][*] > | false;
+function array_unset (&$a ::: array, any $key) ::: ^1[*];
```
DrDet, line 270: I think `mixed` type for `$key` fits more.  
```suggestion
function array_unset (&$a ::: array, mixed $key) ::: ^1[*];
```
Now you can pass class instance to `$key`, that will lead to gcc error

DrDet (approved):

builtin-functions/_functions.txt
drdzyk: Discussed offline, I dislike to confuse user that key is mixed. Actually key maybe have another type like int, string. So better leave as is

PR #449, opened 2022-02-18
clang-tidy: modernize return braced init list

drdzyk:
In this patch I remove duplicating of function return type with help of clang-tidy static analyzer. This check in clang-tidy called 'modernize-return-braced-init-list'

### DIFF
--- a/runtime/ffi.h
+++ b/runtime/ffi.h
@@ -175,7 +175,7 @@ string f$FFI$$string(const T &v) {
 
 template<class T>
 string f$FFI$$string(const T &v, int64_t size) {
-  return string(reinterpret_cast<const char*>(ffi_c_value_ptr(v)), size);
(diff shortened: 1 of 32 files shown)


troy4eg (dismissed):

runtime/ffi.h
```diff
@@ -279,7 +279,7 @@ inline int64_t ffi_c2php(int64_t v) { return v; }
 inline double ffi_c2php(float v) { return v; }
 inline double ffi_c2php(double v) { return v; }
 inline string ffi_c2php(const char *v) { return string(v); }
```
unserialize, line 281: This is unchanged, but below is changed. It is ok?

drdzyk: Checked again, all right.

drdzyk: Discussed offline. Such change can't be done because `string(const char *)` constructor marked `explicit`

PR #463, opened 2022-02-28
confdata stats refactoring

troy4eg:
Prefixes of some metrics are now tags for those metrics for statshouse

### DIFF
--- a/server/confdata-stats.cpp
+++ b/server/confdata-stats.cpp
@@ -13,10 +13,10 @@ double to_seconds(std::chrono::nanoseconds t) noexcept {
 }
 
 void write_event_stats(stats_t *stats, const char *name, const ConfdataStats::EventCounters::Event &event) noexcept {
-  stats->add_gauge_stat(event.total, name, ".total");
-  stats->add_gauge_stat(event.blacklisted, name, ".blacklisted");
-  stats->add_gauge_stat(event.ignored, name, ".ignored");
-  stats->add_gauge_stat(event.ttl_updated, name, ".ttl_updated");
+  stats->add_gauge_stat_with_type_tag(name, "total", event.total);
+  stats->add_gauge_stat_with_type_tag(name, "blacklisted", event.blacklisted);
+  stats->add_gauge_stat_with_type_tag(name, "ignored", event.ignored);
+  stats->add_gauge_stat_with_type_tag(name, "ttl_updated", event.ttl_updated);
 }
 
 } // namespace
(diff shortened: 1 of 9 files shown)


server/confdata-stats.cpp
```diff
@@ -83,16 +83,18 @@ void ConfdataStats::write_stats_to(stats_t *stats, const memory_resource::Memory
   stats->add_gauge_stat("confdata.updates.ignored", ignored_updates);
   stats->add_gauge_stat("confdata.updates.total", total_updates);
 
-  stats->add_gauge_stat("confdata.elements.total", total_elements);
-  stats->add_gauge_stat("confdata.elements.simple_key", simple_key_elements);
-  stats->add_gauge_stat("confdata.elements.one_dot_wildcard", one_dot_wildcard_elements);
-  stats->add_gauge_stat("confdata.elements.two_dots_wildcard", two_dots_wildcard_elements);
-  stats->add_gauge_stat("confdata.elements.predefined_wildcard", predefined_wildcard_elements);
-  stats->add_gauge_stat("confdata.elements.with_delay", elements_with_delay);
+  if (stats->need_aggr_stats()) {
```
DrDet, line 86: Let's rename this method to `need_aggregated_stats()`, please

server/statshouse/statshouse-client.cpp
```diff
@@ -47,6 +47,18 @@ class statshouse_stats_t : public stats_t {
     add_stat(type, key, static_cast<double>(value));
   }
 
+  void add_stat_with_tag_type(char type [[maybe_unused]], const char *key, const char *type_tag, double value) noexcept final {
+    std::vector<std::pair<std::string, std::string>> metric_tags = {{"type", std::string(type_tag)}, {"host", std::string(kdb_gethostname())}};
+    auto metric = make_statshouse_value_metric(normalize_key(key, "_%s", stats_prefix), value, {});
```
DrDet, line 52: It seems, you forgot to pass tags :)
```suggestion
    auto metric = make_statshouse_value_metric(normalize_key(key, "_%s", stats_prefix), value, metric_tags);
```

server/statshouse/worker-stats-buffer.cpp
```diff
@@ -111,6 +111,10 @@ void WorkerStatsBuffer::flush() {
   make_metric(metrics, "kphp_memory_job_common_request_usage", QueryStatKey::job_common_request_memory_usage, tags);
   make_metric(metrics, "kphp_memory_job_common_request_real_usage", QueryStatKey::job_common_request_real_memory_usage, tags);
 
+  if (metrics.empty()) {
+    return;
+  }
```
DrDet, line 116: `last_send_time` is not updated here => if metrics remain empty, flush will be called on every subsequent `cron` call. That can be significantly more often than once per second.

DrDet (changes requested):

DrDet (approved):

PR #466, opened 2022-03-02
add option for binding server to the specified IPv4 address

DrDet:


common/kprintf.cpp
```diff
@@ -305,14 +305,8 @@ int set_verbosity_by_type(const char *name, int value) {
...
 const char *ip_to_print(unsigned ip) {
-  static char abuf[1000], *ptr = abuf;
-  char *res;
-  if (ptr > abuf + 900) {
-    ptr = abuf;
-  }
-  res = ptr;
-  ptr += sprintf (ptr, "%d.%d.%d.%d", ip >> 24, (ip >> 16) & 0xff, (ip >> 8) & 0xff, ip & 0xff) + 1;
-  return res;
+  static char buf[1000];
```
quasilyte, line 308: I wonder if we can use something like INET6_ADDRSTRLEN (~46) instead of a 1000-chars wide buffer for the result.
Also, now this function doesn't have a ring-like scratch buffer, which means calling it the second time will invalidate the data returned by the first call. Is it OK in our context? I haven't checked the call sites, so maybe it's never a problem.

DrDet: I think we can use smaller buffer, I just left it as it was)

It's used as rvalue always for passing to `printfl`-like functions, so yes - it's never a problem

troy4eg (approved):

PR #468, opened 2022-03-05
[runtime] add JSON encoding error object path

quasilyte:
This change allows json_encode functions to report
errors in a more helpful way.

	json_encode(['a' => ['b' => INF]])
	old message: strange double inf in function json_encode
	new message: /['a']['b']: strange double inf in function json_encode

This is helpful if json_encode is called with various data
from multiple places. With precise object path hint, users
can fix their code easier.

The performance impact is small (~1.35% for the worst case):

	name                   old time/op  new time/op  delta
	Json::EncodeSafeArray  1.52µs ± 1%  1.54µs ± 1%  +1.35%

This is due to the fact that vector/map loop is now split into
two parts, reducing the redundant branching. Without this optimization,
code becomes ~8% slower. It comes with a little bit of code duplication.

The object path stack is a fixed array of 8 `const char*`.
This array is allocated on a stack (part of the JsonEncoder object).
If object path is longer than 8 parts, it will be truncated like this:

	/['k1']['k2']['k3']['k4']['k5']['k6']['k7']['k8']...

All integer keys are stored as `.`:

	[9 => ['a' => INF]]  = /[.]['a']
	[10 => ['a' => INF]] = /[.]['a']

It's possible to increase the stack depth limit in the future.
8 -> 16 step is almost free.

### DIFF
--- a/runtime/json-functions.cpp
+++ b/runtime/json-functions.cpp
@@ -38,7 +38,7 @@ bool json_append_char(unsigned int c) noexcept {
 }
 
 
-bool do_json_encode_string_php(const char *s, int len, int64_t options) noexcept {
+bool do_json_encode_string_php(const impl_::JsonPath &json_path, const char *s, int len, int64_t options) noexcept {
   int begin_pos = static_SB.size();
   if (options & JSON_UNESCAPED_UNICODE) {
     static_SB.reserve(2 * len + 2);
@@ -47,8 +47,8 @@ bool do_json_encode_string_php(const char *s, int len, int64_t options) noexcept
   }
   static_SB.append_char('"');
 
-  auto fire_error = [begin_pos](int pos) {
-    php_warning("Not a valid utf-8 character at pos %d in function json_encode", pos);
+  auto fire_error = [json_path, begin_pos](int pos) {
+    php_warning("%s: Not a valid utf-8 character at pos %d in function json_encode", json_path.to_string().c_str(), pos);
     static_SB.set_pos(begin_pos);
     static_SB.append("null", 4);
     return false;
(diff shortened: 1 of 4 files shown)


runtime/json-functions.cpp
```diff
@@ -215,12 +215,41 @@ bool do_json_encode_string_vkext(const char *s, int len) noexcept {
...
+    const char *key = arr[i];
+    if (key == nullptr) {
+      // int key indexing
+      result.append("[.]");
+    } else {
+      // string key indexing
+      result.append("['");
+      result.append(arr[i]);
+      result.append("']");
+    }
```
unserialize, line 239: Why did you choose the format
```
/['a'][.]['b']
```
instead of 
```
/a/#/b
```
?

runtime/json-functions.cpp
```diff
@@ -215,12 +215,41 @@ bool do_json_encode_string_vkext(const char *s, int len) noexcept {
...
+  return result;
+}
+
 JsonEncoder::JsonEncoder(int64_t options, bool simple_encode) noexcept:
   options_(options),
   simple_encode_(simple_encode) {
 }
 
-bool JsonEncoder::encode(bool b) const noexcept {
+bool JsonEncoder::encode(bool b) noexcept {
```
unserialize, line 252: why did you remove `const` here and in some places below?

runtime/json-functions.cpp
quasilyte: Because it mimics the way you would get that element. The chain of array index operations.
/-based paths are OK too, it they will be self-explanatory for people.
I opted for the format that is probably hard to be misinterpreted.

runtime/json-functions.cpp
quasilyte: Because now encoder methods modify the internal state - json path object.
It's more efficient than passing a copy to every next frame.

unserialize: Yes, but I don't see, where it modifies `json_path_` in this method or in `encode(int64_t i)` for example

quasilyte: I would say that it looks like an oversight. I'll return const qualifiers to the places where that's possible.

I don't remember if there was a real reason to remove them in these places.

quasilyte: I remembered now.
Having different constness there makes implicit T->Optional conversion a viable candidate for cases like `encode(true)` which we don't want.
Since we're dispatching the encoding like `encode(value)`, it's cumbersome to rename functions to `encode_int` and so on to disambiguate between overloadings. Removing const qualifiers from encode<int> and encode<bool> looked like a lesser evil.

unserialize (approved):

PR #472, opened 2022-03-15
optimization of sending statistics to statshouse

troy4eg:


server/statshouse/statshouse-client.cpp
```diff
@@ -123,8 +137,7 @@ bool StatsHouseClient::init_connection() {
 }
 
 void StatsHouseClient::master_send_metrics() {
-  auto [result, len] = prepare_statshouse_stats(statshouse_stats_t{tags});
-  send_metrics(result, len);
+  prepare_statshouse_stats(statshouse_stats_t{tags});
```
DrDet, line 140: let's inline this function? It's not quite `prepare` anymore

troy4eg: done

server/statshouse/statshouse-client.cpp
```diff
@@ -123,8 +131,14 @@ bool StatsHouseClient::init_connection() {
...
 
 void StatsHouseClient::master_send_metrics() {
-  auto [result, len] = prepare_statshouse_stats(statshouse_stats_t{tags});
-  send_metrics(result, len);
+  statshouse_stats_t stats{tags};
+  stats.stats_prefix = "kphp";
+  char *buf = get_engine_default_prepare_stats_buffer();
+
+  sb_init(&stats.sb, buf, STATS_BUFFER_LEN);
+  stats.sb.pos = STATSHOUSE_HEADER_OFFSET;
```
DrDet, line 139: Now `flush` can be called at the middle of adding stat. Did you test this by setting up small `STATSHOUSE_UDP_BUFFER_THRESHOLD` or something similar?
I think something can go wrong in some rare cases.

For example this logic is actually initialization. The same logic must be called at the end of each `flush`.
But now, for example, `sb_init` is not called there, that can lead to inconsistency of some internal fields (`sb.overflowed`, etc.)

DrDet (changes requested):
See comments:
https://github.com/VKCOM/kphp/pull/472/files#r848446686
https://github.com/VKCOM/kphp/pull/472/files#r848448947

DrDet (approved):

PR #483, opened 2022-03-29
add FFI arrays support

quasilyte:
* Added support for dynamically-sized C arrays
* Extended fixed-size C arrays support
* Enabled C pointers dereferencing (read) and write
* Made memcpy, memset, memcmp and other FFI functions work with arrays

Both arrays and pointers have the same FFI read/write API
in PHP and KPHP. Ideally, they would look like this:

	$arr[$i] = $value;
	var_dump($arr[$i]);

Instead, FFI array operations are implemented via functions
right now to simplify the implementation:

	ffi_array_set($arr, $i, $value);
	var_dump(ffi_array_get($arr, $i));

Some examples of how C arrays can be created:

	// A fixed size array of 5 ints
	$arr = FFI::new('int[5]');
	// Creating dynamic-sized array involve some
	// concatenations or string interpolation;
	$size = 10;
	$dynamic_arr1 = FFI::new("int[$size]");
	$dynamic_arr2 = FFI::new('int[' . ($size*2) ']');

kphp-polyfills include these functions starting with `v1.0.17`.


runtime/ffi.h
```diff
@@ -231,11 +298,29 @@ CDataPtr<typename std::remove_pointer<T>::type> ffi_new_cdata_ptr() {
...
+}
+
 template<class ToType, class FromType>
 CDataRef<ToType> ffi_cast_scalar(FromType &v) {
   return CDataRef<ToType>{(ToType*)(&v->c_value)};
 }
 
+template<class ToType, class T>
+CDataPtr<typename std::remove_pointer<ToType>::type> ffi_cast_array(class_instance<CDataArray<T>> &a) {
+  return CDataPtr<typename std::remove_pointer<ToType>::type>{(ToType)a->data};
```
DrDet, line 321:
```suggestion
  return CDataPtr<std::remove_pointer_t<ToType>>{(ToType)a->data};
```

runtime/ffi.h
```diff
@@ -231,11 +298,29 @@ CDataPtr<typename std::remove_pointer<T>::type> ffi_new_cdata_ptr() {
...
+  return cdata;
+}
+
 template<class ToType, class FromType>
 CDataRef<ToType> ffi_cast_scalar(FromType &v) {
   return CDataRef<ToType>{(ToType*)(&v->c_value)};
 }
 
+template<class ToType, class T>
+CDataPtr<typename std::remove_pointer<ToType>::type> ffi_cast_array(class_instance<CDataArray<T>> &a) {
```
DrDet, line 320:
```suggestion
auto ffi_cast_array(class_instance<CDataArray<T>> &a) {
```

runtime/ffi.h
```diff
@@ -199,6 +252,14 @@ int64_t f$FFI$$sizeof(const CDataRef<T> &ref) {
   return sizeof(*ref.c_value);
 }
 
+template<class T1, class T2>
+int64_t f$FFI$$memcmp(T1 ptr1, T2 ptr2, int64_t size)  {
+  int64_t result = std::memcmp(ffi_c_value_ptr(ptr1), ffi_c_value_ptr(ptr2), size);
+  // contrary to what PHP documentation says, it actually only returns -1, 0 and 1 values,
+  // so we need to adjust memcmp results here
+  return std::clamp(result, -1L, 1L);
```
DrDet, line 260: Let's add `#include <algorithm>` at the file beginning to fix mac os build

quasilyte: Huh. I wonder if it would be better to just do clamp operation inplace.
Is it idiomatic in our code base by the way to use clamp? I don't see it that often.

compiler/ffi/ffi_parser.cpp
```diff
@@ -124,5 +135,15 @@ std::pair<const FFIType*, FFIParseError> ffi_parse_type(const std::string &type_
...
   }
+
+  // wrap the result_type in array types
+//  while (array_dims > 0) {
+//    auto *array_type = new FFIType{FFITypeKind::Array};
+//    array_type->members = {result_type};
+//    array_type->num = -1;
+//    result_type = array_type;
+//    array_dims--;
+//  }
```
unserialize, line 146: Here (and above) lines are commented out, is this intentional?

runtime/ffi.h
unserialize: +1 for in-place

compiler/ffi/ffi_parser.cpp
quasilyte: Good catch.
It's a leftover from the time before I fixed the C parser grammar file.
Now it can be parsed without this kludge.

I removed the comment.

runtime/ffi.h
quasilyte: Done

unserialize (approved):

DrDet (dismissed):

DrDet (approved):

PR #491, opened 2022-04-12
implement CompileTimeLocation (a parameter auto-filled on call side)

unserialize:
`CompileTimeLocation` introduces an ability to access a location of a caller inside a called function.
In other words, from `function f()` body, now it's available have an exact place, where `f()` was called from.

For example, we want to implement `log_info($message)`, to be used like this
```
log_info("start calculation");
$arr = [1,2,3,4,5];
foreach ($arr as $v) {
  if ($v > 3) 
    log_info("big v=$v");
}
log_info("end calculation");
```

which would print a text log like
```
start calculation (in demo5 at dev.php:55)
big v=4 (in demo5 at dev.php:59)
big v=5 (in demo5 at dev.php:59)
end calculation (in demo5 at dev.php:61)
```

**Before this MR, it was impossible to achieve this.**
Now, `log_info()` can be implemented and looks quite trivial:
```
function log_info(string $message, CompileTimeLocation $loc = null) {
  $loc = CompileTimeLocation::calculate($loc);
  echo "$message (in {$loc->function} at {$loc->file}:{$loc->line})\n";
}
```

1) In PHP, it's polyfilled by calling `debug_backtrace` and creating `new CompileTimeLocation` assigning its fields from a trace (https://github.com/VKCOM/kphp-polyfills/pull/44)

2) In KPHP, it works at compile-time: KPHP inserts an implicit call parameter for every call:
```
log_info($s);
// is implicitly replaced to
log_info($s, new CompileTimeLocation(__RELATIVE_FILE__, __METHOD__, __LINE__));
```

It's very handy for logging/debugging purposes, cause now we are able to modify a function and get callers' locations — without modifying how a function is actually called.

By the way, one can easily wrap a logger function into another one, proxying a caller location:
```
function my_log(CompileTimeLocation $loc = null) {
  $loc = CompileTimeLocation::calculate($loc);
  ...
  generic_log($loc);
}
```
Here `my_log` proxies a location, so `general_log` prints out a location where a wrapper `my_log` is being called every time.

### DIFF
--- a/compiler/pipes/check-func-calls-and-vararg.cpp
+++ b/compiler/pipes/check-func-calls-and-vararg.cpp
@@ -4,6 +4,7 @@
 
 #include "compiler/pipes/check-func-calls-and-vararg.h"
 
+#include "compiler/data/src-file.h"
 #include "compiler/gentree.h"
 #include "compiler/name-gen.h"
 #include "compiler/type-hint.h"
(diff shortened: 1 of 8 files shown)


compiler/pipes/check-func-calls-and-vararg.cpp
```diff
@@ -87,6 +88,72 @@ VertexAdaptor<op_func_call> CheckFuncCallsAndVarargPass::process_varargs(VertexA
...
+  if (current_function->is_lambda()) {  // like tok_method_c in gentree (PHP polyfill also emits the same format)
+    v_function->str_val = "{closure}";
+  } else if (!current_function->is_main_function()) {
+    v_function->str_val = !current_function->modifiers.is_nonmember()
+                          ? current_function->class_id->name + "::" + current_function->local_name()
+                          : current_function->name;
+  }
+
+  auto v_line = VertexAdaptor<op_int_const>::create().set_location(call_location);
+  v_line->str_val = std::to_string(call_location.get_line());
```
quasilyte, line 127: How about `GenTree::create_int_const(call_location.get_line())`?

compiler/pipes/check-func-calls-and-vararg.cpp
```diff
@@ -87,6 +88,72 @@ VertexAdaptor<op_func_call> CheckFuncCallsAndVarargPass::process_varargs(VertexA
...
+  auto v_alloc = VertexAdaptor<op_alloc>::create().set_location(call_location);
+  v_alloc->allocated_class = klass_CompileTimeLocation;
+
+  auto v_loc = VertexAdaptor<op_func_call>::create(v_alloc, v_file, v_function, v_line).set_location(call_location);
+  v_loc->func_id = klass_CompileTimeLocation->members.get_instance_method("__construct")->function;
+  v_loc->extra_type = op_ex_constructor_call;
+  return v_loc;
+}
+
+VertexAdaptor<op_func_call> CheckFuncCallsAndVarargPass::add_call_arg(VertexPtr to_add, VertexAdaptor<op_func_call> call, bool prepend) {
```
quasilyte, line 138: I found only 1 usage of this function and it looks like it's always called with `prepend=false`.
Is that intended?
Do you have plans to use this function later on in different contexts where prepend could be true?

compiler/pipes/check-func-calls-and-vararg.cpp
unserialize: Done. Forgot about this function

compiler/pipes/check-func-calls-and-vararg.cpp
unserialize: Yes, it's intentionally. This function is also used in generics and json (there branches are in a development stage yet)

quasilyte (approved):

quasilyte (approved):

PR #500, opened 2022-04-22
support 'object' keyword, which auto-converts a function into generics

unserialize:
This PR brings some kind of support for the `object` keyword.
Like `callable`, it's not a real type, because there can't be "just any callable", likewise there can't be "just any object". Nevertheless, `object` keyword can be useful in various cases if is has special treatment.

## 1) `object` as a function parameter acts like `@kphp-template`
This:
```
function f(object $o) {
  $o->method();
}
```
Is now the same as
```
/** 
 * @kphp-template T
 * @kphp-param T $o
 */
function f($o) {
  $o->method();
}
```
Like `callable`, a standalone `object` implicitly converts a function into a template one.

## 2) `object` in `_functions.txt` means "any object"
For example, before this MR, `instanse_serialize()` was declared like this:
```
function instance_serialize($instance ::: any) ::: string | null;
```
Now, it's declared in this way:
```
function instance_serialize(object $instance) ::: string | null;
```
So, if passed not an instance, KPHP will fail on a compilation stage, not on gcc stage.

## 3) Better errors for using `callable`/`object` in a wrong way
For example, if one writes `/** @var callable[] */`, KPHP will give an error when applying this phpdoc. Before, it failed on gcc phase.

## 4) In future generics, `object` can be used as a "where" restriction
In a branch with generics, `function f(object $o)` will be converted to `function f<T:object>(T $o)`, which disallows passing primitives / arrays of objects / etc. to this function. The alternative syntax will also be available, since "object" is now a valid type hint.
```
/** 
 * @kphp-template T:object
 * @param T $o
 */
function f($o)
```

### DIFF
--- a/builtin-functions/_functions.txt
+++ b/builtin-functions/_functions.txt
@@ -321,7 +321,7 @@ function is_string ($v ::: mixed) ::: bool;
 function is_array ($v ::: any) ::: bool;
 function is_object ($v ::: any) ::: bool;
 function get_class ($v ::: any) ::: string;
-function get_hash_of_class ($klass ::: any) ::: int;
+function get_hash_of_class (object $klass) ::: int;
 function print_r ($v ::: any, $buffered ::: bool = false) ::: string;
 function var_export ($v ::: any, $buffered ::: bool = false) ::: string;
 function print ($v ::: string) ::: int;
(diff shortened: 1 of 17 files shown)


unserialize:
Closing this PR, `object` keyword will appear in PR with generics implementation

unserialize:
Closing this PR, object keyword will appear in PR with generics implementation

PR #511, opened 2022-04-28
Dummy visitor methods for all builtin classes

DrDet:


drdzyk (approved):

drdzyk (commented):
I suggest to inherit from `DummyVisitorMethods` privately in order to prevent polymorphic references, and add this one line: `using DummyVisitorMethods::accept;` to each successors.

drdzyk (approved):

PR #251, opened 2021-05-31
add value-based function call rewrites for some builtin funcs

quasilyte:
Rewrite patterns:

	microtime() -> microtime_string()
	microtime(false) -> microtime_string()
	microtime(true) -> microtime_float()

	preg_match($pat, $s, $m) -> ({ $tmp = preg_match_strings($pat, $s); $m = $tmp[1]; $tmp[0]; })
	preg_match($pat, $s, $m, 0, $offset) -> ({ $tmp = preg_match_strings($pat, $s, $offset); $m = $tmp[1]; $tmp[0] })
	preg_match_all($pat, $a, $m) -> ({ $tmp = preg_match_all_strings($pat, $s); $m = $tmp[1]; $tmp[0]; })

New functions:

	microtime_float        - microtime(true)
	microtime_string       - microtime(false)
	preg_match_strings     - like preg_match, but returns matches as string[]
	preg_match_all_strings - like preg_match_all, but returns matches as string[][]

TODO:
- [ ] Add numerical prefixes to all tests

PR #510, opened 2022-04-28
JsonEncoder and @kphp-json annotations

drdzyk:
This PR introduces `JsonEncoder` interface that allows to serialize objects to json format with some extra transformations, like renaming fields in camel case style, hiding private fields or hiding explicitly specified fields, changing float precision and others.

### 1.Basic usage
`JsonEncoder` defines `encode` and `decode` functions intended for conversions:
```
class JsonEncoder {
  const fields_rename = 'none'; // or snake_case, or camelCase
  // and some others, see below

  public static function encode(any $instance, bool $pretty_print = false, array $metadata = []) : string;

  /** @kphp-extern-func-info cpp_template_call */
  public static function decode(string $json, string $class_name) ::: instance<^2>;
}
```
Usage is quite simple:
```
class Person {
  public string $name = 'Andrew';
}

$json = JsonEncoder::encode(new Person);
$person = JsonEncoder::decode($json, Person::class);

echo $json;
// {"name":"Andrew"}
```

### 2.Encoders
It is possible to define custom encoders, which are inherited from base `JsonEncoder`. It gives ability to redefine constants that parametrize encoder behavior:
```
class ApiJsonEncoder extends JsonEncoder {
  const fields_rename = 'snake_case';
  // ...
  // undefined constants will be inherited from base class
  // it is possible to override only constants
}
```
`ApiJsonEncoder::encode($obj)` will transform all member fields of `$obj` in snake case form in output json. **So, encoders concept is intended for mass control over properties.** You can rename properties in snake case form in one part of a project, or hide all private properties in another.  It is possible to have different encoders (`ApiJsonEncoder`, `PublicFieldJsonEncoder`, etc) and use them for serialization instance of the same class. This instance will be serialized in json in a different ways, also it can be deserialized from json in its original from.
However, each class, or even each class field may override behavior of encoder, see below. 

### 3.Annotations
Annotations is a `@kphp-json` tag, written inside phpdoc over class or class field. Each tag overrides behavior of encoder. Class tag spreads across all fields of class, child classes don't inherit it. Class field tag overrides class tag and doesn't affect other fields.
List of available tags:

**@kphp-json rename=string** (over field)
Renames given class field.

**@kphp-json skip** (over field)
Not to serialize the field to json and not to expect it back from a json string.

**@kphp-json skip_if_default=true|false** (over field, over class, in encoder)
Not to serialize the field if its value is equal to default in runtime. It may also be missing during deserialization.

[...]

PR #515, opened 2022-05-05
[compiler] add hrtime() and microtime() specializations

quasilyte:
This is a port of #251 PR without more problematic regexp
function specializations and refactoring.
preg functions will be ported separately as they require
more unrelated changes to be applied.

In addition to microtime, this patch includes hrtime which
is now generally available as KPHP min PHP version is 7.4 now.

	microtime() -> microtime_string()
	microtime(false) -> microtime_string()
	microtime(true) -> microtime_float()

	hrtime() -> hrtime_array()
	hrtime(false) -> hrtime_array()
	hrtime(true) -> hrtime_int()

The main idea is to get less mixed-typed values in the code
that uses functions those return value type depends on its arguments.

### DIFF
--- a/builtin-functions/_functions.txt
+++ b/builtin-functions/_functions.txt
@@ -349,12 +349,22 @@ function getdate ($timestamp ::: int = PHP_INT_MIN) ::: mixed[];
 function gmdate ($format ::: string, $timestamp ::: int = PHP_INT_MIN) ::: string;
 function gmmktime ($h ::: int = PHP_INT_MIN, $m ::: int = PHP_INT_MIN, $s ::: int = PHP_INT_MIN, $month ::: int = PHP_INT_MIN, $day ::: int = PHP_INT_MIN, $year ::: int = PHP_INT_MIN) ::: int;
 function localtime ($timestamp ::: int = PHP_INT_MIN, $is_associative ::: bool = false) ::: mixed[];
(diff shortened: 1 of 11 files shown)


compiler/pipes/gen-tree-postprocess.cpp
```diff
@@ -89,6 +89,46 @@ VertexAdaptor<op_require> make_require_once_call(SrcFilePtr lib_main_file, Verte
...
+  }
+
+  // hrtime(true) -> hrtime_int()
+  if (GenTree::get_actual_value(args[0])->type() == op_true) {
+    return replace_call_func(call, "hrtime_int");
+  }
+
+  return call;
+}
+
```
unserialize, line 131: It's not okay to replace func calls in GenTreePostprocessPass — for many reasons. For example, 
1) if someone uses CONST_TRUE instead of true, it can't be done here
2) if someone passes wrong arguments count, we want an error to point to original function
3) we need to check all function names here, not only extern calls
4) some more

I suggest to look here: https://github.com/VKCOM/kphp/pull/510/files
Search for `maybe_replace_extern_func_call`: it's exactly what you need.
You can copy that function (dropping json from there), a comment above, and how it's called to your branch and write a body. That PR would be rebased onto master after merging yours.
Btw, in my branch with generics I also use that function.

unserialize (approved):

PR #517, opened 2022-05-18
[runtime] make shutdown funcs called even after a timeout

### LINKED ISSUE #504: Не вызывается  register_shutdown_function() при таймауте
Привет!

Не вызывается мой обработчик при таймауте
```
[pid 2516442] [time 1650535285] in sigalrm_handler
[pid 2516442] [time 1650535285] timeout exit
```

У меня стоит перехват register_shutdown_function()  но видимо он не срабатывает в этом случае

В коде обработчика SIGALRM покопался, нет там вызова пользовательского код, просто прибивает и всё.

quasilyte:
The most challenging part is that timeout is catched in a signal
handler of the running coroutine (run_context) which then
changes the state of the worker and switches the context to exit_context.
Running the shutdown handlers inside the run_context before switching it
is dangerous as arbitrary PHP code is not guaranteed to be safe for
the signal handling context.

If we run these shutdown handlers inside of the worker context,
we may suffer from the `exit()` calls done inside of the shutdown
functions that may cause us to attempt switching the context again
and entering a `finished` state instead of the `error` state we were in.

As a solution to this problem, we use a setjmp+longjmp to implement
our recovery from the `exit()` when running shutdown functions.

We only call shutdown handlers for timeout errors for now.
Executing shutdown handlers should not be a problem for some error
kinds, but some of them are more challenging (like memory limit errors).

Fixes #504

### DIFF
--- a/server/php-worker.cpp
+++ b/server/php-worker.cpp
@@ -295,6 +295,13 @@ void php_worker_run(php_worker *worker) {
       }
       case run_state_t::error: {
         vkprintf (2, "php script [req_id = %016llx]: ERROR (probably timeout)\n", worker->req_id);
+        // see #504: we need to call shutdown handlers during the timeout to make cleanup code more
+        // reliable and to stay more compatible to the way PHP works; running shutdown handlers
+        // for some other error kinds could be possible, but this should be done with care
+        // (it could be dangerous to call a PHP code in case of the memory limit being reached)
+        if (php_script_get_error_type(php_script) == script_error_t::timeout) {
+          php_script_run_shutdown_functions();
+        }
         php_script_finish(php_script);
 
         if (worker->conn != nullptr) {
@@ -326,6 +333,7 @@ void php_worker_run(php_worker *worker) {
       }
       case run_state_t::finished: {
         vkprintf (2, "php script [req_id = %016llx]: OK (still can return RPC_ERROR)\n", worker->req_id);
+        php_script_run_shutdown_functions();
         script_result *res = php_script_get_res(php_script);
         php_worker_set_result(worker, res);
         php_script_finish(php_script);
--- a/runtime/interface.cpp
+++ b/runtime/interface.cpp
@@ -8,6 +8,7 @@
 #include <cassert>
 #include <clocale>
 #include <csignal>
+#include <csetjmp>
 #include <fstream>
 #include <functional>
 #include <getopt.h>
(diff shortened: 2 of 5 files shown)


server/php-worker.cpp
```diff
@@ -326,6 +333,7 @@ void php_worker_run(php_worker *worker) {
       }
       case run_state_t::finished: {
         vkprintf (2, "php script [req_id = %016llx]: OK (still can return RPC_ERROR)\n", worker->req_id);
+        php_script_run_shutdown_functions();
         script_result *res = php_script_get_res(php_script);
```
DrDet, line 337: Now it differs from old behavior in case when shutdown functions append something to response
For example:
```php
<?php

/**
 * @kphp-required
 */
function at_finish()
{
    echo " FINISHED\n";
}
register_shutdown_function('at_finish');

echo "Answer";
```
In vanilla php and old kphp HTTP response contains 'Answer FINISHED'
Now it contains 'Answer' only

To be honest, I don't know how much it's important, but anyway

DrDet (commented):
Also try this:
```php
<?php

/**
 * @kphp-required
 */
function at_finish()
{
    echo " FINISHED\n";
    critical_error("Test assert");
}
register_shutdown_function('at_finish');

echo "Answer";
```
Now such code kills worker, although before that it doesn't

DrDet (changes requested):
I figured out that it also breaks context switches on waiting net from shutdown functions.
Example:
```php
<?php

function foo(int $i): int {
    fwrite(STDERR, "~~~ forkable_f($i) before yield\n");
    sched_yield_sleep(0.5); // wait net
    fwrite(STDERR, "~~~ forkable_f($i) after yield\n");
    return $i;
}

/**
 * @kphp-required
 */
function at_finish() {
  fwrite(STDERR, "~~~ begin at_finish() \n");
  $resp_future = fork(foo(42));
  $resp = wait($resp_future);
  fwrite(STDERR, "~~~ end at_finish() \n");
}

register_shutdown_function('at_finish');

echo "Answer\n";
```
That's because we must run shutdown functions from script context, but now it moves to net context

quasilyte:
Summarizing the comments above, we should run shutdown handlers from the script context.
For both timeout and normal script termination.

quasilyte:
Abandoned in favor of #518 that runs shutdown handlers in script context.
All cases above work correctly now.

PR #518, opened 2022-05-23
[runtime] make shutdown funcs called even after a timeout

### LINKED ISSUE #504: Не вызывается  register_shutdown_function() при таймауте

quasilyte:
This is an alternative implementation that implements shutdown functions during the timeout in the script context.

See also: #517 

----

The most challenging part is that timeout is catched in a signal
handler of the running coroutine (run_context) which then
changes the state of the worker and switches the context to exit_context.
Running the shutdown handlers inside the run_context before switching it
is dangerous as arbitrary PHP code is not guaranteed to be safe for
the signal handling context.

If we run these shutdown handlers inside of the worker context,
we may suffer from the `exit()` calls done inside of the shutdown
functions that may cause us to attempt switching the context again
and entering a `finished` state instead of the `error` state we were in.

As a solution to this problem, we use a setjmp+longjmp to implement
our recovery from the `exit()` when running shutdown functions.

We only call shutdown handlers for timeout errors for now.
Executing shutdown handlers should not be a problem for some error
kinds, but some of them are more challenging (like memory limit errors).

Fixes #504

### DIFF
--- a/runtime/interface.cpp
+++ b/runtime/interface.cpp
@@ -8,6 +8,7 @@
 #include <cassert>
 #include <clocale>
 #include <csignal>
+#include <csetjmp>
 #include <fstream>
 #include <functional>
 #include <getopt.h>
@@ -454,14 +455,20 @@ static const string_buffer *get_headers(int content_length) {//can't use static_
 }
 
 constexpr uint32_t MAX_SHUTDOWN_FUNCTIONS = 256;
+
+namespace {
+
 // i don't want destructors of this array to be called
 int shutdown_functions_count = 0;
 char shutdown_function_storage[MAX_SHUTDOWN_FUNCTIONS * sizeof(shutdown_function_type)];
 shutdown_function_type *shutdown_functions = reinterpret_cast<shutdown_function_type *>(shutdown_function_storage);
-static bool finished = false;
-static bool flushed = false;
-static bool wait_all_forks_on_finish = false;
+shutdown_functions_status shutdown_functions_status_value = shutdown_functions_status::not_executed;
+jmp_buf timeout_exit;
+bool finished = false;
+bool flushed = false;
+bool wait_all_forks_on_finish = false;
 
+} // namespace
 
 void f$fastcgi_finish_request(int64_t exit_code) {
   if (flushed) {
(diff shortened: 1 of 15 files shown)


server/php-runner.cpp
```diff
@@ -151,6 +179,39 @@ void PHPScriptBase::init(script_t *script, php_query_data *data_to_set) {
   PHPScriptBase::ml_flag = false;
 }
 
+// asan_stack_unpoison marks the script stack memory as no longer in use,
+// making it possible to do a longjmp up the stack;
+// if ASAN is disabled, this function does nothing
+// use this function right before doing a longjmp
+void PHPScriptBase::asan_stack_unpoison() {
+#if ASAN7_ENABLED
+  __asan_unpoison_memory_region(run_stack, stack_size);
```
DrDet, line 188: ASAN says users should use macros instead of functions
image

server/php-runner.cpp
```diff
@@ -399,47 +487,31 @@ volatile bool PHPScriptBase::is_running = false;
...
+      // we need to (in that order):
+      // [1] run the shutdown handlers
+      // [2] change the context to the worker
+      // [3] execute the error handling state
+      // previously, we performed [2] and [3] right here,
+      // but now we also need to run the shutdown handlers which is
+      // can contain arbitrary code and should not be executed from the signal handler;
+      // therefore, we break out of this signal handler to a prepared recovery point
+      // (without changing the script context) and then do all required steps
+      PHPScriptBase::current_script->asan_stack_unpoison();
```
DrDet, line 512: What do you think about calling this right after stack allocation here?
https://github.com/VKCOM/kphp/blob/e30c0156dc55f4412b703b11dc7d5b08a8dc4f67/server/php-runner.cpp#L100

Now it's marked as addressable after the first script timeout, that's little bit confusing

DrDet (commented):
Let's add test for checking context switch from shutdown functions, something like:

```php
<?php

function foo(int $i): int {
    fwrite(STDERR, "~~~ forkable_f($i) before yield\n");
    echo "before_yield";
    sched_yield_sleep(0.5); // wait net
    fwrite(STDERR, "~~~ forkable_f($i) after yield\n");
    echo "_after_yield";
    return $i;
}

/**
 * @kphp-required
 */
function at_finish() {
  fwrite(STDERR, "~~~ begin at_finish() \n");
  $resp_future = fork(foo(42));
  while (!wait_concurrently($resp_future)) {}
  fwrite(STDERR, "~~~ end at_finish() \n");
}

register_shutdown_function('at_finish');

echo "Answer_";

```
and check that response is "Answer_before_yield_after_yield"

server/php-runner.cpp
quasilyte: That wouldn't work. The stack gets "poisoned" when it's out of the current frame.
It gets unpoisoned automatically when we return from a function (its portion for the previous frame).

When we allocate the run stack space, it's not poisoned yet.
When we call functions on it, parts of it become poisoned.
Unless we're doing a return normally or perform some other control flow that asan understands (like exceptions, etc), it will have no idea what portions of the stack should be unpoisoned.
This is why we need to give it a hint when we do a non-local jump and only then.

server/php-runner.cpp
quasilyte: Good point.

runtime/interface.cpp
```diff
@@ -458,11 +459,12 @@ constexpr uint32_t MAX_SHUTDOWN_FUNCTIONS = 256;
 int shutdown_functions_count = 0;
 char shutdown_function_storage[MAX_SHUTDOWN_FUNCTIONS * sizeof(shutdown_function_type)];
 shutdown_function_type *shutdown_functions = reinterpret_cast<shutdown_function_type *>(shutdown_function_storage);
+static shutdown_functions_status shutdown_functions_status_value = shutdown_functions_status::not_executed;
+static jmp_buf timeout_exit;
 static bool finished = false;
 static bool flushed = false;
 static bool wait_all_forks_on_finish = false;
```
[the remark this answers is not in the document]
quasilyte: Everything that is static here or just these two variables?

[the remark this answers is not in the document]
DrDet: Some `shutdown_function*` variables are not static now. We can either make them static or replace static keywords with anonymous namespace for the entire highlighted block.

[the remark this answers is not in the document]
quasilyte: Wrapped everything into an anonymous namespace and removed static keywords.

DrDet (dismissed):

DrDet (approved):

PR #531, opened 2022-06-04
Linter to check the file header

Danil42Russia:


Danil42Russia:
We decided not to do it yet

Issue #529, opened 2022-06-03
Incorrect generation of string concatenation code in a class field


Danil42Russia:
### Examle code

```php
<?php

class A {
  public $str = "hello" . " " . "world";
}

$a = new A();
var_dump($a->str);
```

### Expected
`"hello world"`

### Current
`Compilation error at stage: Make, gen by make.cpp:438`

### Generated code
```cpp
// vars0.cpp
v$const_var$u4a9c55e2cb2329d2_0 = string ("hello", 5) concatenation string (" ", 1) concatenation string ("world", 5);
```

### Behavior in PHP 
![image](https://user-images.githubusercontent.com/6605618/171921136-8c703a4e-c9b7-41f1-bd62-7e7dc5a51415.png)

PR #532, opened 2022-06-06
[compiler] fixed concatenation for class fields

Danil42Russia:
### Information
- Type: Bugfix
- Issue: #529

### Code
```php
<?php

class A {
  public $str = "hello" . " " . "world";
}

$a = new A();
var_dump($a->str);
```

### Before
`ompilation error at stage: Make, gen by make.cpp:438`

### After 
`hello world"`

### Generated code
```cpp
// vars0.cpp
v$const_var$u4a9c55e2cb2329d2_0 = str_concat(string ("hello", 5), string (" ", 1), string ("world", 5));
```


compiler/pipes/optimization.cpp
```diff
@@ -316,6 +316,19 @@ VertexPtr OptimizationPass::on_enter_vertex(VertexPtr root) {
...
+        if (class_field.var->init_val) {
+          if (auto var_vertex = class_field.var->init_val.try_as<op_var>()) {
+            if (vk::any_of_equal(var_vertex->var_id->init_val->type(), op_string_build, op_concat)) {
+              auto &op_concat_vertex = var_vertex->var_id->init_val;
+              op_concat_vertex = optimize_string_building(op_concat_vertex);
+            }
+          }
+        }
+      });
+    }
```
unserialize, line 331: This is 100% a wrong place.
I suppose, that the only change you need to do for fixing the issue, is adding
```
        run_function_pass(class_field.var->init_val, this);
```
into `void OptimizationPass::on_finish()` at the top of `if (class_field.var->init_val)`.
This test passes after this change.
Please, check that vkcom still compiles and that vkcom codegen diff is expected.

Danil42Russia: fixed

unserialize (approved):

PR #537, opened 2022-06-17
Force common runtime template types to be instantiated in runtime sources

DrDet:
By default all template types are implicitly instantiated, e.g. the code with substituted types is placed in each translation unit that use the corresponding template. It leads to duplication of common instantiations so many times.

To get rid of this unnecessary duplicates in this PR [explicit template instantiation](https://en.cppreference.com/w/cpp/language/class_template) is applied for common `array<T>` and `optional<T>` instantiation patterns.

This not only reduces the output binary size, but also forces all such instantiations to be compiled with `-O3` that potentially can improve performance in some cases.


runtime/common_template_instantiations.cpp
```diff
@@ -0,0 +1,32 @@
...
+
+template class array<array<int64_t>>;
+template class array<array<string>>;
+template class array<array<double>>;
+template class array<array<bool>>;
+template class array<array<mixed>>;
+
+template class Optional<int64_t>;
+template class Optional<string>;
+template class Optional<double>;
```
unserialize, line 26: Optional<bool> also, it represents "null"

DrDet: It's skipped intentionally because there is an explicit specialization for `Optional<bool>`. And in this case explicit instantiation has no effect, see [cppreference](https://en.cppreference.com/w/cpp/language/class_template#:~:text=Explicit%20instantiation%20has%20no%20effect%20if%20an%20explicit%20specialization%20appeared%20before%20for%20the%20same%20set%20of%20template%20arguments.)

PR #544, opened 2022-06-24
added kphp.readyV3 request

troy4eg:


DrDet (changes requested):
Now it always works incorrectly in case when readyV3 is enabled and `get_lease_mode(lease_mode) == LeaseWorkerMode::QUEUE_TYPES` (config is not empty)

Also sometimes when size of serialized message is not multiple of 8, everything crashes on the assert

So, my suggestion is:
1. Fix test to catch these two bugs and ensure it fails on the current code version
2. Fix these bugs in the PR

DrDet (approved):

PR #557, opened 2022-07-12
[compiler] implement composer classmap autoloading

### LINKED ISSUE #49: composer support: add "autoload.classmap" support
Some projects use `["autoload"]["classmap"]` mechanism:

https://github.com/sebastianbergmann/phpunit/blob/master/composer.json#L72

See https://getcomposer.org/doc/04-schema.md#classmap

quasilyte:
`autoload.classmap` is another popular method of classes autoloading in composer.

Implementing this feature increases the number of composer packages that can be used by KPHP code.

Classmap works like this: for a given files list (dirs or regular files), collect all files with `.inc` and `.php` extension recursively and build autoloading maps for them. Unlike PSR4, classmap doesn't require any conventions. A file can have multiple classes, its name could be anything, namespaces can be arbitrary as well.

It's hard to implement this feature in KPHP without actually parsing the php files. As a compromise,
we're scanning the classmap folders and require all files found during the composer autoload file inclusion.

Fixes #49

### DIFF
--- a/compiler/composer.cpp
+++ b/compiler/composer.cpp
@@ -6,10 +6,127 @@
 
 #include <yaml-cpp/yaml.h> // using YAML parser to handle JSON files
 
+#include <dirent.h>
+#include <sys/stat.h> // TODO: remove when std::filesystem is used everywhere instead of stat
+#include <filesystem>
+
+#include "common/smart_ptrs/unique_ptr_with_delete_function.h"
 #include "common/algorithms/contains.h"
 #include "common/wrappers/fmt_format.h"
 #include "compiler/kphp_assert.h"
 #include "compiler/stage.h"
+
+namespace {
+
+void close_dir(DIR *d) {
+  closedir(d);
+}
+
+void collect_composer_folders(const std::string &path, std::vector<std::string> &result) {
+  // can't use nftw here as it doesn't provide a portable way to stop directory traversal;
+  // we don't want to visit *all* files in the vendor tree
+  //
+  // suppose we have this composer-generated layout:
+  //     vendor/pkg1/
+  //     * composer.json
+  //     * src/
+  //     vendor/ns/pkg2/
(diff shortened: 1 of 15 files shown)


compiler/composer.cpp
```diff
@@ -160,20 +302,26 @@ void ComposerAutoloader::load_file(const std::string &pkg_root, bool is_root_fil
...
       } else if (kv.second.IsScalar()) {
-        add_autoload_dir(prefix, kv.second.as<std::string>());
+        add_autoload_psr4_dir(prefix, kv.second.as<std::string>());
       } else {
         kphp_error(false, fmt_format("load composer file {}: invalid autoload psr-4 item", filename.c_str()));
       }
     }
 
+    // https://getcomposer.org/doc/04-schema.md#classmap
+    const auto classmap_src = autoload["classmap"];
```
unserialize, line 315: const auto &

compiler/composer.h
```diff
@@ -22,15 +22,7 @@ class ComposerAutoloader : private vk::not_copyable {
...
-  // load_file parses the "$pkg_root/composer.json" file and saves
-  // all relevant definitions inside loader;
-  // this method is not thread-safe and should be called only during the compiler init
-  void load_file(const std::string &pkg_root);
-
-  // load_root_file is like load_file, but should be only used
-  // for the main composer file;
-  // this method is not thread-safe and should be called only during the compiler init
-  void load_root_file(const std::string &pkg_root);
+  void load(const std::string&);
```
unserialize, line 25: (code style) spacing before &

quasilyte: Fixed.

compiler/composer.cpp
quasilyte: Fixed.

quasilyte:
Abandoned in favor of #605

PR #560, opened 2022-07-14
add function to send 103 http code

Ramzeeset:
# Add support for HTTP Status Code 103 for Indicating Early Hints

Introduction
---------------
HTTP code 103 (Early hints) informational status code indicates to the client that the server is likely to send a final response with the header fields included in the informational response.

A client can speculatively evaluate the header fields included in a 103 (Early Hints) response while waiting for the final response. For example, a client might recognize a Link header field value containing the relation type "preload" and start fetching the target resource. However, these header fields only provide hints to the client; they do not replace the header fields on the final response.

More information [here](https://datatracker.ietf.org/doc/html/rfc8297)

Implementation
--------------------
Added new function `send_http_103_early_hints($headers ::: string[]) ::: void` that send HTTP 103 response with passed headers immediately **without buffering**.
Note that headers that have been set using the `header (...)` function **will not be sent** as hints.

Testing
---------
Also was developed a test that checks the correctness of sending a response by checking the code and timeout.

### DIFF
--- a/runtime/interface.cpp
+++ b/runtime/interface.cpp
@@ -333,11 +333,11 @@ array<string> f$headers_list() {
 }
 
 void f$send_http_103_early_hints(const array<string> & headers) {
-  string header("HTTP/1.1 103 Early Hints\n");
+  string header("HTTP/1.1 103 Early Hints\r\n");
   for (const auto & h : headers) {
-    header.append(string(h.get_value()).append("\n"));
+    header.append(h.get_value().c_str()).append("\r\n");
   }
-  http_send_query(header.c_str(), header.size(), "\n", 1);
+  http_send_immediate_response(header.c_str(), header.size(), "\r\n", 2);
 }
 
 void f$setrawcookie(const string &name, const string &value, int64_t expire, const string &path, const string &domain, bool secure, bool http_only) {
--- a/runtime/interface.cpp
+++ b/runtime/interface.cpp
@@ -332,6 +332,14 @@ array<string> f$headers_list() {
   return result;
 }
 
+void f$send_http_103_early_hints(const array<string> & headers) {
+  string header("HTTP/1.1 103 Early Hints\n");
+  for (const auto & h : headers) {
+    header.append(string(h.get_value()).append("\n"));
+  }
+  http_send_query(header.c_str(), header.size(), "\n", 1);
+}
+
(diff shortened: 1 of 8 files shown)


tests/python/tests/http_server/test_request.py
```diff
@@ -9,6 +11,25 @@ def test_send_simple_request_path(self):
         self.assertEqual(response.status_code, 200)
         self.assertEqual(response.content, b"Hello world!")
 
+    def test_send_request_with_early_hints(self):
+        msg = b"GET /?hints=yes HTTP/1.1\n\n"
+        with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
+            s.connect(('127.0.0.1', self.kphp_server.http_port))
+            s.send(msg)
+            early_hints = s.recv(len(msg))
```
DrDet, line 19: let's settimeout(0.1) before that to ensure early hints is received right after call of `send_http_103_early_hints` on server side

tests/python/tests/http_server/test_request.py
```diff
@@ -9,6 +11,25 @@ def test_send_simple_request_path(self):
...
+            s.settimeout(1)
+            response = b''
+            try:
+                response = s.recv(4096)
+            except OSError:
+                pass
+            self.assertEqual(response, b'')
```
DrDet, line 26: Why is that needed?

DrDet (changes requested):

tests/python/tests/http_server/test_request.py
Ramzeeset: If the socket fails to read within the allotted timeout, an exception is thrown.
Now this part of the code has become more understandable.

DrDet (dismissed):

runtime/interface.cpp
```diff
@@ -332,6 +332,14 @@ array<string> f$headers_list() {
   return result;
 }
 
+void f$send_http_103_early_hints(const array<string> & headers) {
+  string header("HTTP/1.1 103 Early Hints\r\n");
+  for (const auto & h : headers) {
+    header.append(h.get_value().c_str()).append("\r").append("\n");
```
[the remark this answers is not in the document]
Ramzeeset: fixed

DrDet (approved):

PR #564, opened 2022-07-22
added instance_deserialize_safe and instance_serialize_safe methods

troy4eg:


runtime/msgpack-serialization.h
```diff
@@ -70,6 +70,20 @@ inline Optional<string> f$instance_serialize(const class_instance<InstanceClass>
...
 }
 
+template<class InstanceClass>
+inline string f$instance_serialize_safe(const class_instance<InstanceClass> &instance) noexcept {
+  vk::msgpack::packer_float32_decorator::clear();
+  vk::msgpack::CheckInstanceDepth::depth = 0;
+  string err_msg;
+  auto result = f$msgpack_serialize(instance, &err_msg);
+  if (vk::msgpack::CheckInstanceDepth::is_exceeded()) {
+    THROW_EXCEPTION(new_Exception(string(__FILE__), __LINE__, string("maximum depth of nested instances exceeded")));
```
DrDet, line 80: `THROW_EXCEPTION` macro doesn't return from the current function, you need to do it manually. You can check other usages

runtime/msgpack-serialization.h
```diff
@@ -70,6 +70,20 @@ inline Optional<string> f$instance_serialize(const class_instance<InstanceClass>
   return result;
 }
 
+template<class InstanceClass>
+inline string f$instance_serialize_safe(const class_instance<InstanceClass> &instance) noexcept {
```
DrDet, line 74: I think it should return `Optional<string>` too: `nullopt` when our exception is 'thrown'

runtime/msgpack-serialization.h
```diff
@@ -129,3 +143,13 @@ template<class ResultClass>
 inline ResultClass f$instance_deserialize(const string &buffer, const string&) noexcept {
   return f$msgpack_deserialize<ResultClass>(buffer);
 }
+
+template<class ResultClass>
+inline ResultClass f$instance_deserialize_safe(const string &buffer, const string&) noexcept {
+  string err_msg;
+  auto res = f$msgpack_deserialize<ResultClass>(buffer, &err_msg);
+  if (!err_msg.empty()) {
+    THROW_EXCEPTION (new_Exception(string(__FILE__), __LINE__, err_msg));
```
DrDet, line 152: and here

DrDet (changes requested):
I think it's worth extracting the common code between `f$instance_serialize` and `f$instance_serialize_safe` to reduce copypaste

runtime/msgpack-serialization.h
```diff
@@ -53,17 +53,25 @@ inline string f$msgpack_serialize_safe(const T &value) noexcept {
   }
   return res.val();
 }
-
 template<class InstanceClass>
-inline Optional<string> f$instance_serialize(const class_instance<InstanceClass> &instance) noexcept {
+inline Optional<string> f$common_instance_serialize(const class_instance<InstanceClass> &instance, string *out_err_msg) noexcept {
```
DrDet, line 57: `f$` at start of the name means there's such a function in `functions.txt`, but it's not the case.
Let's rename it to smth like `instance_serialize_impl`

DrDet (approved):

builtin-functions/_functions.txt
```diff
@@ -1302,8 +1302,12 @@ function instance_cache_update_ttl(string $key, int $ttl = 0) ::: bool;
 function instance_cache_delete(string $key) ::: bool;
 
 function instance_serialize($instance ::: any) ::: string | null;
+/** @kphp-extern-func-info can_throw */
+function instance_serialize_safe($instance ::: any) ::: string;
```
unserialize, line 1306: Here is 'string', but in polyfills it's '?string'

PR #570, opened 2022-08-03
[runtime] run shutdown handlers for null iface object method calls

### LINKED ISSUE #569: Implement exception handling for null reference by interface
Hi!

Need to call  shutdown function when null reference exception caused.


```php
<?php

register_shutdown_function(function () {
  $a = error_get_last();

  if (!$a) return;//ini_set("display_errors", 0);
  echo "\ncatch error\n";
});


try {
interface I {
  public function f();
}

class C implements I {
  public function f() {}
}

function f(I $i) {
  $i->f();
}

f(null);

}catch(Throwable $e ){
  echo "\ncatch ex\n";
}

echo "\ndone\n";
```
```
------- kphp result---------------------
[1659472461] [2705875] Warning: Critical error "call method(I::f) on null object" in file /home/comm/data/projs/kphp-test/kphp_out/kphp/o_23/I@@f.cpp on line 26
------- Stack Backtrace -------
(0) ./kphp_out/cli : php_error(char const*, ...)+0x9c [0x57fe7c]
(1) ./kphp_out/cli : f$I$$f(class_instance<C$I> const&)+0xdf [0x4b97e5]
(2) ./kphp_out/cli : f$src_testNullReferenceInterfacee745a436ffd789c9()+0x6d [0x4b98af]
(3) ./kphp_out/cli : f$src_testNullReferenceInterfacee745a436ffd789c9$run()+0xd [0x4b9697]
(4) ./kphp_out/cli : PHPScriptBase::run()+0xb1 [0x61e7e1]
(5) /lib/x86_64-linux-gnu/libc.so.6 : +0x5b510 [0x7fcd2a0bc510]
-------------------------------

[pid 2705875] [time 1659472461] in php_assert_handler (SIGRTMIN+1 signal)
[pid 2705875] [time 1659472461] php assert error
[2705875][2022-08-02 23:34:21.494522 php-runner.cpp  325] Critical error during script execution: php assert error


------- end result---------------------
------- php result---------------------

catch ex

done
------- php end result---------------------
```


Other case: need same behaviour for KPHP for null reference on class:
```php
<?php
class Foo {                                                                                                       
  public $x = 10;                                                                                                 
  public function f() { echo $this->x; }                                                                          
  public static function get():?self
  {
    
    return null;
  }
}          

register_shutdown_function(function () {
  $a = error_get_last();

  if (!$a) return;//ini_set("display_errors", 0);
  echo "\ncatch error\n";
});

                                                                                                       
                                                                                                                  
$foo = Foo::get();                                                                                                 

try {
  $foo->f();
}catch(Throwable $e ){
  echo "\ncatch ex\n";
}

echo "\ndone\n";
```
```
------- kphp result---------------------
[1659471062] [2702052] Warning: Trying to access property of null object
------- Stack Backtrace -------
(0) ./kphp_out/cli : php_warning(char const*, ...)+0x9c [0x57fddc]
(1) ./kphp_out/cli : class_instance<C$Foo>::warn_on_access_null() const+0x2e [0x4b9a52]
(2) ./kphp_out/cli : f$src_testNullReference97986cf4cc5b98a4()+0xb2 [0x4b97c8]
(3) ./kphp_out/cli : f$src_testNullReference97986cf4cc5b98a4$run()+0xd [0x4b940e]
(4) ./kphp_out/cli : PHPScriptBase::run()+0xb1 [0x61e7e1]
(5) /lib/x86_64-linux-gnu/libc.so.6 : +0x5b510 [0x7f0da06a1510]
-------------------------------

10
done

------- end result---------------------
------- php result---------------------

catch ex

done
------- php end result---------------------
```

quasilyte:
When null interface method is called, the script ends abnormally.
This disrupts the normal script termination process and shutdown
handlers will never be executed.

Since calling a method on a null object is not a fatal error inside
our runtime (we can still continue running some other KPHP code),
we can and should do something like a normal `f$exit(1)` and allow
the shutdown handlers to be executed.

To do that, a new runtime function is introduced: `soft_critical_error()`.
The only difference is that it calls `f$exit(1)` instead of doing
some signal raising business.

We may want to use this new function in other places, where recoverable
critical errors are preferred to the current criticals.
For now, it's only used in one place.

It's also debatable whether we should switch to non-fatal critical errors
in the future, so user-generated critical error do not bypass the
shutdown handlers (I'm not sure users expect that behavior anyway).

Fixes #569

### DIFF
--- a/builtin-functions/_functions.txt
+++ b/builtin-functions/_functions.txt
@@ -164,6 +164,16 @@ function error_reporting ($e ::: int = TODO) ::: int;
 function warning ($message ::: string) ::: void;
 /** @kphp-no-return */
 function critical_error($message ::: string) ::: void;
+/**
+ * Raise a soft critical error.
+ * Unlike critical_error() this function will let the script to finish normally.
+ * In particular, the shutdown handlers will be executed.
+ *
+ * It will behave almost identically to critical_error() if -K runtime flag was used.
+ *
+ * @kphp-no-return
+ */
+function soft_critical_error($message ::: string) ::: void;
 /** @kphp-no-return */
 function exit($code = 0) ::: void;
 /** @kphp-no-return */
--- a/compiler/pipes/generate-virtual-methods.cpp
+++ b/compiler/pipes/generate-virtual-methods.cpp
@@ -55,8 +55,12 @@ VertexAdaptor<op_func_call> generate_critical_error_call(std::string msg) {
   auto msg_v = VertexAdaptor<op_string>::create();
   msg_v->str_val = std::move(msg);
 
+  // this error is bad, we can't continue the execution;
+  // but it's not bad enough to interrupt the normal script termination routine,
+  // so we emit a soft critical error instead of a harsh one;
+  // this way, shutdown handlers can be executed (#569)
   auto call_ce = VertexAdaptor<op_func_call>::create(msg_v);
-  call_ce->str_val = "critical_error";
+  call_ce->str_val = "soft_critical_error";
   call_ce->func_id = G->get_function(call_ce->str_val);
 
   return call_ce;
--- a/runtime/kphp_core.h
+++ b/runtime/kphp_core.h
@@ -464,6 +464,8 @@ inline int64_t f$error_reporting();
 
 inline void f$warning(const string &message);
 
+#define f$soft_critical_error(message) \
+  php_soft_critical_error("%s", (message).c_str())
 #define f$critical_error(message) \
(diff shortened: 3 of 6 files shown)


DrDet (commented):
When soft critical error is raised, script terminates via `finished` state instead of `error` one.
I think, it's not fully correct. If such script was executed in HTTP server mode, it would respond with 200 OK and body containing some pieces of response that were added to output buffer before null dereferencing happened. Like everything is fine. But it is an error actually and we should use appropriate response code at least

PR #571, opened 2022-08-03
allow dirname() calls inside require_once()

drdzyk:
This PR adds support of `dirname()` calls inside `require_once()`, so that both next `require_once()` calls will be equals:
```<?php
require_once dirname(__FILE__) . '/filename.php';
require_once __DIR__ . '/filename.php';
```

### DIFF
--- a/compiler/const-manipulations.h
+++ b/compiler/const-manipulations.h
@@ -4,6 +4,7 @@
 
 #pragma once
 
+#include <libgen.h>
 #include <string>
 
 #include "common/php-functions.h"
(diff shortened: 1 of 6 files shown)


compiler/const-manipulations.h
```diff
@@ -246,13 +247,20 @@ struct CheckConstAccess final
 };
 
 // moved from the collect-required-and-classes.cpp to make it more reusable
-static inline std::string collect_string_concatenation(VertexPtr v) {
+static inline std::string collect_string_concatenation(VertexPtr v, bool allow_dirname = false) {
   if (auto string = v.try_as<op_string>()) {
     return string->str_val;
   }
+  if (auto call = v.try_as<op_func_call>(); allow_dirname && call && call->str_val == "dirname") {
+    kphp_assert_msg(call->size() == 1, "call to 'dirname' must have one arg");
```
quasilyte, line 255: I think it makes sense to use a "compile error" level here and add a `kphp_should_fail` test that verifies that we give an expected error.
`kphp_should_fail` are cheap and executed almost instantly.

drdzyk: done

quasilyte (approved):

compiler/const-manipulations.h
```diff
@@ -246,13 +247,20 @@ struct CheckConstAccess final
 };
 
 // moved from the collect-required-and-classes.cpp to make it more reusable
-static inline std::string collect_string_concatenation(VertexPtr v) {
+static inline std::string collect_string_concatenation(VertexPtr v, bool allow_dirname = false) {
   if (auto string = v.try_as<op_string>()) {
     return string->str_val;
   }
+  if (auto call = v.try_as<op_func_call>(); allow_dirname && call && call->str_val == "dirname") {
```
unserialize, line 254: Does it work if called with leading slash? `\dirname()`

drdzyk: Yes it does, I've checked.

PR #575, opened 2022-08-04
add function ignore_user_abort

Ramzeeset:
# Add a new function ignore_user_abort analogue from PHP

Introduction
---------------
**ignore_user_abort** — Set whether aborting should abort script execution. This implementation **matches the description** [here](https://www.php.net/manual/en/function.ignore-user-abort.php) with **one exception** unlike the implementation in PHP, the current version considers the **nesting level** and does not complete the script until it becomes zero

Description
---------------
```php
ignore_user_abort(?bool $enable = null): int
```

- **Parameter List**
Takes a flag **value** and sets it with the **appropriate** nesting level **change**. If the argument is **null**, then returns the current level **without changing** it

- **Return Values**
Return nesting level previous value

Usage
--------

```php
function important_work(...) {
    ignore_user_abort(true);
    ...
    ignore_user_abort(false);

// start critical section which will be executed regardless of the state of the connection
ignore_user_abort(true);
important_work(...);
// some other work to be carried out as rpc, resumable or function calls
ignore_user_abort(false);
// termination of the script if the connection was interrupted
```

Testing
---------
This pr also contains tests to check the correct operation of the server using this function

### DIFF
--- a/tests/python/tests/http_server/php/index.php
+++ b/tests/python/tests/http_server/php/index.php
@@ -51,7 +51,6 @@ class RpcWorker implements I {
     }
 
     public function work() {
-        fwrite(STDERR, $this->port . "\n");
         $job = function() {
             $conn = new_rpc_connection('localhost', $this->port, 0, 5);
             $req_id = rpc_tl_query_one($conn, ["_" => "engine.sleep",
@@ -59,7 +58,9 @@ class RpcWorker implements I {
             $resp = rpc_tl_query_result_one($req_id);
             assert($resp['result']);
         };
-        $job();
+        for ($i = 0; $i < 3; ++$i) {
+            $job();
+        }
         fwrite(STDERR, "test_ignore_user_abort/finish_rpc_work_" . $_GET["level"] . "\n");
    }
 }
--- a/tests/python/tests/http_server/php/index.php
+++ b/tests/python/tests/http_server/php/index.php
@@ -8,6 +8,65 @@ class A {
   public $b = "hello";
 }
 
+/**
+ * @kphp-required
+ */
+function shutdown_function() {
+    fwrite(STDERR, "shutdown_function was called\n");
+}
+
+function assert($flag) {
+    if ($flag !== true) {
+        critical_error("failed assert");
+    }
+}
+
+interface I {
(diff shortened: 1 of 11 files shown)


tests/python/tests/http_server/php/index.php
```diff
@@ -8,6 +8,55 @@ class A {
...
+        $this->port = $port;
+    }
+
+    public function work() {
+        fwrite(STDERR, $this->port . "\n");
+        $job = function() {
+            $conn = new_rpc_connection('localhost', $this->port, 0, 3);
+            $req_id = rpc_tl_query_one($conn, ["_" => "engine.sleep",
+                                                "time_ms" => 60]);
+            $resp = rpc_tl_query_result_one($req_id);
```
DrDet, line 50: let's add an assertion that response is TL Bool true

tests/python/tests/http_server/php/index.php
```diff
@@ -8,6 +8,55 @@ class A {
...
+        $job = function($x) {
+                sched_yield_sleep(2);
+                return $x;
+        };
+        $futures = [];
+        for ($i = 0; $i < 10; ++$i) {
+            $futures[] = fork($job($i));
+        }
+        wait_multi($futures);
+        fwrite(STDERR, "test_ignore_user_abort/finish_resumable_work\n");
```
DrDet, line 33: And here

DrDet (changes requested):

tests/python/tests/http_server/php/index.php
Ramzeeset: fixed

tests/python/tests/http_server/php/index.php
Ramzeeset: fixed

DrDet (changes requested):

tests/python/tests/http_server/php/index.php
Ramzeeset: fix

DrDet (approved):

DrDet (dismissed):

DrDet (approved):

PR #580, opened 2022-08-05
add more information about workers in logs

Ramzeeset:


server/php-engine-vars.h
```diff
@@ -21,6 +21,8 @@
 #define DEFAULT_SCRIPT_TIMEOUT 30
 #define MAX_SCRIPT_TIMEOUT (60 * 7)
 
+enum class ProcessType { master, general_worker, job_worker };
```
DrDet, line 24: Let's distinguish `rpc_worker` and  `http_worker` instead of single `general_worker`

We can do this by checking that `rpc_clients` are not empty. In that case it's `rpc_worker`, otherwise - `http_worker`
https://github.com/VKCOM/kphp/blob/2c3594f04c432dd262f2cc38efadb7e59fde5501/server/lease-rpc-client.h#L24

server/json-logger.cpp
```diff
@@ -12,6 +12,7 @@
 #include "common/fast-backtrace.h"
 #include "common/wrappers/likely.h"
 #include "server/json-logger.h"
+#include "server/php-worker.h"
```
DrDet, line 15: why do we need this include?

server/json-logger.cpp
```diff
@@ -224,6 +225,16 @@ void JsonLogger::write_log(vk::string_view message, int type, int64_t created_at
...
     json_out_it->append_raw(tags_);
   }
+  if (process_type == ProcessType::master) {
+    json_out_it->append_key("process_type").append_string("master");
+  } else if (process_type == ProcessType::general_worker) {
+    json_out_it->append_key("process_type").append_string("general_worker");
+    json_out_it->append_key("worker_unique_index").append_integer(logname_id);
+  } else if (process_type == ProcessType::job_worker) {
+    json_out_it->append_key("process_type").append_string("job_worker");
+    json_out_it->append_key("worker_unique_index").append_integer(logname_id);
```
DrDet, line 235: and here

tests/python/lib/kphp_server.py
```diff
@@ -145,11 +145,21 @@ def assert_json_log(self, expect, message="Can't wait expected json log", timeou
...
+                        and (got_tags.get("worker_unique_index") is None or
+                             isinstance(got_tags.get("worker_unique_index"), int)) \
+                        and (isinstance(got_tags.get("pid"), int)):
                     expected_record_copy = expected_record.copy()
                     expected_record_copy["msg"] = ""
                     json_log_record_copy = json_log_record.copy()
                     json_log_record_copy["msg"] = ""
+                    json_log_record_copy["tags"].pop("process_type")
+                    json_log_record_copy["tags"].pop("pid")
+                    json_log_record_copy["tags"].pop("worker_unique_index", "No Key found")
```
DrDet, line 162: I think you should do it in another place, somewhere there:
https://github.com/VKCOM/kphp/blob/2c3594f04c432dd262f2cc38efadb7e59fde5501/tests/python/lib/kphp_server.py#L125

Also it's more handy to write like `got_tags.get("process_type") in ['master', 'general_worker', 'job_worker']` instead of multiple `get` and `or`

DrDet (changes requested):

server/json-logger.cpp
DrDet, line 235: let's call it `logname_id`
```suggestion
    json_out_it->append_key("logname_id").append_integer(logname_id);
  } else if (process_type == ProcessType::job_worker) {
    json_out_it->append_key("process_type").append_string("job_worker");
    json_out_it->append_key("logname_id").append_integer(logname_id);
```

server/json-logger.cpp
Ramzeeset: change include to make it more understandable

server/php-master.cpp
```diff
@@ -596,6 +597,13 @@ int run_worker(WorkerType worker_type) {
     pid = getpid();
 
     master_sfd = -1;
+    if (worker_type == WorkerType::job_worker) {
+      process_type = ProcessType::job_worker;
+    } else if (RpcClients::get().rpc_clients.size() > 0) {
```
DrDet, line 602:
```suggestion
    } else if (!RpcClients::get().rpc_clients.empty()) {
```

DrDet (approved):

PR #582, opened 2022-08-05
support Modulite inside kphp (everything except composer)

unserialize:
# Modulite is a conception of isolating parts of code

PHP, as a language, allows calling any public class/function from any other place in the code. For huge PHP projects it's bad practice, because if you develop some kernel db connectors and public API, you expect your users to use this API, but you have no guarantee that nobody calls your internals directly, even if you expose a Composer package. PHP has no "modules" in a language — and that's what Modulite would take care about.

The concept of colored functions is more flexible, but Modulite is more generalized. It's designed in a way to be integrated into IDE.

The description below does not pretend to be a documentation. Once I finish all implementation (in multiple steps), along with Modulite plugin for PhpStorm, along with Psalm/PhpStan integration — I'll create a giant landing describing its features and usage scenarios. Nevertheless, a brief intro is below.

# A directory can now contain `.modulite.yaml` file
```
name: "@some-name"
namespace: "This\\Namespace\\"
export:
  - ... # list of exported symbols
force-internal:
  - ... # list of methods/constants/fields unaccessible even if a class is exported
requires:
  - ... # dependencies to other modulites and global code
allow-internal-access:
  # exceptions, to allow external specific functions/methods access internal symbols
```

Say, you have a directory
```
Rpc/
  .modulite.yaml
  Logger.php   // class Rpc\Logger
  Query.php    // class Rpc\Query
```
Where `.modulite.yaml` is
```yaml
name: "@rpc"
namespace: "Rpc\\"
exports:
  - "Query"
requires:
```

It means, that `Query` is public, and `Logger` is internal. `Logger` can be accessed only within @rpc modulite, but trying to access it from the outside would give a compilation error:
```php
// Feed/Post.php
Rpc\Logger::logRaw(...);  // error: Rpc\Logger is internal in @rpc
```

**That's how it works from the one side — you can't access internal symbols from external code.**

**From the other side — Modulite controls, how you access external code from within your modulite.** `requires` is a list of dependencies you are allowed to access. Say, it's empty, then this is incorrect:
```
// Rpc/Logger.php
$user_id = id();  // error: id() is not listed in requires
```
Only adding `id()` to requires in yaml would allow this code to compile. As a result, no unexpected calls can appear in the code without specifying them explicitly. And when you specify them, they are visible in merge requests, they are stored in git, etc. 
```
requires:
  - "id()"
```

# All in all, Modulite disallows uncontrolled entropy growth

Modulite operates all kind of symbols: classes, functions, constants, static fields, globals. Even `define()` made inside a modulite must be exported in order to be used outerwise. All `global` symbols are needed to be explicitly specified in `requires`. And so on.

BTW, instance methods are not analyzed, because type inferring would indeed differ between IDE and KPHP, which will lead to misunderstanding. Other kinds of symbols are expected to have no difference in analysis.

# Nested modulites

Nested modulites are also supported. Say, @msg is a parent modulite, with 3 children — @msg/channels, @msg/folders and @msg/core. Core is pretended to be used only within @msg, extern code can't use it, but it can use channels and folders (only exported parts of theirs, of course). 
```
// Messages/.modulite.yaml
name: "@msg"
export:
  - "@msg/channels"
  - "@msg/folders"
```

# Composer packages 

Composer packages are not supported now, they will be supported in the future.

# Implementation details

KPHP searches for `.modulite.yaml` files in a manner Node.JS does for `package.json`: it traverses the tree up until one found. In order to implement this, `SrcDir` and `dirs_ht` were introduced, every `SrcFile` has `->dir` now.

`ModuliteData` stores all parsed info from a yaml file.

Every function / class / constant now belong to some modulite: `FunctionData::modulite`, `ClassData::modulite`, `DefineData::modulite`. It's null for global code. They are filled like this: all functions in a file belong to the same modulite => all files in a directory belong to the same modulite => `f->modulite = f->file_id->dir->nested_files_modulite`. Same for others.

Two sync pipes were renamed:
* `CheckRequires` => `WaitForAllClassesAndLoadModulitesF`
* `CalcRealDefinesValuesF ` => `CalcRealDefinesAndAssignModulitesF`

Yaml files are parsed in the first one, after all classes and functions have been discovered, as yaml parsing involves symbols lookup by name.
`FunctionData::modulite` and others are assigned in the latter.
Read detailed comments in cpp file, why the logic was placed inside that exact pipes (short answer: we rely on sync points).

Registering defines pass was moved very early, so that inlining defines step would already know all info about modulites.

Take a look at the beautiful `modulite-check-rules.h`. It exports 5 functions:
```
void modulite_check_when_use_class(...);
void modulite_check_when_use_constant(...);
void modulite_check_when_call_function(...);
void modulite_check_when_use_static_field(...);
void modulite_check_when_global_keyword(...);
```
All the rest code is written just to be able to call these functions :)

# PhpStorm integration

Writing yaml manually is very disturbing, but that's not what you should do actually: Modulite plugin for PhpStorm checks your code on the fly, highlighting errors and providing quick fixes which lead to yaml modification. It also has wizards for modulites creating, renaming and extracting, couples by lots of inspections and useful UI integrations. It will be open sourced in late autumn.

IDE integration is possible, because we need only to check two edges (caller and callee), we don't need to have a full call graph.

# Future development 

This MR is waiting to be merged, despite it's ready only partially — we are going to start using Modulite in VK, so we need its experimental support in production. It's disabled by default and can be enabled by `KPHP_ENABLE_MODULITE` env.


compiler/compiler-core.cpp
```diff
@@ -194,6 +196,20 @@ FFIRoot &CompilerCore::get_ffi_root() {
   return ffi;
 }
 
+vk::string_view CompilerCore::calc_relative_name(SrcFilePtr file, bool builtin) const {
+  vk::string_view full_file_name = file->file_name;
+  if (full_file_name.starts_with(settings_->base_dir.get())) {
+    return full_file_name.substr(settings_->base_dir.get().size());
+    return full_file_name.substr(settings_->base_dir.get().size());
```
Danil42Russia, line 203: copypasted :)

Ahmed-su-do:
في الجمعة، ٥ أغسطس ٢٠٢٢ ٥:١١ م Aleksandr Kirsanov ***@***.***>
كتب:

compiler/compiler-core.cpp
unserialize: fixed

i582 (approved):

compiler/compiler-core.cpp
```diff
@@ -227,27 +243,55 @@ SrcFilePtr CompilerCore::register_file(const std::string &file_name, LibPtr owne
...
+
+    TSHashTable<SrcDirPtr>::HTNode *node_parent_dir = dirs_ht.at(vk::std_hash(parent_dir_name));
+    if (!node_parent_dir->data) {
+      AutoLocker<Lockable *> locker(node_parent_dir);
+      if (!node_parent_dir->data) {
+        node_parent_dir->data = register_dir(parent_dir_name);
+      }
+    }
+    dir->parent_dir = node_parent_dir->data;
+//    printf("%s -> %s\n", dir->full_dir_name.c_str(), dir->parent_dir->full_dir_name.c_str());
```
quasilyte, line 289: A debug leftover?

PR #592, opened 2022-08-17
add initial SPL ArrayIterator class support

quasilyte:
ArrayIterator provides a way to iterate the array without
using `foreach` statement or legacy `next()`/`current()` APIs.

In PHP it has almost identical performance to the above mentioned
pair of `next()`/`current()`.
In KPHP it works with minimal overhead when compared to a normal
`foreach` statement. This means that is can be used in lower
level components that need to iterate the array in unusual way.

Some performance comparison results:

	Direct1000         7065.0 ns/op   0 B/op      0 allocs/op
	ArrayIterator1000  8375.0 ns/op   0 B/op      0 allocs/op
	CopyKeys1000       40541.0 ns/op  16088 B/op  1 allocs/op
	Reslice1000        74655.0 ns/op  88088 B/op  1001 allocs/op

Direct - `foreach` array iteration.
ArrayIterator uses the newly introduced class.
CopyKeys uses `array_keys()` to iterate the array.
Reslice uses `array_slice()` to extract the array portion.

The 0 allocations are possible with KPHP function `reset_array_iterator`
that can be used to re-use the array iterator objects when possible.
This function is trivially polyfilled in PHP.

Until KPHP supports generic classes, this ArrayIterator only works
with `mixed[]` typed arrays.

### DIFF
--- a/runtime/spl/array_iterator.h
+++ b/runtime/spl/array_iterator.h
@@ -1,0 +1,62 @@
+// Compiler for PHP (aka KPHP)
+// Copyright (c) 2022 LLC «V Kontakte»
+// Distributed under the GPL v3 License, see LICENSE.notice.txt
+
+#pragma once
+
+#include "common/algorithms/hashes.h"
+#include "common/wrappers/string_view.h"
+
+#include "runtime/dummy-visitor-methods.h"
+#include "runtime/kphp_core.h"
+#include "runtime/refcountable_php_classes.h"
+
+// C$ArrayIterator implements SPL ArrayIterator class.
+struct C$ArrayIterator : public refcountable_php_classes<C$ArrayIterator>, private DummyVisitorMethods {
+  // we store an array to keep a living reference to it while iterator is valid;
+  // also we may implement rewind() method later if we feel like it
+  array<mixed> arr;
+  array<mixed>::const_iterator it;
+  array<mixed>::const_iterator end;
+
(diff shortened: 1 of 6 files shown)


runtime/spl/array_iterator.h
```diff
@@ -0,0 +1,62 @@
...
+#include "runtime/kphp_core.h"
+#include "runtime/refcountable_php_classes.h"
+
+// C$ArrayIterator implements SPL ArrayIterator class.
+struct C$ArrayIterator : public refcountable_php_classes<C$ArrayIterator>, private DummyVisitorMethods {
+  // we store an array to keep a living reference to it while iterator is valid;
+  // also we may implement rewind() method later if we feel like it
+  array<mixed> arr;
+  array<mixed>::const_iterator it;
+  array<mixed>::const_iterator end;
```
unserialize, line 20: Does PHP version of ArrayIterator work when an array is modified?
This version does not, since modifying arr could leave these iterators stale.

quasilyte: There are tests that check this behavior.
ArrayIterator iterates over the provided array without reference.
So if that array is modified externally, we'll end up iterating over a copy that is held by the ArrayIterator.

unserialize: I see, thank you

unserialize (approved):

PR #593, opened 2022-08-21
add stripcslashes function

quasilyte:
We already have addcslashes, so having the opposite operation
does make sense. This function is useful when parsing string
literals with C-style escape sequences like `\n` etc.

### DIFF
--- a/runtime/string_functions.cpp
+++ b/runtime/string_functions.cpp
@@ -1213,6 +1213,104 @@ string f$sprintf(const string &format, const array<mixed> &a) {
   return result;
 }
 
+string f$stripcslashes(const string &str) {
+  if (str.empty()) {
+    return str;
+  }
+
+  // this implementation is an adapted version from php-src
+
+  auto len = str.size();
+  auto new_len = len;
+  string result(len, false);
+  char *result_c_str = &result[0];
+  char num_tmp[4]; // we need up to three digits + a space for null-terminator
+  int j = 0;
+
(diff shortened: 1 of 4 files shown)


runtime/string_functions.cpp
```diff
@@ -1213,6 +1213,98 @@ string f$sprintf(const string &format, const array<mixed> &a) {
...
+          }
+          /* fall-through */
+        default:
+          j = 0;
+          while (i < len && str[i] >= '0' && str[i] <= '7' && j < 3) {
+            num_tmp[j++] = str[i++];
+          }
+          if (j) {
+            num_tmp[j] ='\0';
+            *result_c_str++ =static_cast<char>(strtol(num_tmp, nullptr, 8));
```
unserialize, line 1290: (code style) space after =

runtime/string_functions.cpp
```diff
@@ -1213,6 +1213,98 @@ string f$sprintf(const string &format, const array<mixed> &a) {
...
+              new_len -= 3;
+            } else {
+              num_tmp[1] = '\0';
+              new_len -= 2;
+            }
+            *result_c_str++ = static_cast<char>(strtol(num_tmp, nullptr, 16));
+            break;
+          }
+          /* fall-through */
+        default:
```
unserialize, line 1283: please, add a // comment, what we are doing here

unserialize (approved):

PR #598, opened 2022-08-31
add DateTime class

drdzyk:
This PR adds some functionality of php [DateTime](https://www.php.net/manual/en/class.datetime.php) class. Namelly, the following functions:

```

class DateTime {
  /** @kphp-extern-func-info can_throw */
  public function __construct(string $datetime = "now", ?DateTimeZone $timezone = null);
  public static function createFromFormat(string $format, string $datetime, ?DateTimeZone $timezone = null): ?DateTime;
  public static function getLastErrors(): array|false;
  public function modify(string $modifier): ?DateTime;
  public function setDate(int $year, int $month, int $day): DateTime;
  public function setISODate(int $year, int $week, int $dayOfWeek = 1): DateTime;
  public function setTime(
      int $hour,
      int $minute,
      int $second = 0,
      int $microsecond = 0
  ): DateTime;
  public function setTimestamp(int $timestamp): DateTime;
  public function format(string $format): string;
}
```

### DIFF
--- a/runtime/timelib_wrapper.cpp
+++ b/runtime/timelib_wrapper.cpp
@@ -631,3 +631,10 @@ std::pair<bool, string> php_timelib_date_modify(timelib_time *t, const string &m
 
   return {true, {}};
 }
+
+void php_timelib_date_date_set(timelib_time *t, int64_t y, int64_t m, int64_t d) {
+  t->y = y;
+  t->m = m;
+  t->d = d;
+  timelib_update_ts(t, nullptr);
+}
--- a/runtime/timelib_wrapper.cpp
+++ b/runtime/timelib_wrapper.cpp
@@ -308,16 +308,7 @@ Optional<array<mixed>> php_timelib_date_get_last_errors() {
   return false;
 }
 
-static const char *const mon_full_names[] = {"January", "February", "March",     "April",   "May",      "June",
-                                             "July",    "August",   "September", "October", "November", "December"};
-
-static const char *const mon_short_names[] = {"Jan", "Feb", "Mar", "Apr", "May", "Jun", "Jul", "Aug", "Sep", "Oct", "Nov", "Dec"};
-
-static const char *const day_full_names[] = {"Sunday", "Monday", "Tuesday", "Wednesday", "Thursday", "Friday", "Saturday"};
-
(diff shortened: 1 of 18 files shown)


runtime/timelib_wrapper.cpp
```diff
@@ -215,3 +225,511 @@ std::pair<int64_t, bool> php_timelib_strtotime(const string &tz_name, const stri
...
+std::pair<timelib_time *, string> php_timelib_date_initialize(const string &tz_name, const string &time_str, const char *format) {
+  auto script_guard = make_malloc_replacement_with_script_allocator();
+
+  const string &time_str_new = time_str.empty() ? NOW : time_str;
+  timelib_error_container *err = nullptr;
+  timelib_time *t = format
+    ? timelib_parse_from_format(format, time_str.c_str(), time_str.size(), &err, timelib_builtin_db(), timelib_parse_tzfile)
+    : timelib_strtotime(time_str_new.c_str(), time_str_new.size(), &err, timelib_builtin_db(), timelib_parse_tzfile);
+
+  /* update last errors and warnings */
```
quasilyte, line 267: I'm not sure it's common to use `/**/` style  comments in our codebase (except for some top-level doc comments).

runtime/timelib_wrapper.cpp
```diff
@@ -215,3 +225,511 @@ std::pair<int64_t, bool> php_timelib_strtotime(const string &tz_name, const stri
...
+  if (err && err->error_count) {
+    /* spit out the first library error message, at least */
+    return {false, gen_parse_error_msg(*err, modifier)};
+  }
+
+  std::memcpy(&t->relative, &tmp_time->relative, sizeof(timelib_rel_time));
+  t->have_relative = tmp_time->have_relative;
+  t->sse_uptodate = 0;
+
+  if (tmp_time->y != -99999) {
```
quasilyte, line 651: Maybe add a named const for `-99999`?
It's declared in timelib as follows:
```
#define TIMELIB_UNSET   -99999
```

runtime/timelib_wrapper.cpp
drdzyk: I prefer `//` comments too, `/**/` appears just from copy-paste, fixed.

runtime/timelib_wrapper.cpp
drdzyk: fixed

quasilyte (approved):

PR #601, opened 2022-08-31
clean engine logs at the end of each test

Ramzeeset:


tests/python/lib/testcase.py
```diff
@@ -107,6 +107,20 @@ def custom_setup(cls):
...
+    def teardown_method(self, method):
+        self.custom_teardown_method(method)
+
+    @classmethod
+    def custom_setup_method(cls, method):
+        pass
+
+    @classmethod
+    def custom_teardown_method(cls, method):
+        pass
```
DrDet, line 122: These methods shouldn't be static
Similar as basic `setup_method` and `teardwon_method`

DrDet (changes requested):

tests/python/lib/testcase.py
Ramzeeset: fix

tests/python/lib/testcase.py
```diff
@@ -217,6 +229,14 @@ def custom_teardown(cls):
...
                 raise RuntimeError("Got unexpected sanitizer log '{}'".format(sanitizer_log))
 
+    @classmethod
+    def custom_setup_method(cls, method):
+        pass
+
+    @classmethod
+    def custom_teardown_method(cls, method):
+        cls.kphp_server._engine_logs = []
+
```
DrDet, line 239: and here

Ramzeeset: fix

DrDet (approved):

PR #610, opened 2022-09-07
[runtime] add fragmentation-related info to OOM error messages

quasilyte:
This change adds some extra info to the OOM error messages to make fragmentation issues tracking easier.

The `buffer_free_space` measures how much memory the script buffer has left when the OOM happened. Ideally, this value should be low, like `0.1`, meaning there is only 10% memory left.

When calculating the fragmentation human-readable hint, we're taking the size of the allocation into the account.

Examples:
```
Warning: Can't allocate 524328 bytes (normal fragmentation, buffer free space: 0.70)
Warning: Can't allocate 72 bytes (low fragmentation, buffer free space: 0.02)
```

### DIFF
--- a/runtime/memory_resource/monotonic_buffer_resource.cpp
+++ b/runtime/memory_resource/monotonic_buffer_resource.cpp
@@ -35,4 +35,32 @@ void monotonic_buffer::critical_dump(void *mem, size_t size) const noexcept {
     stats_.real_memory_used, stats_.max_real_memory_used);
 }
 
+void monotonic_buffer_resource::raise_oom(size_t size) const noexcept {
+  auto mem_pool_size = stats_.memory_limit;
+  size_t mem_available = mem_pool_size - stats_.memory_used;
+  if (mem_available < size) {
+    // to avoid the confusion, don't report this case as "high fragmentation";
+    // this is a genuine OOM case when user requested too much memory
+    php_out_of_memory_warning("Can't allocate %zu bytes (out of memory)", size);
+  } else {
+    // try to add some meaningful info to the OOM error message;
+    // buffer_free_space is a percentage of memory we have at our disposal
+    // if this value is high, then we have significant memory fragmentation issues;
+    // values below 0.1 are very good
+    auto fragmentation_rate = static_cast<double>(mem_available - size) / static_cast<double>(mem_pool_size);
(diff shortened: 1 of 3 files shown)


runtime/memory_resource/monotonic_buffer_resource.cpp
```diff
@@ -35,4 +35,32 @@ void monotonic_buffer::critical_dump(void *mem, size_t size) const noexcept {
...
+    // to avoid the confusion, don't report this case as "high fragmentation";
+    // this is a genuine OOM case when user requested too much memory
+    php_out_of_memory_warning("Can't allocate %zu bytes (out of memory)", size);
+  } else {
+    // try to add some meaningful info to the OOM error message;
+    // buffer_free_space is a percentage of memory we have at our disposal
+    // if this value is high, then we have significant memory fragmentation issues;
+    // values below 0.1 are very good
+    auto fragmentation_rate = static_cast<double>(mem_available - size) / static_cast<double>(mem_pool_size);
+    auto buffer_free_space = static_cast<double>(mem_available) / static_cast<double>(mem_pool_size);
```
DrDet, line 51: Why do we need `fragmentation_rate` when we have `buffer_free_space`?

DrDet (dismissed):

runtime/memory_resource/unsynchronized_pool_resource.cpp
```diff
@@ -80,7 +80,15 @@ void *unsynchronized_pool_resource::allocate_small_piece_from_fallback_resource(
     fallback_resource_.init(smallest_huge_piece, details::memory_chunk_tree::get_chunk_size(smallest_huge_piece));
     memory_debug("fallback resource was empty, took piece from huge map\n");
   }
-  return fallback_resource_.get_from_pool(aligned_size);
+  // call get_from_pool with safe=true, so we can handle OOM here,
+  // inside the script allocator;
+  // without that, we'll get fallback resource memory stats used when
+  // calculating the memory usage rate (fragmentation impact)
+  mem = fallback_resource_.get_from_pool(aligned_size, true);
+  if (!mem) {
```
quasilyte, line 88: I should probably add `unlikely` hint to this branch.

DrDet (approved):

PR #612, opened 2022-09-09
Improve cluster name

DrDet:
This patch consists of the following improvements:
- Allow cluster name to contain digits and '-'
- Check cluster name for incorrect symbols on start instead of force quiet replacing them with '_'
- Handle the case when cluster name is longer than 108 bytes to be able to create `UNIX_SOCKET` for graceful restart. Before this patch, server just crashed when attempting to create such socket on graceful restart.

### DIFF
--- a/server/cluster-name.cpp
+++ b/server/cluster-name.cpp
@@ -2,11 +2,16 @@
 // Copyright (c) 2020 LLC «V Kontakte»
 // Distributed under the GPL v3 License, see LICENSE.notice.txt
 
-#include <algorithm>
-#include <cstring>
-#include <cctype>
-
 #include "server/cluster-name.h"
+
+#include <algorithm>
+
+#include <cctype>
+#include <cinttypes>
+#include <cstdio>
+#include <cstring>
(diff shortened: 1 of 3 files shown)


server/cluster-name.cpp
```diff
@@ -2,19 +2,24 @@
...
+#include "common/crc32.h"
 
 ClusterName::ClusterName() {
   set_cluster_name("default");
 }
 
 const char *ClusterName::set_cluster_name(const char *name) noexcept {
   // reserved for shmem, statsd, socket suffixes and prefixes
-  constexpr size_t reserved = 32;
+  constexpr size_t reserved = 42;
```
Ramzeeset, line 22: Maybe it should be looks like
```c++
constexpr size_t reserver = std::strlen(socket_prefix)
``` 
To avoid magic numbers

DrDet: Get rid of magic number and add the necessary logic for it

Ramzeeset (approved):

PR #616, opened 2022-09-12
[compiler] better type inference for switch cond var

quasilyte:
Instead of producing a mixed-typed variable for every switch statement, try to find a more suitable type.

Most of the switches are optimized to int and string switches, so in those cases having non-mixed var is very beneficial.

Even for expr (var) switches we can get a measurable improvement if all cases and condition expr itself evaluate to some properly-typed value, like int or string.

Here is a non-compherensive list of why mixed-typed vars are not optimal:

* Expensive destructor, `mixed::~mixed` shows up in the CPU profiles
* The need to do `mixed::as_string()` many times inside optimized string switches
* mixed-typed overloadings are used in `eq2()` calls

The implementation used in this PR tries to avoid excessive work from the compiler. If it can be assumed to be a int/string switch, we won't do any extra work. For other switch statements we'll find an LCA type of condition expr and all expression cases results. It could still be mixed, but sometimes it's inferred to something like int and improves performance significantly.

	name                      old time/op  new time/op  delta
	MultiSwitch::MultiString  83.0ns ± 0%  59.3ns ± 1%  -28.55%  (p=0.000 n=8+10)
	MultiSwitch::MultiInt     23.3ns ± 3%   7.0ns ± 0%  -69.96%  (p=0.000 n=10+10)

### DIFF
--- a/compiler/pipes/collect-main-edges.cpp
+++ b/compiler/pipes/collect-main-edges.cpp
@@ -18,7 +18,46 @@
 #include "compiler/inferring/restriction-non-void.h"
 #include "compiler/inferring/type-node.h"
 #include "compiler/type-hint.h"
+#include "common/php-functions.h"
 
+namespace {
+
+SwitchKind get_switch_kind(VertexAdaptor<op_switch> s) {
+  int num_const_int_cases = 0;
+  int num_const_string_cases = 0;
+  int num_value_cases = 0;
+
+  for (auto one_case : s->cases()) {
+    if (one_case->type() == op_default) {
+      continue;
+    }
+
+    num_value_cases++;
+    auto val = GenTree::get_actual_value(one_case.as<op_case>()->expr());
+    if (is_const_int(val)) {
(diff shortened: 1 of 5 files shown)


compiler/pipes/collect-main-edges.cpp
```diff
@@ -18,7 +18,45 @@
 #include "compiler/inferring/restriction-non-void.h"
 #include "compiler/inferring/type-node.h"
 #include "compiler/type-hint.h"
+#include "common/php-functions.h"
 
+namespace {
+
+SwitchKind get_switch_kind(VertexAdaptor<op_switch> s) {
+  bool all_cases_are_ints = true;
```
unserialize, line 26: maybe, rewrite to `num_const_int_cases` to be similar to others?

quasilyte: Done.

unserialize (approved):

PR #623, opened 2022-09-16
enable leak sanitizer

drdzyk:
Leak sanitizer(lsan) by default run along with address sanitizer(asan), but in case of kphp lsan fails every time with error:
image
There are 2 reason of that:
1) User code works within coroutine, whose stack memory is protected with `mprotect(run_stack, getpagesize(), PROT_NONE)` call. That makes lsan failed to examine stack memory.
2) __sanitizer_start_switch_fiber/__sanitizer_finish_switch_fiber sanitizer's hooks called incorrect. Quoting [comment](https://github.com/llvm-mirror/compiler-rt/blob/69445f095c22aac2388f939bedebf224a6efcdaf/include/sanitizer/common_interface_defs.h#L302) from sanitizer's source code:
> /// Before switching to a different stack, you must call
/// <c>__sanitizer_start_switch_fiber()</c> with a pointer to the bottom of the
/// destination stack and with its size. When code starts running on the new
/// stack, it must call <c>__sanitizer_finish_switch_fiber()</c> to finalize
/// the switch.

So, __sanitizer_finish_switch_fiber should be called at very start of coroutine, while in our code it called at the end. It's confirmed by internal [test](https://github.com/llvm/llvm-project/blob/main/compiler-rt/test/asan/TestCases/Linux/swapcontext_annotation.cpp) of sanitzers in llvm trunk.

Let's check how it works finally. Add artificial memory leak into our `echo` function, and call it:
```
// in c++
void f$echo(const string& s) {
  std::malloc(8); // <-- it is a leak
  print(s);
}

// in php
<?php
echo "hello\n";
```
The above script compiled with kphp and launched now produce next output:
 
image

Note, that for now I keep lsan disabled in our tests since they show some amount of leaks. I'm going to fix memory leaks and enable lsan in tests in separate patch.

### DIFF
--- a/server/php-runner.h
+++ b/server/php-runner.h
@@ -64,8 +64,8 @@ void init_handlers();
  * It stores state of the script: current execution point, pointers to allocated script memory, stack for script context, etc.
  */
 class PhpScript {
-  double cur_timestamp, net_time, script_time;
-  int queries_cnt;
+  double cur_timestamp{0}, net_time{0}, script_time{0};
+  int queries_cnt{0};
   int long_queries_cnt{0};
 
 private:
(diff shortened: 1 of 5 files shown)


unserialize (dismissed):

server/php-runner.h
```diff
@@ -59,46 +58,63 @@ void dump_query_stats();
 
 void init_handlers();
 
+class CoroutineStack : vk::not_copyable {
```
DrDet, line 61: Such a name can be confusing. Coroutines in kphp don't have their own stack actually, but use the single common one.
Let's rename it to smth like 'CustomStack' or 'PhpScriptStack'

DrDet (approved):

PR #625, opened 2022-09-19
DO NOT SUBMIT [runtime,compiler] use custom ctype functions implementations

quasilyte:
Functions from `<cctype>` are not inlined by `g++` or `clang++` even with `-O2` and `-O3`. This causes the major slowdown in multiple functions, `strtolower`, `strtoupper`, `strcasecmp`, `lcfirst,` `ucfirst` - just to name a few.

These custom functions should be used when php ctype functions will be added too.

This patch replaces all usages of `ctype.h` functions from both runtime and compiler.

Tests check that new functions return identical results for all possible inputs (values from 0 to 255, inclusive).

`strcasecmp` benchmark results for difference input string lengths:

	name               old time/op  new time/op  delta
	Strcasecmp::X32     344ns ± 2%    90ns ± 2%  -73.76%  (p=0.000 n=10+10)
	Strcasecmp::X256   2.74µs ± 1%  0.48µs ± 0%  -82.58%  (p=0.000 n=10+8)
	Strcasecmp::X1024  10.9µs ± 3%   1.8µs ± 0%  -83.12%  (p=0.000 n=10+8)
	[Geo mean]         2.17µs       0.43µs       -80.24%

### DIFF
--- a/common/php-functions.h
+++ b/common/php-functions.h
@@ -6,11 +6,82 @@
 
 #include <climits>
 #include <cstring>
-#include <cctype>
 #include <limits>
 #include <type_traits>
 
 #include "common/sanitizer.h"
+
+namespace impl {
+
+// for functions that can be implemented as efficient range check, we
+// don't use a lookup table if compilers can generate more efficient code there;
+// for complex classes as XDIGIT we'll do one memory read.
+extern const uint8_t php_ascii_char_props[256];
+constexpr uint8_t CHAR_PROP_ALPHA = 1 << 0; // [a-zA-Z]
+constexpr uint8_t CHAR_PROP_ALNUM = 1 << 1; // [a-zA-Z0-9]
+constexpr uint8_t CHAR_PROP_PUNCT = 1 << 2;
+constexpr uint8_t CHAR_PROP_SPACE = 1 << 3;
+constexpr uint8_t CHAR_PROP_XDIGIT = 1 << 4; // [a-fA-F0-9]
+
+extern const unsigned char php_tolower_conv_table[256];
+extern const unsigned char php_toupper_conv_table[256];
+
+} // namespace impl
+
+inline bool php_isalnum(char ch) noexcept {
+  return (impl::php_ascii_char_props[static_cast<unsigned char>(ch)] & impl::CHAR_PROP_ALNUM) != 0;
+}
+
+inline bool php_isalpha(char ch) noexcept {
+  return (impl::php_ascii_char_props[static_cast<unsigned char>(ch)] & impl::CHAR_PROP_ALPHA) != 0;
+}
+
+inline bool php_iscntrl(char ch) noexcept {
+  unsigned char u = ch;
+  return u <= 31 || u == 127;
+}
+
+inline bool php_isdigit(char ch) noexcept {
+  return ch >= '0' && ch <= '9';
+}
+
+inline bool php_isgraph(char ch) noexcept {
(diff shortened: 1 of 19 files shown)


common/php-functions.h
```diff
@@ -12,6 +12,78 @@
 
 #include "common/sanitizer.h"
 
+namespace impl {
+
+// for functions that can be implemented as efficient range check, we
+// don't use a lookup table is compilers can generate more efficient code there;
```
DrDet, line 18: is it a typo?
```suggestion
// don't use a lookup table as compilers can generate more efficient code there;
```

common/php-functions.h
```diff
@@ -12,6 +12,78 @@
```
DrDet, line 12: Let's remove `#include <cctype>` or `#include <ctype.h>` from everywhere as it shouldn't be used anymore

runtime/string_functions.cpp
```diff
@@ -2011,8 +2009,8 @@ string f$strtoupper(const string &str) {
 
   // same optimization as in strtolower
   const char *end = str.c_str() + n;
-  const char *lowercase_pos = std::find_if(str.c_str(), end, [](unsigned char ch) {
-    return ch >= 'a' && ch <= 'z';
+  const char *lowercase_pos = std::find_if(str.c_str(), end, [](char ch) {
+    return php_islower(ch);
   });
```
DrDet, line 2014: The lambda is redundant here
```suggestion
  const char *lowercase_pos = std::find_if(str.c_str(), end, php_islower);
```

runtime/string_functions.cpp
```diff
@@ -1983,12 +1983,10 @@ string f$strtolower(const string &str) {
...
-  // use memcpy to copy everything before that pos;
-  // note: do not use islower() here, the compiler does not inline that function call;
-  // it could be beneficial to use 256-byte LUT here, but SIMD approach could be even better
+  // use memcpy to copy everything before that pos
   const char *end = str.c_str() + n;
-  const char *uppercase_pos = std::find_if(str.c_str(), end, [](unsigned char ch) {
-    return ch >= 'A' && ch <= 'Z';
+  const char *uppercase_pos = std::find_if(str.c_str(), end, [](char ch) {
+    return php_isupper(ch);
   });
```
DrDet, line 1990: and here

common/php-functions-test.cpp
```diff
@@ -0,0 +1,59 @@
+// Compiler for PHP (aka KPHP)
```
DrDet, line 1: Can we move this file somewhere to `tests/cpp`?

common/php-functions.h
quasilyte: It should have been `if`. Thank you. :)

common/php-functions-test.cpp
quasilyte: Done.

quasilyte:
Well. It's harder than it looks.

In some places, PHP functions use their own notion of space/digit/etc.
In some other places, ctype functions are used (or their wrappers), those can depend on the current locale setting.
It also looks like there is a subtle difference in ctype behavior on linux and macos (hence the tests failure).
It's still possible to use optimized ctype implementations in some cases; it's also possible to write functions that check that the current locale is never changed, so optimized routines can be used instead. Still, that's more complicated than I anticipated.

Let's make replacements step-by-step, without global changes. We don't have enough locale tests to make this kind of experiment right now.

PR #631, opened 2022-09-28
Add extra info to udp socket errors

Ramzeeset:


runtime/udp.cpp
```diff
@@ -25,23 +25,26 @@ static long long opened_udp_sockets_last_query_num = -1;
...
-  error_number = error_no;                              \
-  error_description = CONST_STRING(error);              \
-  RETURN
-
-#define RETURN_ERROR_FORMAT(error_no, format, ...)                   \
-  error_number = error_no;                                           \
-  error_description = f$sprintf (                                    \
-    CONST_STRING(format), array<mixed>::create(__VA_ARGS__));          \
-  RETURN
+  auto quit = [&](int socket_fd, const Optional<string> & extra_info = Optional<string>()) {
```
DrDet, line 28:
```suggestion
  auto quit = [&](int socket_fd, const Optional<string> & extra_info = {}) {
```

runtime/udp.cpp
```diff
@@ -97,12 +105,14 @@ static Stream udp_stream_socket_client(const string &url, int64_t &error_number,
   sock_fd = socket(AF_INET, SOCK_DGRAM, 0);
   if (sock_fd == -1) {
     dl::leave_critical_section();
-    RETURN_ERROR(-4, "Can't create udp socket");
+    set_error(-4, "Can't create udp socket");
+    return quit(sock_fd, Optional<string>(string("System call \'socket(...)\' got error \'").append(strerror(errno)).append("\'")));
```
DrDet, line 109: `Optional<T>` is implicitly convertible of `T`
```suggestion
    return quit(sock_fd, string("System call \'socket(...)\' got error \'").append(strerror(errno)).append("\'"));
```

Also can we write single quotes without slash?

runtime/udp.cpp
```diff
@@ -97,12 +105,14 @@ static Stream udp_stream_socket_client(const string &url, int64_t &error_number,
...
   }
   if (connect(sock_fd, reinterpret_cast <const sockaddr *> (&addr), addrlen) == -1) {
     if (errno != EINPROGRESS) {
       dl::leave_critical_section();
-      RETURN_ERROR(-5, "Can't connect to udp socket");
+      set_error(-5, "Can't connect to udp socket");
+      return quit(sock_fd, Optional<string>(string("System call \'connect(...)\' got error \'").append(strerror(errno)).append("\'")));
```
DrDet, line 115: the same as above

runtime/udp.cpp
```diff
@@ -112,7 +122,14 @@ static Stream udp_stream_socket_client(const string &url, int64_t &error_number,
     double left_time = end_time - microtime_monotonic();
     if (left_time <= 0 || poll(&poll_fds, 1, timeout_convert_to_ms(left_time)) <= 0) {
       dl::leave_critical_section();
-      RETURN_ERROR(-6, "Can't connect to udp socket");
+      Optional<string> extra_info;
```
DrDet, line 125: the same as above it can be just `string`

runtime/udp.cpp
Ramzeeset: fix

runtime/udp.cpp
Ramzeeset: fix

runtime/udp.cpp
Ramzeeset: fix

DrDet (approved):

PR #633, opened 2022-09-28
[compiler] add a special case for int64_t min value codegen

quasilyte:
Producing `-9223372036854775808_i64` literal will result in C++ compiler warning about integer overflow.

The recommended way to express min value is `(-9223372036854775807_i64 - 1_i64)`.

### DIFF
--- a/compiler/code-gen/vertex-compiler.cpp
+++ b/compiler/code-gen/vertex-compiler.cpp
@@ -9,6 +9,7 @@
 
 #include "common/wrappers/field_getter.h"
 
+#include "common/wrappers/likely.h"
 #include "compiler/code-gen/common.h"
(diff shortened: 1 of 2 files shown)


compiler/code-gen/vertex-compiler.cpp
```diff
@@ -28,6 +28,19 @@
 
 namespace {
 
+struct IntLit {
+  vk::string_view value;
+
+  void compile(CodeGenerator &W) const {
+    if (value == "-9223372036854775808") {
```
unserialize, line 35: wrap with `unlikely()`?

quasilyte: Done.

unserialize (approved):

PR #643, opened 2022-10-06
turn on memory leak sanitizer in tests

drdzyk:
This PR turns on memory leak sanitizer in tests. Also memory leaks have been fixed in timelib_wrapper.cpp and suppressed in regexp.cpp.

### DIFF
--- a/runtime/datetime/timelib_wrapper.cpp
+++ b/runtime/datetime/timelib_wrapper.cpp
@@ -1,8 +1,11 @@
 #include "runtime/datetime/timelib_wrapper.h"
 
-#include <kphp/timelib/timelib.h>
-#include <sys/time.h>
 #include <functional>
+#include <kphp/timelib/timelib.h>
+#if ASAN_ENABLED
+#include <sanitizer/lsan_interface.h>
+#endif
+#include <sys/time.h>
 
 #include "common/containers/final_action.h"
 #include "common/smart_ptrs/singleton.h"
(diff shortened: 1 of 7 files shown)


DrDet (commented):
Does it report about bugs like this now?
https://github.com/VKCOM/kphp/pull/607

drdzyk:
> Does it report about bugs like this now? #607

yes, I've already mentioned it in our internal chat:
image

runtime/timelib_wrapper.cpp
```diff
@@ -167,17 +170,36 @@ bool php_timelib_is_valid_date(int64_t m, int64_t d, int64_t y) {
   return y >= 1 && y <= 32767 && timelib_valid_date(y, m, d);
 }
 
+static timelib_time *timelib_strtotime_leak_safe(const string &time, timelib_error_container **errors) {
+  timelib_time *t = timelib_strtotime(time.c_str(), time.size(), errors, timelib_builtin_db(), timelib_parse_tzfile);
+  // sometimes 't->tz_info' contains allocated timezone object, but sometimes doesn't. And we are unable to merely delete this object
+  // because it can be shared among different e.g. DateTimeImmutable vs DateTime objects. Given that it's quite rare situation this leak is just suppressed.
+  // (it's not a leak in 'php-src' because all 'tz_info *' there are cached. We are unable cache it in kphp because of our memory managment)
```
DrDet, line 177: Why are we unable to cache it in kphp? Timelib caches it in heap, doesn't it?

drdzyk: Because these objects allocated in script memory, thus its lifetime limited by lifetime of request, but cache leaves forever. php-src cache it, while timelib itself doesn't

DrDet (dismissed):

DrDet (approved):

PR #647, opened 2022-10-12
[compiler] move some helper methods from GenTree to VertexUtil

quasilyte:
Most of the time when some code includes "gentree.h" it only needs one or two methods like `get_actual_value` or `create_int_const`.

Since our parser code is already messy as it is, adding even more helpers is not very convenient as it makes gentree files even more bloated.

Moving helpers to another file/class gives these benefits:

1. Less code includes "gentree.h" => faster compilation
2. Less code inside gentree files => less complexity in the parser files
3. Easier to maintain helpers (files that contain them are smaller and simpler)
4. It's easier to refactor the parser and helpers separately
  4.1. We can separate AST generation like `create_int_const` and other ops if we want to
  4.2. GenTree can be treated as parser, not as "everything related to vertices + parsing + more"

This is only a first step.

### DIFF
--- a/compiler/const-manipulations.h
+++ b/compiler/const-manipulations.h
@@ -10,12 +10,14 @@
 #include "common/php-functions.h"
 #include "compiler/data/data_ptr.h"
 #include "compiler/data/define-data.h"
-#include "compiler/gentree.h"
+#include "compiler/data/var-data.h"
+#include "compiler/compiler-core.h"
 #include "compiler/name-gen.h"
 #include "compiler/operation.h"
 #include "compiler/pipes/check-access-modifiers.h"
 #include "compiler/utils/string-utils.h"
 #include "compiler/vertex.h"
+#include "compiler/vertex-util.h"
 
 template<typename T>
 struct ConstManipulations {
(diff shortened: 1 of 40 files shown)


compiler/const-manipulations.h
```diff
@@ -10,7 +10,9 @@
 #include "common/php-functions.h"
 #include "compiler/data/data_ptr.h"
 #include "compiler/data/define-data.h"
-#include "compiler/gentree.h"
+#include "compiler/data/var-data.h"
+#include "compiler/vertex-util.h"
+#include "compiler/compiler-core.h"
```
unserialize, line 15: We usually try to keep includes sorted (here and everywhere below, when gentree.h replaced by vertex-util.h)

compiler/pipes/generate-virtual-methods.cpp
```diff
@@ -11,6 +11,7 @@
 #include "compiler/data/function-data.h"
 #include "compiler/data/src-file.h"
 #include "compiler/gentree.h"
+#include "compiler/vertex-util.h"
```
unserialize, line 14: gentree.h still needed?

compiler/vertex-util.cpp
```diff
@@ -0,0 +1,120 @@
...
+
+bool VertexUtil::is_constructor_call(VertexAdaptor<op_func_call> call) {
+  return !call->args().empty() && call->str_val == ClassData::NAME_OF_CONSTRUCT;
+}
+
+bool VertexUtil::is_positive_constexpr_int(VertexPtr v) {
+  auto actual_value = get_actual_value(v).try_as<op_int_const>();
+  return actual_value && parse_int_from_string(actual_value) >= 0;
+}
+bool VertexUtil::is_const_int(VertexPtr root) {
```
unserialize, line 96: Empty line missing

compiler/vertex-util.h
```diff
@@ -0,0 +1,78 @@
...
+  static void func_force_return(VertexAdaptor<op_function> func, VertexPtr val = {});
+
+  static bool is_superglobal(const std::string &s);
+  static bool is_constructor_call(VertexAdaptor<op_func_call> call);
+  static bool is_positive_constexpr_int(VertexPtr v);
+  static bool is_const_int(VertexPtr root);
+};
+
+// TODO: does it really needs to be a template function?
+// we get ~0 benefits from it being a templated function
```
unserialize, line 38: I don't think it should really be left a template function, let's just use a parameter and move to .cpp

compiler/vertex-util.h
```diff
@@ -0,0 +1,78 @@
...
+    case tp_mixed:
+      return VertexAdaptor<op_conv_mixed>::create(x).set_location(x);
+
+    default:
+      return x;
+  }
+}
+
+// TODO: does it really needs to be a template function?
+// we get ~0 benefits from it being a templated function
```
unserialize, line 69: same: let's move to .cpp

compiler/pipes/generate-virtual-methods.cpp
quasilyte: Yes. It uses `GenTree::create_switch_vertex` which is only needed inside gentree and 1 more place: this file.
I tried to move only things that are generally useful.
`create_switch_vertex` could be a candidate too, but I avoided it for the first iteration.

compiler/vertex-util.h
quasilyte: Done.

compiler/vertex-util.h
quasilyte: Done.

compiler/const-manipulations.h
quasilyte: Fixed.

quasilyte:
ping

unserialize (approved):

PR #648, opened 2022-10-13
add DateTimeImmutable

drdzyk:
This PR adds some functionality of php [DateTimeImmutable](https://www.php.net/manual/en/class.datetimeimmutable.php) class. Namelly, the following functions:
```
class DateTimeImmutable implements DateTimeInterface {
  /** @kphp-extern-func-info can_throw */
  public function __construct(string $datetime = "now", ?DateTimeZone $timezone = null);
  public static function createFromFormat(string $format, string $datetime, ?DateTimeZone $timezone = null): ?DateTimeImmutable;
  public static function createFromMutable(DateTime $object): DateTimeImmutable;
  public static function getLastErrors(): array|false;
  public function modify(string $modifier): ?DateTimeImmutable;
  public function setDate(int $year, int $month, int $day): DateTimeImmutable;
  public function setISODate(int $year, int $week, int $dayOfWeek = 1): DateTimeImmutable;
  public function setTime(
      int $hour,
      int $minute,
      int $second = 0,
      int $microsecond = 0
  ): DateTimeImmutable;
  public function setTimestamp(int $timestamp): DateTimeImmutable;
  public function format(string $format): string;
}
```


quasilyte:
LGTM

quasilyte (approved):

PR #653, opened 2022-10-20
optimize float to string conversion

drdzyk:
Currently conversion from double to string performed inside `string(double f)` constructor using `snprintf` call. There is new 'Grisu' algorithm for that purpose, which is few times faster than one from `printf` family. You can read more about this [here](http://www.serpentine.com/blog/2011/06/29/here-be-dragons-advances-in-problems-you-didnt-even-know-you-had/). I know few implementations of Grisu algo: [made by algo author](https://github.com/google/double-conversion), from [rapidjson](https://github.com/Tencent/rapidjson) and from [libfmt](https://github.com/fmtlib/fmt).  I chose libfmt since we've already used it in our code.

Benchmark shows that on regular double values new algo is about 3 times faster:
```
------------------------------------------------------------------
Benchmark                        Time             CPU   Iterations
------------------------------------------------------------------
BM_fmt_new/zero_double         212 ns          212 ns      3367623
BM_fmt_new/small_double        217 ns          217 ns      3214906
BM_fmt_new/mid_double          221 ns          221 ns      3165814
BM_fmt_new/long_double         218 ns          218 ns      3194925

BM_fmt_old/zero_double         279 ns          279 ns      2501699
BM_fmt_old/small_double        245 ns          245 ns      2869304
BM_fmt_old/mid_double          652 ns          652 ns      1063836
BM_fmt_old/long_double         655 ns          655 ns      1061488
```

In order to run this benchmark:
1) `git checkout 806116bbc89ec6a19fa572692d5ba5a53b3b0641`
2) `cmake -Bbuild_benchmark && make -Cbuild_benchmark string-float-benchmark`
3) `./build_benchmark/string-float-benchmark`

`string(double f)` constructor is used in all places where conversion from double to string needed like e.g: `operator<<(string_buffer &sb, double f)`, in numerous conversions where arguments first converted to strings, `f$strval()`, encode json, etc. So I guess this patch will give some performance benefit.


drdzyk:
I've also tested performance using `ktest` for `strval`, `array_diff_assoc` and `JsonEncoder`. Such benchmark BenchmarkGrisu.php:
```
<?php

class JsonRepr {
  public string $str = "foo";
  public float $f = 1.12345;
  public array $a1 = ["a" => 1.1234, "b" => 123.341435, "c" => 0.13245];
  public array $a2 = ["foo", "bar", "baz"];
}

class BenchmarkGrisu {
  private JsonRepr $json_obj;
  function __construct() {
    $this->json_obj = new JsonRepr;
  }

  public function benchmarkStrval() {
    $s1 = strval(0.12345);
    $s1 += strval(1.2);
    $s1 += strval(123.4567);
    $s1 += strval(123456789.123456789);
    return $s1;
  }

  public function benchmarkArrayDiffAssoc() {
    $array = ["a" => 1.12345, "b" => 324.34524, "c" => 2.00000000000001];
    return array_diff_assoc(["a" => "foo", "b" => "bar", "c" => "baz"], $array);
  }

  public function benchmarkJsonEncoder() {
    return JsonEncoder::encode($this->json_obj);
  }
}
```
Gives next results:
```
slava@slava-x1:~/vk/kphp/dev$ ../../ktest/vendor/bin/ktest benchstat old.txt new.txt 
name                   old time/op  new time/op  delta
Grisu::Strval          2.59µs ± 0%  1.30µs ± 0%  -49.69%  (p=0.000 n=10+9)
Grisu::ArrayDiffAssoc  1.44µs ± 6%  0.85µs ± 1%  -41.06%  (p=0.000 n=10+9)
Grisu::JsonEncoder     1.65µs ± 6%  1.09µs ± 0%  -33.69%  (p=0.000 n=10+9)
```
Meaning that with new algo code runs faster in 1.5 - 2 times.

unserialize:
I am strictly against adding libfmt to runtime.
When it was added to compiler, compilation time of kphp2cpp increased in several times.
I don't want to increase compilation time and pch readers for all vkcom.

drdzyk:
> I am strictly against adding libfmt to runtime. When it was added to compiler, compilation time of kphp2cpp increased in several times. I don't want to increase compilation time and pch readers for all vkcom.

It's not participate in compilation, but needed only for linkage(fmt.h included only in runtime/string.cpp, thus it's not propagated at runtime-headers.pch). So I think that compilation time won't change.

PR #654, opened 2022-10-20
add ctype family of functions

drdzyk:
This PR adds [ctype](https://www.php.net/manual/en/book.ctype.php) family of functions.


builtin-functions/_functions.txt
```diff
@@ -709,6 +709,20 @@ function rtrim ($s ::: string, $what ::: string = " \n\r\t\v\0") ::: string;
 function xor_strings ($s ::: string, $t ::: string) ::: string;
 function similar_text ($first ::: string, $second ::: string, float &$percent = TODO) ::: int;
 
+function extension_loaded(string $extension): bool;
```
drdzyk, line 712: This function needed for launching tests from zend repo. Namely each test for ctype_* function in zend repo has next code:
```
<?php	if (!extension_loaded('ctype')) die('skip ctype extension not available');?>
```

Ramzeeset: Let's write declaration in common style like functions above
```suggestion
function extension_loaded($extension ::: string) ::: bool;
```

builtin-functions/_functions.txt
```diff
@@ -709,6 +709,20 @@ function rtrim ($s ::: string, $what ::: string = " \n\r\t\v\0") ::: string;
...
+
+function ctype_alnum(mixed $text): bool;
```
Ramzeeset, line 714: Same for these functions

drdzyk: why?

Ramzeeset (approved):

PR #659, opened 2022-10-27
implement bcsqrt function

drdzyk:
The [bcsqrt()](https://www.php.net/manual/en/function.bcsqrt.php) api function added.

### DIFF
--- a/tests/phpt/dl/483_bcmath.php
+++ b/tests/phpt/dl/483_bcmath.php
@@ -160,4 +160,73 @@ function test_bcmath_negative_zero() {
   #endif
 }
 
+function test_bcsqrt($nums) {
+  foreach ($nums as $num) {
+    echo bcsqrt($num), "\n";
+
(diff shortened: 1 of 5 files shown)


tests/phpt/dl/483_bcmath.php
```diff
@@ -160,4 +160,73 @@ function test_bcmath_negative_zero() {
   #endif
 }
 
+function test_bcsqrt($nums) {
+  foreach ($nums as $num) {
+    echo bcsqrt($num), "\n";
+
+    for ($i = 0; $i < 30; ++$i) {
+      echo "num: ", $num, "; scale: ", $i, "; sqrt: ", bcsqrt($num, $i), "\n";
```
DrDet, line 168: nit: php variables in double quoted strings are substituted with their values
```suggestion
      echo "num: $num; scale: $i; sqrt: ", bcsqrt($num, $i), "\n";
```

DrDet (dismissed):

PR #667, opened 2022-11-02
[runtime] add the ability to specify json config to use inside the script

Ramzeeset:
# Add the ability to specify json config to use inside the script

specifying
--------
It is now possible to specify a special json congifuration when starting the server, which will be available from runtime
```
./server --runtime-config /path/to/config.json
```

Usage
--------
```php
...
$config = kphp_runtime_config();
echo $config["tmp"];
...
```

Some important things
------------------------------

* Changing the received config is valid only until the end of the current script and does not affect subsequent scripts and other workers
* If no config was specified at start, the function will return null


server/php-engine.cpp
```diff
@@ -2245,6 +2263,7 @@ void parse_main_args(int argc, char *argv[]) {
                                                                                              "memory limit = per_process_memory * processes_count");
   parse_option("job-workers-shared-messages-process-multiplier", required_argument, 2031, "Coefficient used to calculate the total count of the shared messages for job workers related communication:\n"
                                                                                           "messages count = coefficient * processes_count");
+  parse_option("runtime-congifuration", required_argument, 2032, "Smart text"); // add option for runtime
```
quasilyte, line 2266: What's "smart text"?

Ramzeeset: oyyyyyy

Ramzeeset: fix

DrDet (commented):
Let's add python tests for the following cases:
- read only normal usage
- write usage to check that copy-on-write to script memory is working correctly

server/php-engine.cpp
```diff
@@ -2143,6 +2145,22 @@ int main_args_handler(int i, const char *long_option) {
...
+        return -1;
+      }
+      std::stringstream stringstream;
+      stringstream << file.rdbuf();
+      auto [config, success] = json_decode(string(stringstream.str().c_str()));
+      if (!success) {
+        kprintf("--%s option : file is not json\n", long_option);
+        return -1;
+      }
+      runtime_config = config;
```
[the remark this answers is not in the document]
Ramzeeset: fix

server/php-engine.cpp
```diff
@@ -2245,6 +2263,7 @@ void parse_main_args(int argc, char *argv[]) {
                                                                                              "memory limit = per_process_memory * processes_count");
   parse_option("job-workers-shared-messages-process-multiplier", required_argument, 2031, "Coefficient used to calculate the total count of the shared messages for job workers related communication:\n"
                                                                                           "messages count = coefficient * processes_count");
+  parse_option("runtime-configuration", required_argument, 2032, "name of the json file that will be available at runtime");
```
DrDet, line 2266:
```suggestion
  parse_option("runtime-config", required_argument, 2032, "JSON file path that will be available at runtime as 'mixed' via 'kphp_runtime_config()'");
```

tests/python/tests/http_server/test_runtime_config.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+    def test_read_only(self):
+        response = self.kphp_server.http_request(uri="/test_runtime_config?mode=read_only")
+        self.assertEqual(response.content, b'name : simple configuration version : 1.0.0 OK')
+
+    def test_modifying(self):
+        response = self.kphp_server.http_request(uri="/test_runtime_config?mode=modify")
+        print(response.content)
+        self.assertEqual(response.content, b'modified version : 1.0.1 old version : 1.0.0 OK')
```
DrDet, line 19: Let's assert that response code is 200 OK

tests/python/tests/http_server/php/data/runtime_configuration.json
```diff
@@ -0,0 +1,4 @@
+{
+  "name": "simple configuration",
+  "version": "1.0.0"
+}
```
DrDet, line 4: nit: new line, pls

tests/python/tests/http_server/test_runtime_config.py
```diff
@@ -0,0 +1,19 @@
...
+        })
+
+
+    def test_read_only(self):
+        response = self.kphp_server.http_request(uri="/test_runtime_config?mode=read_only")
+        self.assertEqual(response.content, b'name : simple configuration version : 1.0.0 OK')
+
+    def test_modifying(self):
+        response = self.kphp_server.http_request(uri="/test_runtime_config?mode=modify")
+        print(response.content)
```
DrDet, line 18: redundant print

server/php-engine.cpp
Ramzeeset: fix

tests/python/tests/http_server/test_runtime_config.py
Ramzeeset: fix

DrDet (dismissed):

PR #668, opened 2022-11-05
Fix unexpected memory piece error

DrDet:
Most likely, the key reason of the error - _malloc_ replacement isn't rolled back correctly sometimes.
In production it happens in 2 cases: at `qmem_clear` function on runtime environment freeing and at `rpcx_execute` on receiving   
RPC response. 
Now we force _malloc_ rollback in this 2 cases to fix the error itself and then report the critical error with the stacktrace of last _malloc_ replacement to be able to investigate it more precisely.

relates to `KPHP-1511`

### DIFF
--- a/runtime/allocator.cpp
+++ b/runtime/allocator.cpp
@@ -247,7 +247,6 @@ public:
 
   void replace_malloc(bool is_malloc_replaced_before) noexcept {
     php_assert(is_malloc_replaced_before == is_malloc_replaced_);
-    CriticalSectionGuard critical_section;
 
     if (!is_malloc_replaced_before) {
       last_malloc_replacement_backtrace_size_ = fast_backtrace(last_malloc_replacement_backtrace_.data(), last_malloc_replacement_backtrace_.size());
--- a/runtime/allocator.cpp
+++ b/runtime/allocator.cpp
@@ -11,6 +11,7 @@
 
 #include "common/algorithms/find.h"
 #include "common/containers/final_action.h"
+#include "common/fast-backtrace.h"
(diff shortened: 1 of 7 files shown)


runtime/allocator.cpp
```diff
@@ -244,16 +247,28 @@ class MallocStateHolder : vk::not_copyable {
 
   void replace_malloc(bool is_malloc_replaced_before) noexcept {
     php_assert(is_malloc_replaced_before == is_malloc_replaced_);
+    CriticalSectionGuard critical_section;
```
drdzyk, line 250: move it under `if`?

runtime/kphp-backtrace.cpp
```diff
@@ -14,7 +14,7 @@
 
 std::forward_list<char **> KphpBacktrace::last_used_symbols_;
 
-KphpBacktrace::KphpBacktrace(void **raw_backtrace, int32_t size) noexcept {
+KphpBacktrace::KphpBacktrace(void *const *raw_backtrace, int32_t size) noexcept {
```
drdzyk, line 17: looks a bit ugly, maybe revert it? up to you though.

DrDet: Without it I can't make `MallocStateHolder::get_last_malloc_replacement_backtrace()` const.
Because it doesn't fit [`backtrace_symbols`](https://linux.die.net/man/3/backtrace_symbols) const signature

runtime/allocator.cpp
DrDet: removed

drdzyk (approved):

PR #669, opened 2022-11-07
eliminate deprecated malloc hooks

drdzyk:
Get rid of deprecated [malloc hooks](https://man7.org/linux/man-pages/man3/malloc_hook.3.html). These hooks is removed from further versions of glibc making compilation of kphp unsuccessful under e.g. ubuntu 22.

### DIFF
--- a/runtime/allocator.cpp
+++ b/runtime/allocator.cpp
@@ -287,7 +287,7 @@ MemoryReplacementGuard::~MemoryReplacementGuard() {
 
 extern "C" void *__libc_malloc(size_t size);
 extern "C" void *__libc_calloc(size_t nmemb, size_t size);
-extern "C" void *__libc_realloc(void *mem, std::size_t new_size);
+extern "C" void *__libc_realloc(void *mem, size_t new_size);
 extern "C" void __libc_free(void *mem);
 extern "C" void *__libc_memalign(size_t alignment, size_t size);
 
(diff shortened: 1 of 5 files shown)


runtime/allocator.cpp
```diff
@@ -306,6 +282,48 @@ MemoryReplacementGuard::~MemoryReplacementGuard() {
 
 } // namespace dl
 
+// sanitizers aren't happy with custom realization of malloc-like functions
+#if !ASAN_ENABLED and !defined(__APPLE__)
+
+extern "C" void *__libc_malloc(size_t size);
+extern "C" void *__libc_calloc(size_t nmemb, size_t size);
+extern "C" void *__libc_realloc(void *mem, std::size_t new_size);
```
DrDet, line 290: Let's make it consistent with other declarations:
```suggestion
extern "C" void *__libc_realloc(void *mem, size_t new_size);
```
Or maybe it's better to add ’std::’ in other declarations

runtime/allocator.cpp
```diff
@@ -306,6 +282,48 @@ MemoryReplacementGuard::~MemoryReplacementGuard() {
...
+extern "C" void __libc_free(void *mem);
+extern "C" void *__libc_memalign(size_t alignment, size_t size);
+
+void *malloc(size_t size) {
+  return dl::is_malloc_replaced() ? dl::script_allocator_malloc(size) : __libc_malloc(size);
+}
+
+void *calloc(size_t nmemb, size_t size) {
+  return dl::is_malloc_replaced() ? dl::script_allocator_calloc(nmemb, size) : __libc_calloc(nmemb, size);
+}
```
DrDet, line 300: We didn't replace calloc before, but now it's replaced. Is this intentional?

runtime/allocator.cpp
drdzyk: true, prefix removed

DrDet (approved):

runtime/allocator.cpp
drdzyk: Yes it is. `calloc` under the hood uses `__malloc_hook`, substituted in our previous implementation. Now we ought to override `calloc` if we want to preserve the same behaviour.

PR #670, opened 2022-11-08
replaced sprintf to snprintf

troy4eg:


common/algorithms/simd-int-to-string-test.cpp
```diff
@@ -11,23 +11,27 @@
 
 namespace {
 std::string int32_to_string(int32_t x) {
-  char buffer[11];
-  return std::string{buffer, simd_int32_to_string(x, buffer)};
+  const size_t buffer_size = 11;
+  char buffer[buffer_size];
```
drdzyk, line 15: please, use std::array<char, 11> in c++ code, or even std::array<std::byte, 11> having c++17

drdzyk: same for code below

common/algorithms/simd-int-to-string-test.cpp
```diff
@@ -11,23 +11,27 @@
...
+  return std::string{buffer, simd_int32_to_string(x, buffer, buffer_size)};
```
drdzyk, line 16: please, just `return {buffer, simd_int32_to_string(x, buffer, buffer_size)};`,
or given in mind previous comment `return {buffer, simd_int32_to_string(x, buffer, buffer.size())};`

drdzyk: same for code below

common/algorithms/simd-int-to-string.cpp
```diff
@@ -183,26 +183,26 @@ char *simd_uint32_to_string(uint32_t value, char *out_buffer) noexcept {
...
   }
-  if (value < 100000000u) {
+  if (value < 100000000U) {
     return impl_::uint32_greater10000_less100000000_to_string(static_cast<uint32_t>(value), out_buffer);
   }
 
-  if (value < 10000000000000000ull) {
-    const auto v0 = static_cast<uint32_t>(value / 100000000u);
-    const auto v1 = static_cast<uint32_t>(value % 100000000u);
+  if (value < 10000000000000000ULL) {
```
drdzyk, line 203: I'm not sure about tons of such changes. It spoils git history and not easy for review while profit is dubious.
I know clang-tidy highlighted such code, how about to tweak .clang-tidy to just suppress such alert?

net/net-memcache-server.cpp
```diff
@@ -756,10 +758,10 @@ int return_one_key_list (struct connection *c, const char *key, int key_len, int
   ptr += 14;
   if (res != 0x7fffffff) {
     if (mode >= 0) {
-      w = sprintf (ptr, "%d", res);
+      w = snprintf(ptr, ptr_size - 14, "%d", res);
```
DrDet, line 761: why `ptr_size - 14`?
It's actually 498 when "%d" is maximum 12 bytes including null terminated symbol. Does it make sense?

net/net-memcache-server.cpp
```diff
@@ -771,64 +773,69 @@ int return_one_key_list (struct connection *c, const char *key, int key_len, int
...
+        return -1;
+      }
       s = ptr + 480;
     }
     if (mode >= 0) {
       if (i || res != 0x7fffffff) {
         *ptr++ = ',';  w++;
       }
-      w += t = sprintf (ptr, "%d", R[i]);
+      w += t = snprintf(ptr, 10, "%d", R[i]);
```
DrDet, line 785: 10 is definitely not enough
"%d" is 10 digits + potential minus symbol + null terminated byte, in total 12 bytes at least.

net/net-memcache-server.cpp
```diff
@@ -771,64 +773,69 @@ int return_one_key_list (struct connection *c, const char *key, int key_len, int
...
     } else {
       w += t = 4;
-      *((int *) ptr) = R[i];
+      *(reinterpret_cast<int *>(ptr)) = R[i];
     }
     ptr += t;
   }
-  size_ptr[sprintf (size_ptr, "% 9d", w)] = '\r';
-  memcpy (ptr, "\r\n", 2);
+  size_ptr[snprintf(size_ptr, 12, "% 9d", w)] = '\r';
```
DrDet, line 792: Are you sure this is enough taking alignment into the account?

net/net-memcache-server.cpp
```diff
@@ -771,64 +773,69 @@ int return_one_key_list (struct connection *c, const char *key, int key_len, int
...
 
-  memcpy (ptr, " 0 .........\r\n", 14);
+  memcpy(ptr, " 0 .........\r\n", 14);
   char *size_ptr = ptr + 3;
 
   ptr += 14;
   if (res != 0x7fffffff) {
     if (mode >= 0) {
-      w = sprintf (ptr, "%d", res);
+      w = snprintf(ptr, 10, "%d", res);
```
DrDet, line 835: The same here, 10 is not enough

DrDet, line 835: The same here, 10 is not enough

net/net-memcache-server.cpp
```diff
@@ -838,31 +845,33 @@ int return_one_key_list_long (struct connection *c, const char *key, int key_len
...
       s = ptr + 480;
     }
     if (mode >= 0) {
       if (i || res != 0x7fffffff) {
-        *ptr++ = ',';  w++;
+        *ptr++ = ',';
+        w++;
       }
-      w += t = sprintf (ptr, "%lld", R[i]);
+      w += t = snprintf(ptr, 14, "%lld", R[i]);
```
DrDet, line 860: Why 14 and not at least 21?

net/net-memcache-server.cpp
```diff
@@ -838,31 +845,33 @@ int return_one_key_list_long (struct connection *c, const char *key, int key_len
...
     } else {
       w += t = 8;
-      *((long long *) ptr) = R[i];
+      *(reinterpret_cast<long long *>(ptr)) = R[i];
     }
     ptr += t;
   }
-  size_ptr[sprintf (size_ptr, "% 9d", w)] = '\r';
-  memcpy (ptr, "\r\n", 2);
+  size_ptr[snprintf(size_ptr, 12, "% 9d", w)] = '\r';
```
DrDet, line 867: same

runtime/misc.cpp
```diff
@@ -445,7 +446,7 @@ mixed f$getimagesize(const string &name) {
   result.push_back(width);
   result.push_back(height);
   result.push_back(type);
-  int len = sprintf((char *)buf, "width=\"%d\" height=\"%d\"", width, height);
+  int len = snprintf((char *)buf, min_size,"width=\"%d\" height=\"%d\"", width, height);
```
DrDet, line 449:
```suggestion
  int len = snprintf(reinterpret_cast<char *>(buf), min_size, "width=\"%d\" height=\"%d\"", width, height);
```

runtime/string.inl
```diff
@@ -218,10 +218,10 @@ string::string(int64_t i) {
   const char *end = nullptr;
   if (static_cast<int64_t>(i32) == i) {
     p = create(STRLEN_INT32, true);
-    end = simd_int32_to_string(i32, p);
+    end = simd_int32_to_string(i32, p, STRLEN_INT32);
```
DrDet, line 221: Terminated null byte must be included in that length
```suggestion
    end = simd_int32_to_string(i32, p, STRLEN_INT32 + 1);
```

runtime/string.inl
```diff
@@ -218,10 +218,10 @@ string::string(int64_t i) {
...
   } else {
     p = create(STRLEN_INT64, true);
-    end = simd_int64_to_string(i, p);
+    end = simd_int64_to_string(i, p, STRLEN_INT64);
```
DrDet, line 224: The same
```suggestion
    end = simd_int64_to_string(i, p, STRLEN_INT64 + 1);
```

runtime/string.inl
```diff
@@ -444,9 +444,8 @@ string &string::append(int64_t i) {
   }
 
   reserve_at_least(size() + STRLEN_INT64);
-  const char *end = simd_int64_to_string(i, p + size());
+  const char *end = simd_int64_to_string(i, p + size(), STRLEN_INT64);
```
DrDet, line 447: same

runtime/string_buffer.inl
```diff
@@ -92,19 +92,19 @@ string_buffer &operator<<(string_buffer &sb, bool x) {
 
 string_buffer &operator<<(string_buffer &sb, int32_t x) {
   sb.reserve_at_least(11);
-  sb.buffer_end = simd_int32_to_string(x, sb.buffer_end);
+  sb.buffer_end = simd_int32_to_string(x, sb.buffer_end, 11);
```
DrDet, line 95: 12, null terminated must be included

runtime/string_buffer.inl
```diff
@@ -92,19 +92,19 @@ string_buffer &operator<<(string_buffer &sb, bool x) {
...
   return sb;
 }
 
 string_buffer &operator<<(string_buffer &sb, uint32_t x) {
   sb.reserve_at_least(10);
-  sb.buffer_end = simd_uint32_to_string(x, sb.buffer_end);
+  sb.buffer_end = simd_uint32_to_string(x, sb.buffer_end, 10);
```
DrDet, line 101: 11

runtime/string_buffer.inl
```diff
@@ -92,19 +92,19 @@ string_buffer &operator<<(string_buffer &sb, bool x) {
...
   return sb;
 }
 
 string_buffer &operator<<(string_buffer &sb, int64_t x) {
   sb.reserve_at_least(20);
-  sb.buffer_end = simd_int64_to_string(x, sb.buffer_end);
+  sb.buffer_end = simd_int64_to_string(x, sb.buffer_end, 20);
```
DrDet, line 107: 21

DrDet, line 107: 21

vkext/vkext.cpp
```diff
@@ -255,14 +255,14 @@ void write_buff_long(long x) {
...
-  wptr += sprintf(wptr, "%ld", x);
+  wptr += snprintf(wptr, 25, "%ld", x);
 }
 
 void write_buff_double(double x) {
   if (unlikely (wptr + 100 > buff + BUFF_LEN)) {
     flush_buff();
   }
-  wptr += sprintf(wptr, "%lf", x);
+  wptr += snprintf(wptr, 100, "%f", x);
```
DrDet, line 265: why not "%lf" as before?
```suggestion
  wptr += snprintf(wptr, 100, "%lf", x);
```

troy4eg: Because php has **snprintf.h** with 

```
#ifdef snprintf
#undef snprintf
#endif
#define snprintf ap_php_snprintf
```

If we use "lf" it will throw an exception:
> PHP Fatal error:  Illegal length modifier specified 'f' in s[np]printf call

I found some ~~legacy~~ info about that: https://externals.io/message/14592

But we can use just "f" and it will be work fine, because when arguments of type float are passed as variadic parameters, such arguments are implicitly converted to type double.

runtime/string.inl
troy4eg: I think i found a better solution
```suggestion
    end = simd_int32_to_string(i32, p, inner()->capacity);
```

runtime/string.inl
troy4eg:
```suggestion
    end = simd_int64_to_string(i, p, inner()->capacity);
```

net/net-http-server.cpp
```diff
@@ -834,28 +832,28 @@ int write_basic_http_header (struct connection *c, int code, int date, int len,
...
+		     content_type ? content_type : "text/html",
+		     (D->query_flags & QF_KEEPALIVE) ? "keep-alive" : "close",
+		     (D->query_flags & QF_EXTRA_HEADERS) && extra_http_response_headers ? extra_http_response_headers : "",
 		     add_header ?: "");
     D->query_flags &= ~QF_EXTRA_HEADERS;
-    assert (ptr < buff + B_SZ - 64);
+    assert(ptr < buff + buff_size - 64);
     if (len >= 0) {
-      ptr += sprintf (ptr, "Content-Length: %d\r\n", len);
+      ptr += snprintf(ptr, 29, "Content-Length: %d\r\n", len);
```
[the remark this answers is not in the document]
DrDet: 31 at least, because there can be 10 digits without sign

common/algorithms/simd-int-to-string-test.cpp
troy4eg: done

common/algorithms/simd-int-to-string.cpp
troy4eg: done

common/algorithms/simd-int-to-string-test.cpp
troy4eg: done

runtime/string_buffer.inl
troy4eg: done

runtime/string_buffer.inl
troy4eg: done

runtime/string_buffer.inl
troy4eg: done

DrDet (approved):

PR #679, opened 2022-11-17
[runtime] shutdown functions refactoring

Ramzeeset:
# shutdown functions refactoring

What's the problem
-------------------------
At this time shutdown functions are not called if we get an error in the network context. For example, a timeout or connection interruption does not cause the shutdown functions to be called.

Solution
-----------

Let's add function `check_net_error()` that will be called after pause and check for networking errors on the script resume to invoke shutdown_functions.

Also let's resume script in `finish()` if we get network error 

Additions
------------

1. Fix bug of second shutdown function call
```php
<?php

/**
 * @kphp-required
 */
function at_finish0()
{
    fwrite(STDERR, "FINISHED 0\n");
}

/**
 * @kphp-required
 */
function at_finish1()
{
    fwrite(STDERR, "FINISHED 1\n");
    exit(0);
}

register_shutdown_function('at_finish0');
register_shutdown_function('at_finish1');
throw new Exception("example", 1);

```
2. Correct function 'check_tl()' so that it calls the shutdown function


DrDet (changes requested):
The problem is shutdown functions can send/receive net queries and swap execution to network context and then back. So far, it seems that it won't handle this behavior correctly. We've discussed to test it more carefully before merging

PR #682, opened 2022-11-18
WIP: Fix malloc replacement error

DrDet:


runtime/datetime/timelib_wrapper.cpp
```diff
@@ -216,7 +216,7 @@ std::pair<int64_t, bool> php_timelib_strtotime(const string &tz_name, const stri
   }
 
   bool use_heap_memory = (dl::get_script_memory_stats().memory_limit == 0);
-  auto malloc_replacement_guard = make_malloc_replacement_with_script_allocator(!use_heap_memory);
+  auto malloc_replacement_guard = make_malloc_replacement_with_script_allocator_guard(!use_heap_memory);
```
DrDet, line 219: @quasilyte Can you check this bunch of code please?
There is your comment several lines below saying we can't use critical section here, but now we have to use it.

drdzyk (commented):
We discussed offline that it would be nice to use general 'rollback malloc' code in case of error|timeouts|etc. This allows to avoid putting possibly huge amount of code which is under malloc guard in blocking signal state

DrDet:
> We discussed offline that it would be nice to use general 'rollback malloc' code in case of error|timeouts|etc. This allows to avoid putting possibly huge amount of code which is under malloc guard in blocking signal state

done

drdzyk (approved):

PR #686, opened 2022-11-22
[php8] Nullsafe operator implementation

mkornaukhov03:
Main issue: https://github.com/VKCOM/kphp/issues/290
This patch implements nullsafe operator `?->` for methods and properties.
RFC: https://wiki.php.net/rfc/nullsafe_operator


tests/phpt/php8/001_nullsafe_op_simple.php
```diff
@@ -0,0 +1,171 @@
+@ok php8
+
+require_once 'kphp_tester_include.php';
```
i582, line 3: Seems `require_once` should be after `<?php`.

tests/phpt/php8/002_nullsafe_op_special_cases.php
```diff
@@ -0,0 +1,63 @@
+@ok php8
+
+require_once 'kphp_tester_include.php';
```
i582, line 3: And here

compiler/pipes/gen-tree-postprocess.cpp
```diff
@@ -230,13 +232,86 @@ VertexPtr GenTreePostprocessPass::on_enter_vertex(VertexPtr root) {
 }
 
 VertexPtr GenTreePostprocessPass::on_exit_vertex(VertexPtr root) {
+  if (auto as_instance_prop = root.try_as<op_instance_prop>(); as_instance_prop && as_instance_prop->is_null_safe) {
+    if (as_instance_prop->instance().try_as<op_func_call>() || as_instance_prop->instance().try_as<op_seq_rval>() ||
```
quasilyte, line 236: This condition looks complicated and it may require a comment.
It also looks like it's not complete, it covers some of the cases but not the others.
To make it more extensible, maybe move it to a method that would report whether the transformation inside this if body should happen?

tests/phpt/php8/001_nullsafe_op_simple.php
```diff
@@ -0,0 +1,172 @@
+@ok php8
+
```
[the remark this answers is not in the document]
mkornaukhov03: Fixed

compiler/gentree.cpp
```diff
@@ -332,12 +333,12 @@ VertexPtr GenTree::get_postfix_expression(VertexPtr res, bool parenthesized) {
       }
       res.set_location(location);
       need = true;
-    } else if (tp == tok_arrow) {
+    } else if (tp == tok_arrow || tp == tok_nullsafe_arrow) {
```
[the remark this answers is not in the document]
mkornaukhov03: Yes, this one looks better
Fixed

compiler/pipes/gen-tree-postprocess.cpp
mkornaukhov03: I simplified it.
I tried not to create temporary variable if it was possible, in vain. Now it looks more general and obvious.

PR #690, opened 2022-11-28
Added parse_ini_string, parse_ini_file functions

pmswga:
At the moment, the functions do not process an array of values.

**Example:**

[Section3.1]
var1[]=val_1_1
var1[]=val_1_2
var1[]=val_1_3

This section will skip.


runtime/parsing_functions.cpp
```diff
@@ -0,0 +1,325 @@
...
+      while (str[i+1] == ' ') {
+        i++;
+      }
+    }
+  }
+
+  return clear_str;
+}
+
+string tolower(const string &str)
```
quasilyte, line 69: I would suggest using `f$strtolower()` function instead.

runtime/parsing_functions.cpp
```diff
@@ -0,0 +1,325 @@
...
+  string lower_str = {};
+
+  for (string::size_type i = 0; i < str.size(); i++) {
+    lower_str.push_back(tolower(str[i]));
+  }
+
+  return lower_str;
+}
+
+string trim(const string &str)
```
quasilyte, line 84: I would suggest using `f$trim` function instead or any other equivalent.
It's better not to duplicate the functions with the same semantics.

runtime/parsing_functions.cpp
```diff
@@ -0,0 +1,325 @@
...
+  }
+
+  return str.substr(s, (e-s)+1);
+}
+
+/*
+ * INI parsing functions
+ */
+
+bool isINISection(const string &ini_section)
```
quasilyte, line 108: These ini-related functions should not be contained inside a file with generic "parsing functions" name.
I would prefer them to be inside the `cpp` file that implements the ini library functionality.

tests/cpp/runtime/parsing-functions-test.cpp
```diff
@@ -0,0 +1,218 @@
...
+
+TEST(parsing_functions_test, test_clear_extra_spaces) {
+  ASSERT_STREQ(
+    string("This is string without extra spaces").c_str(),
+    clearExtraSpaces(string("This           is     string   without     extra spaces")).c_str()
+  );
+}
+
+
+TEST(parsing_functions_test, test_tolower) {
```
quasilyte, line 29: This test will be redundant if you'll use already available `f$strtolower` function.

quasilyte (changes requested):
This is a first review round. I'm sure @DrDet will find something else here, but let's start with something simple.

runtime/parsing_functions.cpp
```diff
@@ -0,0 +1,325 @@
...
+{
+  return std::regex_match(ini_section.c_str(), std::regex("^\\[.+\\]$", std::regex::ECMAScript));
+}
+
+bool isINIVar(const string &ini_var)
```
DrDet, line 113: We use snake_case naming at runtime

runtime/parsing_functions.h
```diff
@@ -0,0 +1,43 @@
+#pragma once
+
+#include "runtime/kphp_core.h"
+#include <regex>
+#include <iostream>
+#include <fstream>
+#include <string>
```
DrDet, line 7: In most cases it's unsafe to use std utilities in runtime until you 100% sure you're doing.

The problem is kphp runtime can always be interrupted by signal (script timeout for example), which leads to `swapcontext` back to network context and script termination WITHOUT calling any destructors. It will cause memory leaks (because std often uses heap memory) and a lot of other problems.
In your case using any of these dependencies is wrong and unsafe.
See other comments for more detailed explanation for each usage.

runtime/parsing_functions.cpp
```diff
@@ -0,0 +1,325 @@
...
+    }
+  }
+
+  if (!str.empty()) {
+    ini_string.append(str);
+  }
+
+  ifs.close();
+
+  return f$parse_ini_string(string(ini_string.c_str()), process_sections, scanner_mode);
```
DrDet, line 324: Using `std::string` and `std::ifstream` this way is unsafe as I explained above. But here you don't need it, just use `f$file_get_contents` to read the file

runtime/parsing_functions.cpp
```diff
@@ -0,0 +1,325 @@
...
+}
+
+bool isINIFloatVal(const string &ini_val)
+{
+  return std::regex_match(ini_val.c_str(), std::regex(R"(^\d+\.\d*$)", std::regex::ECMAScript|std::regex::icase));
+}
+
+bool isINIStrVal(const string &ini_val)
+{
+  return std::regex_match(ini_val.c_str(), std::regex("^.*$", std::regex::ECMAScript|std::regex::icase));
```
DrDet, line 135: Using std::regex is unsafe here, use `f$preg_match` instead of it

runtime/parsing_functions.h
```diff
@@ -0,0 +1,43 @@
...
+const int INI_SCANNER_RAW = 1;
+const int INI_SCANNER_TYPED = 2;
+
+string get_ini_section(const string &ini_entry);
+
+string get_ini_var(const string &ini_entry);
+
+string get_ini_val(const string &ini_entry);
+
+bool cast_str_to_bool(const string &ini_var);
```
DrDet, line 39: You don't need to expose this functions to header, because you use them only in one translation unit. They should be private for your `parsing_functions.cpp` file

DrDet (changes requested):
In general, I'm not sure it's correct way of parsing ini string. For example, PHP use autogenerated parser for that. Can you explain your parsing method somehow?

runtime/parsing_functions.h
pmswga: Okay, I am understand. But, how I will be test private functions?

pmswga:
> general, I'm not sure it's correct way of parsing ini string. For example, PHP use autogenerated parser for that. Can you explain your parsing method somehow?

I didn't use the automatically generated parsers, because:
1) I didn't think about it
2) I'm not sure that I could correctly implement it in runtime

Therefore, the solution that came to my mind is to check the resulting strings by regular expressions.

runtime/parsing_functions.cpp
```diff
@@ -0,0 +1,325 @@
...
+
+string clearExtraSpaces(const string &str)
+{
+  if (str.empty()) {
+    return {};
+  }
+
+  string clear_str;
+
+  for (string::size_type i = 0; i < str.size(); i++) {
```
Tsygankov-Slava, line 54: When you use postfix increment/decrement (`i++` / `i--`) you make an additional copy. Therefore, it's recommended to use prefix increment/decrement (`++i` / `--i`).

P.S. This comment applies to the entire code.

Tsygankov-Slava (commented):
I have found several places where something can be improved. It would be great if you took these comments into attention.

runtime/ini.cpp
```diff
@@ -0,0 +1,240 @@
...
+
+  return res;
+}
+
+bool cast_str_to_bool(const string &ini_var) {
+  if (ini_var.empty()) {
+    return false;
+  }
+
+  string ini_bool_var = f$strtolower(ini_var);
```
Tsygankov-Slava, line 109: It's better to declare variables that don't change as `const`.

```c++
const string ini_bool_var = f$strtolower(ini_var);
```

runtime/ini.cpp
```diff
@@ -0,0 +1,240 @@
...
+
+  if (process_sections) {
+    if (!ini_section.empty()) {
+      res.set_value(ini_section, section);
+    }
+  } else {
+    return section;
+  }
+
+  return res;
```
Tsygankov-Slava, line 222: In this case, it's better to use `Return Early Pattern`:
```c++
if (!process_sections) {
    return section;
} 
if (!ini_section.empty()) {
    res.set_value(ini_section, section);
}
return res;
```

runtime/ini.cpp
```diff
@@ -0,0 +1,240 @@
...
+
+
+
+array<mixed> f$parse_ini_file(const string &filename, bool process_sections, int scanner_mode) {
+  if (filename.empty()) {
+    php_warning("Filename cannot be empty");
+    return {};
+  }
+
+  Optional<string> ini_string = f$file_get_contents(filename);
```
Tsygankov-Slava, line 233: Also it's better to use `const` and `auto`:
```c++
const auto ini_string = f$file_get_contents(filename);
```

runtime/ini.cpp
```diff
@@ -0,0 +1,240 @@
...
+
+  string ini_string_copy = f$trim(ini_string);
+  ini_string_copy = clear_extra_spaces(ini_string_copy);
+
+  array<mixed> res(array_size(0, 0, true));
+  array<mixed> section(array_size(0, 0, true));
+  string ini_entry;
+  string ini_section;
+
+  for (string::size_type i = 0; i <= ini_string_copy.size(); ++i) {
```
Tsygankov-Slava, line 135: Firstly, also it's better to use `auto` here.
Secondly, it looks very strange that your `for` goes to `ini_string_copy.size()`, suddenly there will be a situation that you turn to `ini_string_copy[ini_string_copy.size()])` (for example, if a string of the form `[text` somehow gets to your function). It's not very good.
It's better to fix on 
```c++
for (auto i = 0; i < ini_string_copy.size(); ++i) {
...
}
```

runtime/ini.cpp
```diff
@@ -0,0 +1,240 @@
...
+      res.set_value(ini_section, section);
+    }
+  } else {
+    return section;
+  }
+
+  return res;
+}
+
+
```
Tsygankov-Slava, line 225: Extra lines.

runtime/ini.cpp
mkornaukhov03: Type of `i` is deduced as `int`, it is not an intention.

Tsygankov-Slava: > Type of `i` is deduced as `int`, it is not an intention.

Thanks, you're right.
Fixed

PR #704, opened 2022-12-13
relogin refactoring

troy4eg:
* removed default user kitten
* user will not be changed if he is already root


DrDet (changes requested):
GitHub actions pipeline is failing now for focal on dummy php script step. Also I remember there're some caveats with user at python tests and mysql especially . Check this out please

DrDet (approved):

Issue #639, opened 2022-10-05
Support for parsing yaml files.


Danil42Russia:
I would like to see native support for parsing yaml files.

As it is already done in PHP: https://www.php.net/manual/en/ref.yaml.php
To work with the files, you can use the attached library: **yaml-cpp**

PR #707, opened 2022-12-18
Add yaml parsing & emitting support

NikOsint:
# KPHP YAML

Solution for Issue #639 

Implementation of 4 functions from [PHP YAML](https://www.php.net/manual/en/ref.yaml.php) using [yaml-cpp](https://github.com/jbeder/yaml-cpp):

```PHP
yaml_emit_file(string $filename, mixed $data): bool
yaml_emit(mixed $data): string
yaml_parse_file(string $filename, int $pos = 0): mixed
yaml_parse(string $data, int $pos = 0): mixed
```

C++ tests are written and implemented.

There is one major issue: strings like "10", "98.13' are emitted without quotes, so then they are parsed as integers and doubles respectively.

More informative README can be found [here](https://github.com/NikOsint/KPHP-YAML).


runtime/yaml.cpp
```diff
@@ -0,0 +1,169 @@
...
+  }
+  if (node.IsSequence()) {
+    for (auto it = node.begin(); it != node.end(); ++it) {
+      mixed data_piece;
+      yaml_node_to_mixed(*it, data_piece);
+      data.push_back(data_piece);
+    }
+  }
+  else if (node.IsMap()) {
+    for (auto it : node) {
```
Tsygankov-Slava, line 40: It's better to indicate here that `it` is constant.
And it's also better to take the value by reference, so as not to initialize it with a copy.
```c++
for (const auto &it : node) {
```

runtime/yaml.cpp
```diff
@@ -0,0 +1,169 @@
...
+    }
+    if (data.is_bool()) {
+      if (!data.as_bool()) {
+          node = false;
+      }
+      else {
+          node = true;
+      }
+      return;
+    }
```
Tsygankov-Slava, line 76: It seems, in this case, it's better to use the code structure given below:
```c++
if (data.is_null()) {
      php_warning("Cannot convert (mixed)null into yaml node");
} else if (data.is_string()) {
      node = std::string(data.as_string().c_str());
} else if (data.is_int()) {
      node = data.as_int();
} else if (data.is_float()) {
      node = data.as_double();
} else if (data.is_bool()) {
     node = data.as_bool();
}
return;
```

runtime/yaml.cpp
```diff
@@ -0,0 +1,169 @@
...
+      else if (data_piece.is_float()) {
+        node.push_back(data_piece.as_double());
+      }
+      else if (data_piece.is_bool()) {
+          node.push_back(data_piece.as_bool());
+      }
+      else if (data_piece.is_null()) {
+        php_warning("Data piece is null. Skipping it");
+      }
+      else php_warning("Unknown data type. Skipping it"); // maybe this is redundant?
```
Tsygankov-Slava, line 102: In this case, you shouldn't indent before `else`.
```c++
if (...) {
    // code
} else if (...) {
    // code
} else if (...) {
    // code
} else {
    // code
}
```

P.S. Below you also allow such a style of code, I didn't mention it there.

Tsygankov-Slava (commented):
I have found several places where something can be improved. It would be great if you took these comments into attention.

NikOsint:
@Tsygankov-Slava Thank you for your review! All the comments were taken into account and the corresponding parts of the code were corrected. Some other small improvements were done.

For now, the problem of emitting number strings such as "10", "5.34" without quotes remains due to poor documentation of yaml-cpp library. I will let you know when a solution is found.

NikOsint:
@Tsygankov-Slava @unserialize The problem of emitting numbers and numbers-as-strings has been solved.

Converting from mixed to a YAML document has been rewritten without use of yaml-cpp as it does not provide any way to emit strings with quotes and numbers without.

Also now booleans are parsed&emitted correctly (as booleans, not strings or integers).

More C++ tests has been written and a complex PHP test has been implemented, too.

For the PHP test to work, a 'yaml.so' extension has been added to the ```tests/python/lib/kphp_run_once.py``` file. Without it, the test would throw an error when running with PHP because PHP does not have its YAML module by default.

NikOsint:
Merged VKCOM:master into my master branch so that there are no conflicts.

runtime/yaml.cpp
```diff
@@ -0,0 +1,160 @@
...
+    } else if (data.is_int()) {
+      string_data.append(data.as_int());
+    } else if (data.is_float()) {
+      string_data.append(data.as_double());
+    } else if (data.is_bool()) {
+      string_data.append((data.as_bool()) ? "true" : "false");
+    }
+    string_data.push_back('\n');
+    return;
+  }
```
Tsygankov-Slava, line 92: As for me, it's better to collect the string first, and then call the `append` function once.
And I guess, this code will look better if you write it as follows:

```c++
  string buffer = "";
  if (!data.is_array()) {
    if (data.is_null()) {
      buffer = '~';
    } else if (data.is_string()) {
      const string& string_data_piece = data.as_string();
      if (string_data_piece.size() < 2 || (string_data_piece[0] != '"' && string_data_piece[string_data_piece.size() - 1] != '"')) {
        buffer = "\"" + string_data_piece + "\"";
      } else {
        buffer = string_data_piece;
      }
    } else if (data.is_int()) {
      buffer = data.as_int();
    } else if (data.is_float()) {
      buffer = data.as_double();
    } else if (data.is_bool()) {
      buffer = (data.as_bool()) ? "true" : "false");
    }
    string_data.append(buffer + '\n');
    return;
  }
```

runtime/yaml.cpp
```diff
@@ -0,0 +1,160 @@
...
+  const array<mixed> &data_array = data.as_array();
+  if (data_array.is_pseudo_vector()) {
+    for (const auto &it : data_array) {
+      const mixed &data_piece = it.get_value();
+      string_data.append(print_tabs(nesting_level));
+      string_data.append("- ");
+      if (data_piece.is_array()) {
+        string_data.push_back('\n');
+      }
```
Tsygankov-Slava, line 101: Here you can also first assemble a string then add:
```c++
  buffer = print_tabs(nesting_level) + "- ";
  if (data_piece.is_array()) {
    buffer += '\n';
  }
  string_data.append(buffer);
```

runtime/yaml.cpp
```diff
@@ -0,0 +1,160 @@
...
+    }
+  } else {
+    for (const auto &it : data_array) {
+      const mixed &data_piece = it.get_value();
+      string_data.append(print_tabs(nesting_level));
+      string_data.append(print_key(it.get_key()));
+      string_data.append(": ");
+      if (data_piece.is_array()) {
+        string_data.push_back('\n');
+      }
```
Tsygankov-Slava, line 112: By analogy:

```c++
  buffer = print_tabs(nesting_level) + print_key(it.get_key()) + ": ";
  if (data_piece.is_array()) {
    buffer += '\n';
  }
  string_data.append(buff);
```

Tsygankov-Slava (commented):
I looked your PR and found a few points that you might want to fix.

unserialize:
I'll try to review it next week

runtime/yaml.cpp
NikOsint: It seems that the ```string``` class has no overloads for ```string& operator=(const string& s)``` and ```string operator+(const string& left, const string& right)```

NikOsint: Also, there is no overload for ```string operator+(const string& lhs, const string& rhs)```. ```string buffer = string("something") + string("anything")``` would not compile as the ```mixed operator+(const mixed& lhs, const mixed& rhs)``` would be called and there is no ```string operator=(const mixed&)```

This can be solved by an implicit cast to ```string```, but I think this is not a good idea

tests/python/lib/kphp_run_once.py
```diff
@@ -40,7 +40,8 @@ def _get_extensions(self):
                 ("extension", "tokenizer.so"),
                 ("extension", "h3.so"),
                 ("extension", "zstd.so"),
-                ("extension", "ctype.so")
+                ("extension", "ctype.so"),
+                ("extension", "yaml.so")
```
Danil42Russia, line 44: add a comma (`,`) so that next time the diff is **1 change**, not 2

runtime/yaml.cpp
Tsygankov-Slava: Yes, I completely forgot that kphp uses it's own strings.
I'm afraid if you rewrite my code with `kphp strings`, then readability will not only improve, but even worsen 😅
Therefore, you are absolutely right, I apologize for the disinformation!

runtime/yaml.cpp
```diff
@@ -0,0 +1,167 @@
...
+    for (auto it = node.begin(); it != node.end(); ++it) {
+      mixed data_piece;
+      yaml_node_to_mixed(*it, data_piece, source);
+      data.push_back(data_piece);
+    }
+  } else if (node.IsMap()) {
+    for (const auto &it : node) {
+      mixed data_piece;
+      yaml_node_to_mixed(it.second, data_piece, source);
+      data[string(it.first.as<std::string>().c_str())] = data_piece;
```
unserialize, line 48: It's better to use `data.set_value()` instead of `[]` operator

runtime/yaml.cpp
```diff
@@ -0,0 +1,167 @@
...
+ * print tabs in quantity of nesting_level (used to print nested YAML entries)
+ */
+static string yaml_print_tabs(const uint8_t nesting_level) noexcept {
+  return string(2 * nesting_level, ' ');
+}
+
+/*
+ * print the key of a YAML map entry
+ */
+static string yaml_print_key(const mixed& data_key) noexcept {
```
unserialize, line 64: In our codebase, the spacing `const T &val` is preferred over `const T& val` (btw, sometimes you use the first one in this PR)

runtime/yaml.cpp
```diff
@@ -0,0 +1,167 @@
...
+  if (data_key.is_string()) {
+    return data_key.as_string();
+  }
+  return string(data_key.as_int()); // key can not be an array; bool and float keys are cast to int
+}
+
+/*
+ * get a YAML representation of mixed in a string variable
+ */
+static void mixed_to_string(const mixed& data, string& string_data, const uint8_t nesting_level = 0) noexcept {
```
unserialize, line 74: A comment above about spaces

runtime/yaml.cpp
```diff
@@ -0,0 +1,167 @@
...
+  string buffer;
+  if (!data.is_array()) {
+    if (data.is_null()) {
+      buffer.push_back('~');
```
unserialize, line 78: Please, add a comment, what does it mean? `~` is a syntax for `null` in yaml?

runtime/yaml.cpp
```diff
@@ -0,0 +1,167 @@
...
+    } else if (data.is_string()) {
+      const string& string_data_piece = data.as_string();
```
unserialize, line 80: Here also a space (and everywhere below)

runtime/yaml.cpp
```diff
@@ -0,0 +1,167 @@
...
+      // check if a string has quotes
+      if (string_data_piece.size() < 2 || (string_data_piece[0] != '"' && string_data_piece[string_data_piece.size() - 1] != '"')) {
```
unserialize, line 82: This condition seems very strange. Quotes are valid characters. Also, they can occur in the middle of a string. Anyway, they must be escaped

runtime/yaml.cpp
```diff
@@ -0,0 +1,167 @@
...
+        // if not, put it in quotes
+        buffer.push_back('"');
+        buffer.append(string_data_piece);
```
unserialize, line 85: What about escaping? Quotes, invisible characters, newlines, etc.?

runtime/yaml.cpp
```diff
@@ -0,0 +1,167 @@
...
+    } else if (data.is_bool()) {
+      buffer = (data.as_bool()) ? string("true") : string("false");
+    }
+    string_data.append(buffer);
+    string_data.push_back('\n');
+    return;
+  }
+  const array<mixed> &data_array = data.as_array();
+  if (data_array.empty()) {
+    string_data.append("[]\n"); // an empty array is represented as [] in YAML
```
unserialize, line 103: `return` after this?

runtime/yaml.cpp
NikOsint: Yes, in YAML tilda represents null. Added a comment about it in the code.

NikOsint:
Now special characters escaping is implemented.

The following characters are escaped:

- Line feed as ```\n```
- Backspace as ```\b```
- Horizontal tab as ```\t```
- Vertical tab as ```\v```
- Backslash as ```\\```
- Double quote as ```\"```

Single quote ```\'``` is not escaped since it does not act as string delimiter because strings are covered in double quotes.

Please tell me if there are other symbols that should be escaped but were not taken into account yet.

tests/phpt/yaml/03_yaml_escaping.php
```diff
@@ -0,0 +1,16 @@
...
+  /** @param mixed $in */
+  $in = yaml_parse_file($filename);
+  echo(serialize($in) . PHP_EOL);
+//   echo(yaml_emit($in) . PHP_EOL); // does not pass because PHP YAML does not use double quotes sometimes
+  /** @param mixed $out */
+  $out = yaml_parse(yaml_emit($in));
+  echo(serialize($out) . PHP_EOL);
+}
+
+test_yaml_escaping();
```
Danil42Russia, line 16: lost a line break

builtin-functions/_functions.txt
```diff
@@ -1652,3 +1652,8 @@ class DateTimeImmutable implements DateTimeInterface {
 }
 
 function getenv(string $varname = '', bool $local_only = false): mixed;
+
+function yaml_emit_file ($filename ::: string, $data ::: mixed) ::: bool;
```
unserialize, line 1656: `(string $filename, mixed $data)` and others, no :::, just type hints. Same for return — just ':' after ')'

runtime/yaml.cpp
```diff
@@ -0,0 +1,191 @@
+#include <yaml-cpp/yaml.h>
+
+#include "runtime/optional.h"
+#include "runtime/streams.h"
+#include "yaml.h"
```
unserialize, line 5: "runtime/yaml.h"

runtime/yaml.cpp
```diff
@@ -0,0 +1,191 @@
...
+
+/*
+ * convert YAML::Node to mixed after parsing a YAML document into YAML::Node
+ */
+static void yaml_node_to_mixed(const YAML::Node &node, mixed &data, const string &source) noexcept {
+  data.clear(); // sets data to NULL
+  if (node.IsScalar()) {
+    const string string_data(node.as<std::string>().c_str());
+    // check whether the primitive is put in quotes in the source YAML
+    const bool string_data_has_quotes = (source[node.Mark().pos] == '"' && source[node.Mark().pos + string_data.size() + 1] == '"');
```
unserialize, line 15: What about single quotes? I suppose, they also mean a string in yaml

runtime/yaml.cpp
```diff
@@ -0,0 +1,191 @@
...
+      data.set_value(string(it.first.as<std::string>().c_str()), data_piece);
+    }
+  }
+  // else node is Null or Undefined, so data is Null
+}
+
+/*
+ * print tabs in quantity of nesting_level (used to print nested YAML entries)
+ */
+static string yaml_print_tabs(const uint8_t nesting_level) noexcept {
```
unserialize, line 56: Why do you use uint8_t for a number? There is no reason not to use just `int` unless memory alignment is a case you care for

runtime/yaml.cpp
```diff
@@ -0,0 +1,191 @@
...
+  return string(data_key.as_int()); // key can not be an array; bool and float keys are cast to int
+}
+
+/*
+ * escape special characters in a string entry
+ */
+static string yaml_escape(const string &data) noexcept {
+  string escaped_data;
+  for (size_t i = 0; i < data.size(); ++i) {
+    if (data[i] == 10) { // line feed - code 10
```
unserialize, line 76: 1) extract data[i] to a separate (char) variable: you use in dozen of times
2) why do you compare char with a numeric code, why not `data[i] == '\n' and get rid of comments everywher?

runtime/yaml.cpp
```diff
@@ -0,0 +1,191 @@
...
+      buffer.append(yaml_escape(data.as_string())); // escape special characters
+      buffer.push_back('"');
+    } else if (data.is_int()) {
+      buffer.append(data.as_int());
+    } else if (data.is_float()) {
+      buffer.append(data.as_double());
+    } else if (data.is_bool()) {
+      buffer = (data.as_bool()) ? string("true") : string("false");
+    }
+    string_data.append(buffer);
```
unserialize, line 120: What for do you need `buffer`, why not to append into `string_data` directly?

runtime/yaml.cpp
```diff
@@ -0,0 +1,191 @@
...
+  string string_data("---\n"); // beginning of a YAML document
+  mixed_to_string(data, string_data);
+  string_data.append("...\n"); // ending of a YAML document
+  return string_data;
+}
+
+mixed f$yaml_parse_file(const string &filename, int pos) {
+  if (filename.empty()) {
+    php_warning("Filename cannot be empty");
+    return {};
```
unserialize, line 173: In PHP, `false` in returned on failure

runtime/runtime.cmake
```diff
@@ -49,6 +49,11 @@ prepend(KPHP_RUNTIME_PDO_PGSQL_SOURCES pdo/pgsql/
         pgsql_pdo_emulated_statement.cpp)
 endif()
 
+if (YAML)
+prepend(KPHP_RUNTIME_YAML_SOURCES /
```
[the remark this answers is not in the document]
NikOsint: I basically copied this `if` from above `if`s, there is also no identation
should I fix it there too?

runtime/yaml.cpp
NikOsint: well, I thought it is unrealistic to exceed 255 levels of nesting in a YAML document, so I decided to limit memory for variable a little bit
+ `uint8_t` is safer in terms of sign

runtime/yaml.cpp
NikOsint: `buffer` was added in prevoius iteration of review, I think it makes code more readable, for example on line 118

NikOsint:
> > There is one major issue: strings like "10", "98.13' are emitted without quotes, so then they are parsed as integers and doubles respectively.
> 
> Is this issue (from PR description) still actual? I can't see such a behaviour looking at source code

NikOsint:
> > There is one major issue: strings like "10", "98.13' are emitted without quotes, so then they are parsed as integers and doubles respectively.
> 
> Is this issue (from PR description) still actual? I can't see such a behaviour looking at source code

No, it is outdated, now everything in double quotes is emitted as string
Thanks for this remark, updated description

runtime/streams.h
```diff
@@ -6,7 +6,7 @@
 
 #include "runtime/kphp_core.h"
 
-using Stream =mixed;
+using Stream = mixed;
```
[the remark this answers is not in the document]
NikOsint: just noticed a missing space there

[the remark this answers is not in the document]
NikOsint: deleted this change

runtime/yaml.cpp
NikOsint: changed to int

runtime/runtime.cmake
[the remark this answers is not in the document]
NikOsint: it honestly looks worse with indentation inside the `if`-clause
I think that's why there is no indentation in above `if`s

Issue #702, opened 2022-12-09
Fix endless recursion at `JsonEncoder`


DrDet:
This code leads to endless recursion and SIGSEGV error now.
```php
<?php

class Base {
    /** @var Derived */
    public $derived = null;
}

class Derived extends Base {
    /** @var Base[] */
    public $bases_array = [];

    public function __construct() {
        parent::__construct();
    }
}

function demo() {
    $b = new Base;
    $d = new Derived;
    $b->derived = $d;
    $d->bases_array[] = $b;
    var_dump(JsonEncoder::encode($b));
}

demo();

```

relates to `KPHP-1531`

PR #710, opened 2022-12-21
Fix JsonEncoder vs JsonWriter bind

drdzyk:
There is a bug occurred with deep nesting objects. Namely, it occurred when JsonWriter's internal stack limit is reached faster than depth level of JsonEncoder.
It is fixed by using dynamic container as JsonWriter stack, meaning that there is no stack limit further.

### DIFF
--- a/runtime/array.inl
+++ b/runtime/array.inl
@@ -2038,6 +2038,11 @@ T array<T>::pop() {
 }
 
 template<class T>
+T &array<T>::back() {
+  return (--end()).get_value();
+}
+
+template<class T>
 T array<T>::shift() {
   if (count() == 0) {
     php_warning("Cannot use array_shift on empty array");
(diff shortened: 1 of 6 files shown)


unserialize:
Could you please mention https://github.com/VKCOM/kphp/issues/702 in description, as this PR solves it?

drdzyk:
> Could you please mention #702 in description, as this PR solves it?

Sure, I'm going to do so after merging this PR in master

DrDet:
> > Could you please mention #702 in description, as this PR solves it?
> 
> Sure, I'm going to do so after merging this PR in master

You don't have to close the issue manually after merging PR. Just write #702 somewhere in PR description and it will be closed automatically when you merge it. Also PR will be visually linked with the issue, like this: 
image

DrDet (approved):

PR #712, opened 2022-12-23
[compiler] group C++ functions together

quasilyte:
The old PHP->C++ translation schema generated a h/cpp file per every PHP function (same for the methods). There are benefits to that approach, but it results in a slower g++ compilation and somewhat bloated combined object files size.

A simple, but good starting idea is to use 1-to-1 files translation from PHP to C++. So, all functions delcared in a file `foo.php` will be inserted into `foo.cpp` and `foo.h`.

This leads to another problem: the amount of C++ code generated for a PHP file can be enormous. 1 PHP line can translate into several (dozens sometimes) C++ lines.

To avoid that issue, we split the PHP file into several C++ file chunks. This chunk depends on the sizes of the functions. Small files are still transtaled in 1-to-1 manner, but bigger files will end up having several chunks.
The chunk consists of multiple functions that form a bundle. (The exact terminology here is not that important.)

As an extra optimization, private methods are bundled separately, using bigger chunks. This is due to a fact that private methods have a much more limited scope: they will be included only in a few C++ files that have access to that private API of a class.

Since lambdas are stored separately, this patch skips them completely, they're still stored in the separate files (and separate o_l directory).

Some stats:

	* about 3-4 times less "make jobs" (object files) for KPHP projects
	* around 25-40% total objects file size reduction (good for nocc, etc.)
	* faster cold (first) build times:
		* KTemplate "hello world" builds 4-5 times faster! (15s vs 75s)
		* kphp game (based on KPHP SDL) builds ~2 times faster (9s vs 21s)
		* KSQLite examples build ~20% faster (7s vs 9s)
		* Bigger projects with nocc systems can have 5-10% build speedup (harder to measure)


unserialize:
I have an idea about lambdas.
I suppose, that a lambda function can be embedded into a parent function's cpp file, since it could be used only there. Probably, lambda function may have not h-file at all.

PR #727, opened 2023-01-12
handle switch over a constexpr string, a pattern for generic factories

unserialize:
Suppose you want something like `$factory->getInstance<T>()`, to be used like
```php
$factory->getInstance(FooService::class)->fooMethod();
$factory->getInstance(AnotherService::class)->anotherMethod();
```

Here is a possible implementation:
```php
/**
 * @kphp-generic TService
 * @param class-string<TService> $serviceClass
 * @return TService
 */
function getInstance(string $serviceClass) {
  switch ($serviceClass) {
  case FooService::class:
    return $this->foo ?? ($this->foo = $this->createFoo());
  case AnotherService::class:
    return $this->getAnother();
  // more services
  }
  
  if (isset(self::HIDDEN[$serviceClass])) {
    throw new Exception("$serviceClass is private and can not be used");
  }
  throw new Exception("$serviceClass not found");
}
```

### This PR makes it work as expected. Before, it did not

Let's see how `getInstance<FooService>` would look like:
```php
function getInstance<FooService>(string $serviceClass = 'FooService'): FooService {
  switch ($serviceClass) {
  case FooService::class:
    // return FooService object, ok
  case AnotherService::class:
    // return another object, ERROR! types mismatch, FooService expected
  }
  ...
}
```

As you see, without any modifications, the code is erroneous. 

But we have a guarantee that `$serviceClass` is 'FooService' in this cocrete generic specialization.
In other terms, it's a **constexpr** compile-time known string. We have a switch over a constexpr string,
so the only case is valid, which is also calculated at compile-time, so `switch` itself could be just replaced 
with a body of a true `case`, or `default` if all cases are false. Also, `$serviceClass` variable can not be modified.

### Implementation

KPHP automatically performs such transformations for `switch` over constexpr a string, with constexpr cases,
where every non-empty case ends with break/return/throw. Constructions like `if/else` are not analyzed.

More precisely, the following AST is produced:
```php
function getInstance<FooService>(string $serviceClass = 'FooService'): FooService {
  switch ($serviceClass) {
  default:
    // body of true case 'FooService::class'
    return $this->foo ?? ($this->foo = $this->createFoo());
  }
  ...
}
```

Instead of a switch with many cases, KPHP leaves a switch with an only `default` case containing body of a true case. It's easier then to replace the switch itself, as we don't have to eliminate `break;` inside.

### DIFF
--- a/compiler/code-gen/vertex-compiler.cpp
+++ b/compiler/code-gen/vertex-compiler.cpp
@@ -1260,6 +1260,9 @@ void compile_switch(VertexAdaptor<op_switch> root, CodeGenerator &W) {
     compile_switch_str(root, W);
   } else if (root->kind == SwitchKind::IntSwitch) {
     compile_switch_int(root, W);
+  } else if (root->kind == SwitchKind::EmptySwitch) {                         // will contain only inlined 'default' case
+    W << "static_cast<void>(" << root->matched_with_one_case() << ");" << NL; // suppress g++ 'unused variable' warning
+    compile_switch_var(root, W);
   } else {
     compile_switch_var(root, W);
   }
--- a/compiler/gentree.cpp
+++ b/compiler/gentree.cpp
@@ -696,7 +696,7 @@ VertexAdaptor<op_ternary> GenTree::create_ternary_op_vertex(VertexPtr condition,
     return VertexAdaptor<op_ternary>::create(condition, true_expr, false_expr);
   }
 
-  auto cond_var = create_superlocal_var("shorthand_ternary_cond").set_location(condition);
+  auto cond_var = VertexUtil::create_superlocal_var("shorthand_ternary_cond", cur_function).set_location(condition);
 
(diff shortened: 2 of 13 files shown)


compiler/pipes/collect-main-edges.cpp
```diff
@@ -621,8 +621,8 @@ void CollectMainEdgesPass::on_finish() {
   if (current_function->is_extern()) {
     return;
   }
-  if (!have_returns) {
-    // hack to work well with functions which always throws
+  if (!have_returns && !current_function->return_typehint) {
```
quasilyte, line 624: I think this change was included in the other PR too.

quasilyte (dismissed):

compiler/pipes/collect-main-edges.cpp
unserialize: Yes, you're right. That PR would be merged first, and after rebasing, there will be no change here.
This change is necessary for some tests to pass.

quasilyte (approved):

PR #728, opened 2023-01-12
not_null(T) and not_false(T) in @param/@return, useful for generics

unserialize:
Previously, there was not way to express "drop null/false from a type". This attempt is wrong:
```php
/**
 * @kphp-generic T
 * @param ?T $n
 * @return T
 */
function dropNull($n) {
  return not_null($n);
}
```
It's wrong, because when called with `T=?int`, then `$n ::: ?(?int)`, and it returns `?int` obviously. We want `int`.

Now, it's possible to be expressed like this:
```php
/**
 * ...
 * @return not_null(T)
 */
```

Along with `not_null(T)`, there is `not_false(T)`. Square braces are also acceptable.
It's a regular type hint, so even `not_null(?int)` is a valid but meaningless notation.

### DIFF
--- a/compiler/type-hint.cpp
+++ b/compiler/type-hint.cpp
@@ -292,6 +292,17 @@ const TypeHint *TypeHintFuture::create(PrimitiveType ptype, const TypeHint *inne
   );
 }
 
+const TypeHint *TypeHintNotNull::create(const TypeHint *inner, bool drop_not_null, bool drop_not_false) {
+  HasherOfTypeHintForOptimization hash(5192347813149011894ULL);
+  hash.feed_inner(inner);
+  hash.feed_hash(static_cast<uint64_t>(drop_not_null));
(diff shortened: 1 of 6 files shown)


compiler/type-hint.cpp
```diff
@@ -292,6 +292,17 @@ const TypeHint *TypeHintFuture::create(PrimitiveType ptype, const TypeHint *inne
   );
 }
 
+const TypeHint *TypeHintNotNull::create(const TypeHint *inner, bool drop_not_null, bool drop_not_false) {
+  HasherOfTypeHintForOptimization hash(5192347813149011894ULL);
```
quasilyte, line 296: I wish we used some hashing function here like `hash("TypeHintNotNull")` instead of magic numbers like we do it right now. But that's a story for later. :D

quasilyte (approved):

PR #733, opened 2023-01-24
[docs] added `bash_profile` reload command in `epoll-shim` configuration instruction

Tsygankov-Slava:
Hi!
I can suggest some improvement in the [documentation](https://vkcom.github.io/kphp/kphp-internals/developing-and-extending-kphp/compiling-kphp-from-sources.html).
I think you need to mention about reloading `bash_profile` from the command line when configure `epoll-shim`.


unserialize:
Seems to be a part of https://github.com/VKCOM/kphp/pull/732

PR #737, opened 2023-02-01
add getenv() function

drdzyk:
add [getenv()](https://www.php.net/manual/en/function.getenv.php) function

### DIFF
--- a/builtin-functions/_functions.txt
+++ b/builtin-functions/_functions.txt
@@ -1614,3 +1614,5 @@ class DateTimeImmutable implements DateTimeInterface {
   public function getOffset(): int;
   public function getTimestamp(): int;
 }
+
+function getenv(string $varname = '', bool $local_only = false): mixed;
--- a/runtime/env.cpp
+++ b/runtime/env.cpp
@@ -11,6 +11,10 @@ static bool valid_environment_name(const char *name, const char *end) noexcept {
   return std::find_if(name, end, [](char s) { return vk::any_of_equal(s, ' ', '.', '['); }) == end;
 }
 
+#ifdef __APPLE__
+extern char **environ;
+#endif
+
 array<string> f$getenv() noexcept {
   array<string> res;
   for (char **env = environ; env && *env; ++env) {
--- a/runtime/env.cpp
+++ b/runtime/env.cpp
@@ -1,0 +1,31 @@
+// Compiler for PHP (aka KPHP)
+// Copyright (c) 2023 LLC «V Kontakte»
+// Distributed under the GPL v3 License, see LICENSE.notice.txt
+
+#include <cstdlib>
+#include <unistd.h>
+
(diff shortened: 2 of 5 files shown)


runtime/env.cpp
```diff
@@ -0,0 +1,40 @@
...
+
+#include "runtime/env.h"
+
+static bool valid_environment_name(const char *name, const char *end) noexcept {
+  for (const char *s = name; s < end; ++s) {
+    if (*s == ' ' || *s == '.' || *s == '[') {
+      return false;
+    }
+  }
+  return true;
```
mkornaukhov03, line 16: Wouldn't 
```c++
return std::find_if(name, end, [](char s) {
  return vk::any_of_equal(s, ' ', '.', '[');
} == end;
```
be more compact and idiomatic?

drdzyk: true, it's done

builtin-functions/_functions.txt
```diff
@@ -1614,3 +1614,5 @@ class DateTimeImmutable implements DateTimeInterface {
   public function getOffset(): int;
   public function getTimestamp(): int;
 }
+
+function getenv(string $varname = '', bool $local_only = false): mixed;
```
mkornaukhov03, line 1618: Why do you define return type as `mixed`, meanwhile the return type is `string | false` aka `Optional<String>`?
Why don't you add `getenv() : array` to this file?
Could explain these moments, please?

mkornaukhov03:
Maybe we should add some C++ test for some corner-cases like empty `environment`, invalid `environment`, null `environment` etc, what do you think?

runtime/env.cpp
```diff
@@ -0,0 +1,35 @@
...
+extern char **environ;
+#endif
+
+array<string> f$getenv() noexcept {
+  array<string> res;
+  for (char **env = environ; env && *env; ++env) {
+    const char *ptr = std::strchr(*env, '=');
+    // malformed entry?
+    if (ptr && ptr != *env && valid_environment_name(*env, ptr)) {
+      const string key{*env, static_cast<uint32_t>(ptr - *env)};
```
mkornaukhov03, line 24: Maybe `std::distance(*env, ptr) ` would be more idiomatic

builtin-functions/_functions.txt
drdzyk: Because I can't overload function in php.

drdzyk:
> Maybe we should add some C++ test for some corner-cases like empty `environment`, invalid `environment`, null `environment` etc, what do you think?

I guess that these functions are quite simple and phpt tests will suffice.

runtime/env.cpp
drdzyk: Let me leave this little thing as is please

mkornaukhov03 (approved):
LGTM

PR #745, opened 2023-02-13
Implement unittests for msgpack

mkornaukhov03:
We previously had only phpt tests for Message Pack. This PR adds GTest unit tests to check the correctness of `f$msgpack_deserialize/f$msgpack_serialize` without building kphp2cpp.

### DIFF
--- a/tests/cpp/runtime/msgpack-test.cpp
+++ b/tests/cpp/runtime/msgpack-test.cpp
@@ -1,36 +1,37 @@
 #include "runtime/kphp_core.h"
 #include "runtime/msgpack-serialization.h"
+#include <array>
 #include <gtest/gtest.h>
 #include <utility>
 
-constexpr int64_t edge_int_values[] = {
-  0,
-  1,
-  -1,
+constexpr auto edge_int_values = std::array{
+  0_i64,
+  1_i64,
+  -1_i64,
   std::numeric_limits<int64_t>::max(),
   std::numeric_limits<int64_t>::min(),
-  std::numeric_limits<int32_t>::max(),
-  std::numeric_limits<int32_t>::min(),
+  static_cast<int64_t>(std::numeric_limits<int32_t>::max()),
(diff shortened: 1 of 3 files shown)


tests/cpp/runtime/msgpack-test.cpp
```diff
@@ -0,0 +1,196 @@
+#include "runtime/kphp_core.h"
+#include "runtime/msgpack-serialization.h"
+#include <gtest/gtest.h>
+#include <utility>
+
+constexpr int64_t edge_int_values[] = {
```
drdzyk, line 6: Please use std::array instead c-style array. Same for code below

tests/cpp/runtime/msgpack-test.cpp
```diff
@@ -0,0 +1,196 @@
...
+  string("42num"),
+  string("num42"),
+  string("multi word"),
+  string("multi 12 word 23 with digits 42"),
+  string("! @ #  % ^ & * ()-_+="),
+  string("! @ #  % ^\0& * ()-_+="),
+};
+
+template<typename T>
+void assert_single(const T data) {
```
drdzyk, line 53: It will be cool if you turn this assert in gtest test fixture. You can grep 'TEST_F' for more info here: https://github.com/google/googletest/blob/main/docs/primer.md

mkornaukhov03: I think it wouldn't improve the code. Firstly, we would lose type deducing in case of making `constexpr` arrays as non-static fields but we could leave them as `static constexpr` fields. Besides, `edge_string_fields` isn't `constexpr` so we also lose type deducing. I suppose this change will lead to inconsistency.

drdzyk: ok, got it

drdzyk (approved):

drdzyk (approved):

PR #748, opened 2023-02-14
add flush() function

Ramzeeset:


runtime/interface.cpp
```diff
@@ -506,68 +509,89 @@ shutdown_function_type *shutdown_functions = reinterpret_cast<shutdown_function_
...
+    flushed_http_connection = true;
   }
+  http_send_immediate_response(http_headers ? http_headers->buffer() : nullptr, http_headers ? http_headers->size() : 0,
+                               http_body->buffer(), http_body->size());
+  oub[ob_system_level].clean();
+  static_SB_spare.clean();
+}
 
-  for (int i = first_not_empty_buffer + 1; i <= ob_cur_buffer; i++) {
-    oub[first_not_empty_buffer].append(oub[i].c_str(), oub[i].size());
```
DrDet, line 532: Now it flushed to zero indexed buffer, before that it was `first_not_emty_buffer`. Why is it changed? Won't it break something?

server/php-worker.h
```diff
@@ -82,3 +82,4 @@ class PhpWorker {
 };
 
 extern PhpWorker *active_worker;
+extern bool flushed_http_connection;
```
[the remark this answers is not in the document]
Ramzeeset: fixed

runtime/interface.cpp
Ramzeeset: Good observation, fixed it

DrDet (approved):

PR #753, opened 2023-02-17
fix memory leak in shutdown functions.

drdzyk:
Despite that that `std::function` objects is allocated via `placment new` in `char shutdown_function_storage[...]` storage, calling it's destructors is still required, because `std::function` can allocate in heap itself when stored object is complex. Quoting [cppreference](https://en.cppreference.com/w/cpp/utility/functional/function/function):

> When the target is a function pointer or a [std::reference_wrapper](https://en.cppreference.com/w/cpp/utility/functional/reference_wrapper), small object optimization is guaranteed, that is, these targets are always directly stored inside the [std::function](https://en.cppreference.com/w/cpp/utility/functional/function) object, no dynamic allocation takes place. Other large objects may be constructed in dynamic allocated storage and accessed by the [std::function](https://en.cppreference.com/w/cpp/utility/functional/function) object through a pointer.

This leak costs about ~1 kB per request.


runtime/interface.cpp
```diff
@@ -2373,7 +2373,15 @@ void init_runtime_environment(php_query_data *data, void *mem, size_t mem_size)
...
 
+static void free_shutdown_functions() {
+  for (std::size_t i = 0; i < shutdown_functions_count; ++i) {
+    shutdown_functions[i].~shutdown_function_type();
+  }
+  shutdown_functions_count = 0;
+}
+
 void free_runtime_environment() {
+  free_shutdown_functions();
```
DrDet, line 2384: I think it's better to move this call into `free_interface_lib`
because this all belongs to interface.cpp and in `init_interface_lib` we reset some shutdown functions related variables, you can check here:
https://github.com/VKCOM/kphp/blob/master/runtime/interface.cpp#L2263

DrDet (commented):
Keeping in mind that `free_runtime_environment` is called on timeouts let's consider the case when timeout is raised inside `register_shutdown_function` in the middle of `std::function` constructor. In that case we will call destructor for `std::fuction` that's in incorrect state. So I offer to add critical section in `register_shutdown_function` and the comment why it is needed

Ahmed-su-do:
def main
import socket
change dictionary
service tax registration
if statement def services while
print ("services pass are wrong ")

runtime/interface.cpp
drdzyk: done

drdzyk:
> Keeping in mind that `free_runtime_environment` is called on timeouts let's consider the case when timeout is raised inside `register_shutdown_function` in the middle of `std::function` constructor. In that case we will call destructor for `std::fuction` that's in incorrect state. So I offer to add critical section in `register_shutdown_function` and the comment why it is needed

done

DrDet (approved):

PR #770, opened 2023-03-07
classof() compatible with constexpr switch for function overloading

### LINKED ISSUE #763: Support `constexpr switch`  over `classof`
[Here](https://vkcom.github.io/kphp/kphp-language/static-type-system/generic-functions.html#constexpr-switch) is described `constexpr switch`, but currently it's supported for `class-string` variables only. There's `classof($o)` function which provides compile-time information of `$o`'s type, but it's not supported for `constexpr switch`.

Code like 

```php
/**
  * @kphp-generic T
  * @param T $o
  */
function foo($o) {
  // Have to check every instanceof until satisfied.
  if ($o instance A) {
    foo_a($o);
  } else if ($o instanceof B) {
    foo_b($o);
  } ...
```

could be optimized into 

```php
/**
  * @kphp-generic T
  * @param T $o
  */
function foo($o) {
  // Only single branch stays after compilation
  switch (classof($o)) {
    case A::class:
      foo_a($o);
      break;
    case B::class:
      foo_b($o);
      break;
    ...
  }
```

unserialize:
Now, `classof($generic_arg)` is handled at generic functions instantiation and produces a constexpr string, so that `switch(classof($generic_arg))` is treated as a constexpr switch within a generic function.

This gives an ability to implement function overloading for arguments if that arguments are instances (even without a common base):
```php
function overloaded(object $o) {
  // common code

  switch (classof($o)) {
  case ModelA::class:
    ...
    break;
  case ModelB::class:
    ...
    break;
  }

  // common code
}
```

From this PR, `classof()` became more strict: think of it like a compiler keyword available only for generic parameters. `classof(new A)` and `classof($local_var)` won't work.

### `classof($o)` vs `$o::class`

The latter syntax is much more pretty and will be available some time after.

For now, it can't be used in PHP, since it's available from PHP 8 only, and is a syntax error in PHP 7.4.

Closes #763

### DIFF
--- a/compiler/pipes/instantiate-generics-and-lambdas.cpp
+++ b/compiler/pipes/instantiate-generics-and-lambdas.cpp
@@ -20,6 +20,7 @@
  *    and therefore pipes __construct and __invoke of that class
  * 2) it instantiates generic functions knowing T1, etc
  *    and therefore pipes a newly-created function (an instantiated clone of generic function)
+ *    (note, that in InstantiateGenericFunctionPass, func_id and similar are not set: f is yet to pass the pipeline)
  */
 
 class InstantiateGenericFunctionPass final : public FunctionPassBase {
@@ -38,6 +39,7 @@ public:
 
   VertexPtr on_enter_vertex(VertexPtr root) override {
     if (auto as_phpdoc_var = root.try_as<op_phpdoc_var>()) {
+      // inside f<A>, replace `@var T[] $v` with `@var A[] $v`
       as_phpdoc_var->type_hint = phpdoc_replace_genericTs_with_reified(as_phpdoc_var->type_hint, instantiationTs);
 
     } else if (auto as_call = root.try_as<op_func_call>()) {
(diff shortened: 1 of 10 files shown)


compiler/pipes/instantiate-generics-and-lambdas.cpp
```diff
@@ -48,15 +50,22 @@ class InstantiateGenericFunctionPass final : public FunctionPassBase {
           type_hint = phpdoc_replace_genericTs_with_reified(type_hint, instantiationTs);
         }
       }
+      // inside f<A>, replace `classof($o)` with "A" if `f<T>(T $o)` called as `f(expr_assumed_A)`
+      if (as_call->str_val == "classof" && as_call->size() == 1 && as_call->extra_type != op_ex_func_call_arrow) {
```
mkornaukhov03, line 54: Why do you check `as_call->extra_type != op_ex_func_call_arrow`?
Could you give an example where it is the case?

unserialize: Just any arrow call with such a method name :)
```
$obj->classof(...)
```
It's represented as `classof($obj, ...)` in AST with `extra_type = op_ex_func_call_arrow`

mkornaukhov03 (approved):

PR #784, opened 2023-03-11
to_array_debug(): suppressed visitting private and protected members.…

comm644:
For to_array_debug(): suppressed visitting private and protected members. Because this method converts class instance to mixed for  json_encode() , but json_encode must not convert private and protected fields.

test case:
```php
<?php
class A
{
        public array $b = [];
        private int $privateVal = 1;
        protected  int $protectedVal = 2;
}
$obj = new A;

//echo json_encode($obj ); //does not accept class instance

$mixed= to_array_debug($obj, false);
print_r(  $mixed ); //for manual check

if  (count($mixed) !== 1 ) throw new \RuntimeException("to_array_debug() generated private and protected fields");
echo "ok\n";
```

### DIFF
--- a/compiler/code-gen/declarations.cpp
+++ b/compiler/code-gen/declarations.cpp
@@ -765,6 +765,8 @@ void ClassDeclaration::compile_generic_accept(CodeGenerator &W, ClassPtr klass) 
                                 << "void generic_accept(Visitor &&visitor) " << BEGIN;
   for (auto cur_klass = klass; cur_klass; cur_klass = cur_klass->parent_class) {
     cur_klass->members.for_each([&W](const ClassMemberInstanceField &f) {
+       if  (f.modifiers.is_private() ) return;
+       if  (f.modifiers.is_protected() ) return;
       // will generate visitor("field_name", $field_name);
       W << "visitor(\"" << f.local_name() << "\", $" << f.local_name() << ");" << NL;
     });


vkaverin:
`to_array_debug()` and `json_encode` have no connection between ones. As the name says, `to_array_debug` is for debug purposes only.

Check [`JsonEncoder` and `kphp-json visibility_policy`](https://vkcom.github.io/kphp/kphp-language/howto-by-kphp/json-encode-decode.html#skip-all-private-fields-kphp-json-visibility_policypublic).

comm644:
It's about JsonEncoder. JsonEncoder - It's good but next step for serialize/unserialize objects. Simple way (and migration simple) is next:
```php
echo json_encode( to_array_debug( $obj )) ;
```
for real simple debug already introduced print_r().

On the other hand: it too difficult mark **ALL CLASSES**  with this attribute when I would like to get same behavior as json_encode().

vkaverin:
The only *correct* way is still `JsonEncoder`.
It has a lot of customizations. You can make your own class that [`extends JsonEncoder` and override `visibility_policy`](https://vkcom.github.io/kphp/kphp-language/howto-by-kphp/json-encode-decode.html#encoders-inheritors-of-jsonencoder), so you don't need to mark classes with any annotations.

comm644:
Yes but in polyfill....

```php
class JsonEncoder {
  // these constants can be overridden in child classes
  const rename_policy     = 'none';
  const visibility_policy = 'all';
  const skip_if_default   = false;
  const float_precision   = 0;

  private function __construct() { }
..
```

1. Constructor of JsonEncoder is private. We can't implement derived class.
2. Method ::encode() declared as  static and can't use properties of my instance of   JsonEncoder

Need to update documentation how to implement json_encode_object($instance) with same behaviour as json_encode(). (only public methods for all classes aka global policy)

vkaverin:
1) Private construction doesn't stop you from extending `JsonEncoder`, since you don't need to instantiate it.
2) You don't need to use overridden constants manually, all you need is just to override them and use `YourJsonEncoder::encode`.

PR #788, opened 2023-03-17
Add new vertex for values that are inlined from defines

mkornaukhov:
Add new vertex `op_define_val` that represents a value that is inlined on CalcRealDefine and InlineConstantValue passes.
In future, it would help us to print more meaningful errors (vertex of type `op_define_val` will have enough info about constant provenance) and implement object usage in defines.
This vertex is replaced by its value on CollectConstVars (on bottom-up traversal) pass because after that we don't need any info about the provenance of defined value.

### DIFF
--- a/compiler/pipes/calc-const-types.cpp
+++ b/compiler/pipes/calc-const-types.cpp
@@ -33,7 +33,7 @@ void CalcConstTypePass::calc_const_type_of_class_fields(ClassPtr klass) {
 VertexPtr CalcConstTypePass::on_exit_vertex(VertexPtr v) {
   if (auto as_define_val = v.try_as<op_define_val>()) {
     as_define_val->const_type = as_define_val->value()->const_type;
-    kphp_assert_msg(as_define_val->const_type == cnst_const_val, "Non-constant in const context");
+    kphp_error(as_define_val->const_type == cnst_const_val, "Non-constant in const context");
     return v;
   }
   if (auto as_func_call = v.try_as<op_func_call>()) {
--- a/compiler/pipes/collect-const-vars.cpp
+++ b/compiler/pipes/collect-const-vars.cpp
@@ -235,14 +235,14 @@ bool CollectConstVarsPass::user_recursion(VertexPtr v) {
     if (current_function->type == FunctionData::func_class_holder) {
       ClassPtr c = current_function->class_id;
       c->members.for_each([&](ClassMemberInstanceField &field) {
-      if (field.var->init_val) {
-        run_function_pass(field.var->init_val, this);
-      }
+        if (field.var->init_val) {
+          run_function_pass(field.var->init_val, this);
+        }
       });
       c->members.for_each([&](ClassMemberStaticField &field) {
-      if (field.var->init_val) {
-        run_function_pass(field.var->init_val, this);
-      }
+        if (field.var->init_val) {
+          run_function_pass(field.var->init_val, this);
+        }
       });
     }
   }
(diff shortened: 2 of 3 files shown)


compiler/pipes/collect-const-vars.cpp
```diff
@@ -10,87 +10,212 @@
...
+
+  static std::string on_conv_regexp(VertexAdaptor<op_conv_regexp> v) {
+    if (v->expr()->type() == op_string) {
+      return gen_const_regexp_name(v->expr().as<op_string>()->str_val);
+    }
+    return fallback(v);
+  }
+};
+
+struct ProcessBeforeReplace : public VertexVisitor<ProcessBeforeReplace, VertexPtr> {
```
unserialize, line 117: why doesn't NameGenerator unwrap op_define_val before generating a name?

compiler/pipes/calc-real-defines-values.cpp
```diff
@@ -131,6 +131,9 @@ void CalcRealDefinesAndAssignModulitesF::process_define(DefinePtr def) {
     def->type() = DefineData::def_const;
     def->val = make_const.make_const(def->val);
     def->val->const_type = cnst_const_val;
+    if (auto s = def->val.try_as<op_define_val>()) {
```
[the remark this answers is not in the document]
mkornaukhov03: Where should I look?
Yes, it is necessary in my opinion. It is like a base case for calc-const-types pass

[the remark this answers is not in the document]
unserialize: It seems very strange, because for
```
const X = SOME_DEF;
```
then def->val is op_define_val, and const_type will be set, but for
```
const X = [SOME_DEF];
```
then nested op_define_val const_type won't be set.
Am I wrong?
Maybe, setitng const_type should be done in MakeConst manupilation?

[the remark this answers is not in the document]
mkornaukhov03: You are wrong because we have recursive `process_define_recursive()`.
I did by analogy with line 133. MakeConst may be used in other contexts and that is why I did so.

compiler/pipes/collect-const-vars.cpp
mkornaukhov03: Because we don't need it. For now, we just unwrap `op_define_val` on bottom-up traversal.

compiler/pipes/calc-const-types.cpp
```diff
@@ -30,6 +31,11 @@ void CalcConstTypePass::calc_const_type_of_class_fields(ClassPtr klass) {
 }
 
 VertexPtr CalcConstTypePass::on_exit_vertex(VertexPtr v) {
+  if (auto as_define_val = v.try_as<op_define_val>()) {
+    as_define_val->const_type = as_define_val->value()->const_type;
+    kphp_assert_msg(as_define_val->const_type == cnst_const_val, "Non-constant in const context");
```
unserialize, line 36: Is it kphp_assert or kphp_error?
(kphp_assert means that we don't expect it here, and compiler crashes, whereas kphp_error points to a wrong php code)

mkornaukhov03: Yes, it should be kphp_error

compiler/vertex-util.cpp
```diff
@@ -17,6 +17,9 @@ VertexPtr VertexUtil::get_actual_value(VertexPtr v) {
       return var->var_id->init_val;
     }
   }
+  if (v->type() == op_define_val) {
```
unserialize, line 20: to be similar to `if` above, maybe rewrite it like
```
if (auto define_val = v.try_as<op_define_val>()) {
  return get_actual_value(define_val->value());
}
```

unserialize (approved):

PR #802, opened 2023-04-03
Improve algorithm of iterating over vars during priority choice

mkornaukhov:
We solve constants initialization order problem with setting priority(`dependency level = dep. level`) for each constant variable.
For it, we generate functions that initialize vars with certain `dep level`. For now, we had an non-optimized approach for it: we iterate over all variables and then check, if the current var has requested (in current iteration step) `dep level`. The complexity is `O(n * m)`, where `n`- amount of variables, `m` - maximal dependency level. This PR change the algorithm: we group variables of each `dep level` int separate container (analogue of flat map), so the complexity now is `O(m + n)`.

### DIFF
--- a/compiler/code-gen/files/vars-cpp.cpp
+++ b/compiler/code-gen/files/vars-cpp.cpp
@@ -132,11 +132,11 @@ static void compile_vars_part(CodeGenerator &W, const std::vector<VarPtr> &vars,
   for (size_t dep_level = 0; dep_level < max_dep_level; ++dep_level) {
     FunctionSignatureGenerator(W) << NL << "void const_vars_init_priority_" << dep_level << "_file_" << part_id << "()" << BEGIN;
 
-    for (const auto& var : const_raw_string_vars.vars_by_dep_level(dep_level)) {
+    for (const auto &var : const_raw_string_vars.vars_by_dep_level(dep_level)) {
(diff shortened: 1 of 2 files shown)


compiler/code-gen/files/vars-cpp.cpp
```diff
@@ -128,37 +124,35 @@ static void compile_vars_part(CodeGenerator &W, const std::vector<VarPtr> &vars,
...
+  size_t arr_idx = 0;
+  for (size_t dep_level = 0; dep_level < max_dep_level; ++dep_level) {
     FunctionSignatureGenerator(W) << NL << "void const_vars_init_priority_" << dep_level << "_file_" << part_id << "()" << BEGIN;
 
-    for (size_t ii = 0; ii < const_raw_string_vars.size(); ++ii) {
-      VarPtr var = const_raw_string_vars[ii];
-      if (var->dependency_level == dep_level) {
-        W << VarName(var) << ".assign_raw (&raw[" << const_string_shifts[ii] << "]);" << NL;
-      }
+    for (const auto& var : const_raw_string_vars.vars_by_dep_level(dep_level)) {
```
unserialize, line 135: Our code style is `const auto &var`

unserialize (approved):

Issue #804, opened 2023-04-07
Failed to build kphp-timelib


andreylzmw:
# The problem

https://github.com/VKCOM/kphp/blob/10db2acf990702ea88393e5d6f89e3e617fb12c8/cmake/external-libraries.cmake#L44-L55

https://github.com/VKCOM/kphp/blob/10db2acf990702ea88393e5d6f89e3e617fb12c8/cmake/init-compilation-flags.cmake#L120-L121

The problem is that if you do not install `kphp-timelib`, then if it is missing and the `DOWNLOAD_MISSING_LIBRARIES` flag is `On`, cmake cannot compile and link it. The problem is compiler flag `-Wno-redundant-move`, which is from `init-compilation-flags.cmake`, which only works for C++/ObjC++ but not for C (`kphp-timelib` written in C).
```
-- ------kphp-timelib---------
-- kphp-timelib library is not found and will be downloaded automatically
...
Scanning dependencies of target kphp-timelib
[  3%] Building C object _deps/kphp-timelib-build/CMakeFiles/kphp-timelib.dir/astro.c.o
[  3%] Built target prepare_unicode_data
cc1: error: command line option ‘-Wno-redundant-move’ is valid for C++/ObjC++ but not for C [-Werror]
[  3%] Building C object _deps/kphp-timelib-build/CMakeFiles/kphp-timelib.dir/dow.c.o
...
make: *** [Makefile:163: all] Error 2
```

# Steps to reproduce:
1. `sudo apt-get remove kphp-timelib`
2. `git clone https://github.com/VKCOM/kphp.git`
3. `cd kphp && mkdir build && cd build && cmake .. -DDOWNLOAD_MISSING_LIBRARIES=On`
4. `make -j$(nproc)`

My system: Ubuntu 20.04

# My solution

https://github.com/VKCOM/kphp/pull/805

After that:
```
-- ------kphp-timelib---------
-- kphp-timelib library is not found and will be downloaded automatically
...
Scanning dependencies of target kphp-timelib
[  6%] Building C object _deps/kphp-timelib-build/CMakeFiles/kphp-timelib.dir/astro.c.o
[  6%] Building CXX object CMakeFiles/kphp2cpp_src.dir/compiler/code-gen/code-gen-task.cpp.o
[  6%] Building C object _deps/kphp-timelib-build/CMakeFiles/kphp-timelib.dir/dow.c.o
[  6%] Built target generate_unicode_utils
[...]

PR #805, opened 2023-04-07
Fix build C libraries inside kphp

andreylzmw:
fix issue https://github.com/VKCOM/kphp/issues/804


builtin-functions/_functions.txt
```diff
@@ -1622,3 +1622,6 @@ class DateTimeImmutable implements DateTimeInterface {
 }
 
 function getenv(string $varname = '', bool $local_only = false): mixed;
+
+/* mbstring */
+function mb_convert_encoding(string $str, string $to, string $from): string;
```
Danil42Russia, line 1627: Is this really relevant to this PR?

andreylzmw: Sorry, no, fixed now.

cmake/external-libraries.cmake
```diff
@@ -54,6 +54,8 @@ else()
     add_link_options(-L${kphp-timelib_SOURCE_DIR}/objs)
 endif()
 
+add_compile_options(-Wno-redundant-move)
```
Ramzeeset, line 57: Let's move the `kphp-timelib` installation to the top, to use the flag to build other libraries. Also let's add an explanatory comment why this option is added here

andreylzmw: Ok, done

troy4eg:
resolved in https://github.com/VKCOM/kphp/pull/808

PR #810, opened 2023-04-20
[runtime] add flag for the implementation version of gmmktime 

Ramzeeset:
General
-----------
This pr adds a function `set_use_updated_gmmktime(...)` to enable the updated implementation of the gmmktime function

Details
---------
In the past, the source code of gmmktime used the trick of changing the environment variable "TZ" and using `mktime`. This could cause it to not work correctly on some platforms. Now taking the timestamp is done with `timegm`. The magic constants were also removed

### DIFF
--- a/builtin-functions/_functions.txt
+++ b/builtin-functions/_functions.txt
@@ -1401,7 +1401,7 @@ function set_detect_incorrect_encoding_names_warning(bool $show) ::: void;
 
 function set_json_log_on_timeout_mode(bool $enabled) ::: void;
 
-function set_use_updated_gmmktime($enable ::: bool) ::: void;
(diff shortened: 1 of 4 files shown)


troy4eg (dismissed):

builtin-functions/_functions.txt
```diff
@@ -1401,6 +1401,8 @@ function set_detect_incorrect_encoding_names_warning(bool $show) ::: void;
 
 function set_json_log_on_timeout_mode(bool $enabled) ::: void;
 
+function set_use_updated_gmmktime($enable ::: bool) ::: void;
```
unserialize, line 1404: bool $enabled, like above

unserialize (approved):

PR #815, opened 2023-04-26
Allow trailing comma in tuples

mkornaukhov03:
Code like `tuple(1,2,)` was compiled with error. This PR fixes this issue,  such code is going to be successfully compiled.

### DIFF
--- a/compiler/gentree.cpp
+++ b/compiler/gentree.cpp
@@ -656,7 +656,7 @@ VertexPtr GenTree::get_expr_top(bool was_arrow, const PhpDocComment *phpdoc) {
       return_flag = false; // array is valid postfix expression operand
       break;
     case tok_tuple:
-      res = get_func_call<op_tuple, op_err>();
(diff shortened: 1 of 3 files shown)


compiler/gentree.cpp
```diff
@@ -656,7 +656,7 @@ VertexPtr GenTree::get_expr_top(bool was_arrow, const PhpDocComment *phpdoc) {
       return_flag = false; // array is valid postfix expression operand
       break;
     case tok_tuple:
-      res = get_func_call<op_tuple, op_err>();
+      res = get_func_call<op_tuple, op_none>();
```
unserialize, line 659: I am not sure, now such code will now be evaluated: `tuple(1,,,,2)`. Previously, it was an error. Can you check it now?

mkornaukhov03: Yes, I have added an appropriate test

unserialize (approved):

PR #816, opened 2023-04-26
changed default encoding name

Danil42Russia:


runtime/string_functions.cpp
```diff
@@ -14,7 +14,7 @@
 #include "runtime/interface.h"
 
 const string COLON(",", 1);
-const string CP1251("1251", 4);
+const string CP1251("cp1251", 6);
```
DrDet, line 17: btw, it works fine like this too
```suggestion
const string CP1251("cp1251");
```

DrDet (approved):

runtime/string_functions.cpp
Danil42Russia: done

DrDet (approved):

PR #830, opened 2023-05-23
implement curl_exec_concurrently()

drdzyk:
This PR introduce `curl_exec_concurrently()` function which may be used as drop-in replacement for regular `curl_exec()` with the only difference that it meets our [fork](https://vkcom.github.io/kphp/kphp-language/best-practices/async-programming-forks.html) concept.

### Intro
Let's assume we have a simple server running on `localhost:9000`, which can accept `sleep_time_s` url param and sleep specified amount of seconds before responding. So e.g. it will respond on `localhost:9000/?sleep_time_s=0.5` request after 0.5 second.

Now let's send 2 request and see what happens:
```
function run_curl_request($remote_sleep_s) {
  fprintf(STDERR, "\nstart run_curl_request($remote_sleep_s)\n");
  $ch = curl_init("localhost:9000/?sleep_time_s=$remote_sleep_s");
  $resp = curl_exec_concurrently($ch);
  curl_close($ch);
  fprintf(STDERR, "\nfinish run_curl_request($remote_sleep_s)\n");
  return $resp;
}
# here we run 2 requests
$fut1 = fork(run_curl_request(0.1));
$fut2 = fork(run_curl_request(0.3));
# and wait for them asynchronously
$res1 = wait($fut1);
$res2 = wait($fut2);
```
The output will be:
```
// we start 2 requests almost simultaneously
start curl_local(0.1)
start curl_local(0.3)
// we get first response after 0.1 sec await
finish curl_local(0.1)
// and now we get the second after another 0.2 sec await
finish curl_local(0.3)
// so total await time is exact 0.3 sec
```
So it is possible to run any amount of forks, and switch between them as soon as one is ready. Needless to say, that with regular `curl_exec()` in this example we would first wait 0.1 sec till end of first request, and then another 0.3 till the second. With total await time 0.4 sec.

### API
The full signature of new function is:
`function curl_exec_concurrently(int $curl_handle, float $timeout = 1.0): ?string;`
As mentioned above, this function is supposed to be used as replacement of `curl_exec()`. It means that you first create curl handle with `curl_init()`, set up any amount of option on that handle and call `curl_exec_concurrently()`.
After that you can either close handle calling `curl_close()`, or re-use further.

There are two differences in signature comparing to regular:
`function curl_exec(int $curl_handle): string|bool;`

1. The return type.
There is an `CURLOPT_RETURNTRANSFER` option affecting return type. Quoting [doc](https://www.php.net/manual/en/function.curl-setopt.php):

> curl_exec() returns true on success or false on failure. However, if the CURLOPT_RETURNTRANSFER option is [set](https://www.php.net/manual/en/function.curl-setopt.php), it will return the result on success, false on failure.

In our case execution performs in async way, so it is unknown when output would be sent to STDOUT. That's why `CURLOPT_RETURNTRANSFER` is forced to `true` making our function returning `?string`

2. Timeout.
There is an `CURLOPT_TIMEOUT_MS`, that:

> set the maximum number of milliseconds to allow cURL functions to execute

It works perfect with `curl_exec()` because you're waiting response, being blocked on the call itself. But this is not the case with `curl_exec_concurrently()`, where you aren't waiting response at all. Instead you just will be notified by epoll when response arrive.
The second param of `curl_exec_concurrently(..., float $timeout)` may be used to set timeout in seconds which works in desired way. 

### Implementation details
Under the hood the curl mutli socket api is used, namely those parts containing [curl_multi_socket_action()](https://curl.se/libcurl/c/curl_multi_socket_action.html). You can read more about it at curl page [here](https://curl.se/libcurl/c/libcurl-multi.html).
As usual, we add file descriptors curl gives us in epoll for monitoring, and call appropriate callback `curl_socketfunction_cb()` in case of activity on them.

For usage of this api we need both easy and multi curl handles. While first we get as param of `curl_exec_concurrently(int $curl_handle, ...)`, the second we create manually under the hood. It is hidden from php code.

**Those file descriptors are removed from epoll as follows:**
1)  curl request was successfully completed. Here curl lib calls `curl_socketfunction_cb(handle, fd, CURL_POLL_REMOVE)` from regular `curl_multi_socket_action()` asking to remove fd.
2) curl request experienced timeout or error. In this case we call `curl_multi_remove_handle()` in order to detach curh easy and curl multi handles. This action also stops all transfers leaving curl easy in valid state for reusage. And again curl calls `curl_socketfunction_cb(handle, fd, CURL_POLL_REMOVE)`.
3) signal received. In this case we will get into global `free_curl_lib()`, where we call `curl_easy_cleanup()` for all handles. This also stops transfers and calls `curl_socketfunction_cb(handle, fd, CURL_POLL_REMOVE)`.

**Order of removing curl's entities**
This is another important topic. [In curl doc](https://curl.se/libcurl/c/curl_multi_cleanup.html) stated, that we firstly should:
1) detach easy and milti handles calling  `curl_multi_remove_handle()`
2) delete easy handle with `curl_easy_cleanup()`
3) and after that same for multi handle with `curl_multi_cleanup()`

Here point 1) fulfilled in any case. But the lifetime of easy handle is under user control, that's why we don't remove curl multi handle at the end of `curl_exec_concurrently()`. Instead, the detached multi handle will be removed at the end of script.

### DIFF
--- a/runtime/curl-async.cpp
+++ b/runtime/curl-async.cpp
@@ -1,0 +1,40 @@
+// Compiler for PHP (aka KPHP)
+// Copyright (c) 2023 LLC «V Kontakte»
+// Distributed under the GPL v3 License, see LICENSE.notice.txt
+
+#include "runtime/curl-async.h"
+
+#include "runtime/resumable.h"
+#include "server/curl-adaptor.h"
+
+class curl_exec_concurrently final : public Resumable {
+private:
+  using ReturnT = Optional<string>;
+
+  std::unique_ptr<CurlRequest> request;
+  double timeout_s{0.0};
+  int resumable_id{0};
+  std::unique_ptr<CurlResponse> response;
+
+public:
+  curl_exec_concurrently(std::unique_ptr<CurlRequest> &&request, double timeout_s) noexcept
+    : request(std::move(request))
+    , timeout_s(timeout_s) {}
+
+  bool run() noexcept final {
+    RESUMABLE_BEGIN
+    resumable_id = vk::singleton<CurlAdaptor>::get().launch_request_resumable(std::move(request));
+    response = f$wait<std::unique_ptr<CurlResponse>, false>(resumable_id, timeout_s);
+    TRY_WAIT(curl_exec_concurrently_label, response, std::unique_ptr<CurlResponse>);
+    RETURN(response ? response->response : ReturnT{false});
+    RESUMABLE_END
+  }
+};
+
+Optional<string> f$curl_exec_concurrently(curl_easy easy_id, double timeout_s) {
+  auto request = CurlRequest::build(easy_id);
+  if (!request) {
+    return false;
+  }
+  return start_resumable<Optional<string>>(new curl_exec_concurrently(std::move(request), timeout_s));
+}
--- a/runtime/curl-async.cpp
+++ b/runtime/curl-async.cpp
@@ -7,6 +7,8 @@
 #include "runtime/resumable.h"
 #include "server/curl-adaptor.h"
 
+namespace curl_async {
+
 class curl_exec_concurrently final : public Resumable {
 private:
   using ReturnT = Optional<string>;
@@ -33,11 +35,12 @@ public:
     RESUMABLE_END
   }
 };
+} // namespace curl_async
 
 Optional<string> f$curl_exec_concurrently(curl_easy easy_id, double timeout_s) {
-  auto request = CurlRequest::build(easy_id);
+  auto request = curl_async::CurlRequest::build(easy_id);
   if (!request) {
     return false;
   }
-  return start_resumable<Optional<string>>(new curl_exec_concurrently(std::move(request), timeout_s));
+  return start_resumable<Optional<string>>(new curl_async::curl_exec_concurrently(std::move(request), timeout_s));
 }
--- a/runtime/curl-async.cpp
+++ b/runtime/curl-async.cpp
@@ -14,23 +14,21 @@ private:
   using ReturnT = Optional<string>;
 
   const double timeout_s{0.0};
-  const int request_id{0};
   int resumable_id{0};
-  std::unique_ptr<CurlRequest> request;
+  const CurlRequest request;
   std::unique_ptr<CurlResponse> response;
 
 public:
-  curl_exec_concurrently(std::unique_ptr<CurlRequest> &&request, double timeout_s) noexcept
+  curl_exec_concurrently(const CurlRequest &request, double timeout_s) noexcept
     : timeout_s(timeout_s)
-    , request_id(request->request_id)
-    , request(std::move(request)) {}
+    , request(request) {}
 
   bool run() noexcept final {
     RESUMABLE_BEGIN
-    resumable_id = vk::singleton<CurlAdaptor>::get().launch_request_resumable(std::move(request));
+    resumable_id = vk::singleton<CurlAdaptor>::get().launch_request_resumable(request);
     response = f$wait<std::unique_ptr<CurlResponse>, false>(resumable_id, timeout_s);
     TRY_WAIT(curl_exec_concurrently_label, response, std::unique_ptr<CurlResponse>);
-    vk::singleton<CurlAdaptor>::get().finish_request(request_id);
+    vk::singleton<CurlAdaptor>::get().finish_request(request);
     RETURN(response ? response->response : ReturnT{false});
     RESUMABLE_END
   }
@@ -38,9 +36,10 @@ public:
 } // namespace curl_async
 
 Optional<string> f$curl_exec_concurrently(curl_easy easy_id, double timeout_s) {
-  auto request = curl_async::CurlRequest::build(easy_id);
-  if (!request) {
+  try {
+    auto request = curl_async::CurlRequest::build(easy_id);
+    return start_resumable<Optional<string>>(new curl_async::curl_exec_concurrently(request, timeout_s));
+  } catch (...) {
     return false;
   }
-  return start_resumable<Optional<string>>(new curl_async::curl_exec_concurrently(std::move(request), timeout_s));
 }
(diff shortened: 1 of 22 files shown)


runtime/curl.h
```diff
@@ -67,3 +67,29 @@ struct CurlMemoryUsage : vk::not_copyable {
 
   friend class vk::singleton<CurlMemoryUsage>;
 };
+
+class CurlRequest : private vk::not_copyable {
```
DrDet, line 71: Let's use only script memory for `CurlRequest`, `CurlResponse` and others if any

DrDet, line 71: Also let's wrap in separate namespace everything related to asynchronous curl

server/curl-adaptor.h
```diff
@@ -0,0 +1,38 @@
...
+
+  std::unique_ptr<CurlResponse> withdraw_response(int request_id) noexcept;
+  void finish_request(int request_id) noexcept;
+  void reset() noexcept;
+
+private:
+  class CurlRequestResumable;
+  int finish_request_processing(int request_id, std::unique_ptr<CurlResponse> &&response) noexcept;
+
+  SignalSafeHashtable<int, std::unique_ptr<CurlRequest>> processing_requests_holder;
```
DrDet, line 36: why do we need separate request holders container? It seems that request holder is resumable `curl_exec_concurrently` itself

server/curl-adaptor.cpp
```diff
@@ -0,0 +1,97 @@
...
+  processing_requests_holder.clear();
+}
+
+int CurlAdaptor::launch_request_resumable(std::unique_ptr<CurlRequest> &&request) noexcept {
+  const int request_id = request->request_id;
+  net_query_t *query = create_net_query();
+  query->slot_id = request_id;
+  query->data = request.release();
+  int resumable_id = register_forked_resumable(new CurlRequestResumable{request_id});
+  processing_requests.insert(request_id, CurlRequestInfo{resumable_id});
```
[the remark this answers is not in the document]
drdzyk: AFAIU, `wait_net(0)` is about to flush 'net_event' (i.e. response), but not just created 'net_query' (i.e. request).
So it looks irrelevant from this POV. Could you elaborate your proposal a bit more?

runtime/curl.cpp
```diff
@@ -979,3 +989,133 @@ void free_curl_lib() noexcept {
...
+  auto curl_response = std::make_unique<CurlResponse>(std::move(response), request_id);
+  vk::singleton<CurlAdaptor>::get().finish_request_resumable(std::move(curl_response));
+}
+
+void CurlRequest::detach_multi_and_easy_handles() const noexcept {
+  f$curl_multi_remove_handle(multi_id, easy_id);
+}
+
+static int curl_epoll_cb(int fd, void *data, event_t *ev) {
+  auto *curl_request = static_cast<CurlRequest *>(data);
```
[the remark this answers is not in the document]
drdzyk: done

runtime/curl.h
drdzyk: good suggesstion, done

drdzyk: done

server/curl-adaptor.h
drdzyk: `processing_requests_holder` needed to maintain `CurlRequest` instances. Those instances must be alive till end of curl requests processing they are responsible for. So that callbacks may do safely:
```
auto *curl_request = static_cast<CurlRequest *>(userp);
```

DrDet: As far as I can see, curl request processing starts at the beginning of `curl_exec_concurrently` and ends at the end of it after `finish_request` here
https://github.com/VKCOM/kphp/pull/830/files#diff-86197d94d86236d2c5587aecc9b48ee592f4c12c1ced26be354ee376de2f9d82R33

And none of callbacks related to this curl request will be called after that point. Therefore we can say that lifetime of `curl_exec_concurrently` resumable always includes lifetime of the corresponding curl request processing. Do I miss something?

drdzyk: You are right, I've got rid of this map.

runtime/curl-async.cpp
```diff
@@ -14,33 +14,32 @@ class curl_exec_concurrently final : public Resumable {
...
   }
 };
 } // namespace curl_async
 
 Optional<string> f$curl_exec_concurrently(curl_easy easy_id, double timeout_s) {
-  auto request = curl_async::CurlRequest::build(easy_id);
-  if (!request) {
+  try {
+    auto request = curl_async::CurlRequest::build(easy_id);
+    return start_resumable<Optional<string>>(new curl_async::curl_exec_concurrently(request, timeout_s));
```
DrDet, line 41: maybe use `std::move(request)` and pass it as rvalue reference? So we will get rid of one extra copy

runtime/curl-async.cpp
```diff
@@ -14,33 +14,32 @@ class curl_exec_concurrently final : public Resumable {
...
     TRY_WAIT(curl_exec_concurrently_label, response, std::unique_ptr<CurlResponse>);
-    vk::singleton<CurlAdaptor>::get().finish_request(request_id);
+    vk::singleton<CurlAdaptor>::get().finish_request(request);
     RETURN(response ? response->response : ReturnT{false});
     RESUMABLE_END
   }
 };
 } // namespace curl_async
 
 Optional<string> f$curl_exec_concurrently(curl_easy easy_id, double timeout_s) {
```
DrDet, line 38: let's mark it as `noexcept` just in case

runtime/curl-async.cpp
drdzyk: I've already done this, but clang-tidy complains that object is trivially copyable. So I changed it to avoid yellow-painted code.

runtime/curl-async.cpp
drdzyk: Let me leave as is pls

PR #832, opened 2023-05-23
added --server-config and --master-name options

troy4eg:
and deprecated --cluster-name option


server/php-engine.cpp
```diff
@@ -1840,12 +1841,23 @@ int main_args_handler(int i, const char *long_option) {
       return 0;
     }
     case 's': {
```
DrDet, line 1843: Also let's add warning in sentry about deprecated option here with this macro `OPTION_ADD_DEPRECATION_MESSAGE('-s/--cluster-name')`

DrDet (approved):

PR #840, opened 2023-06-06
move function bodies of auto generated classes in source files 

drdzyk:
Currently all member function definitions of auto generated classes are located in header files. This slows down compilation. Here I move definitions of next functions in sources files:
1) `msgpack_pack()` and `msgpack_unpack()`
2) `accept()` method for `JsonEncoder`
3) and template:
```
  template<class Visitor>
  void generic_accept(Visitor &&visitor) noexcept;
```
This also requires explicit instantiation of template in source file.

This measures allow to reduce total size of all object files at ~30% and cold compilation time at ~35% in vkcom.

### DIFF
--- a/compiler/code-gen/declarations.cpp
+++ b/compiler/code-gen/declarations.cpp
@@ -809,7 +809,7 @@ static void compile_json_visitor_call(CodeGenerator &W, ClassPtr json_encoder, C
   }
 }
 
-void ClassDeclaration::compile_accept_json_visitor(CodeGenerator &W, ClassPtr klass, bool to_encode, ClassPtr json_encoder) {
+static void do_compile_accept_json_visitor(CodeGenerator &W, ClassPtr klass, bool to_encode, ClassPtr json_encoder, bool compile_declaration_only) {
   bool parent_has_method = false;
   if (ClassPtr parent = klass->parent_class) {
     parent_has_method |= parent->json_encoders.end() != std::find(parent->json_encoders.begin(), parent->json_encoders.end(), std::pair{json_encoder, to_encode});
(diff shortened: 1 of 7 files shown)


compiler/code-gen/declarations.cpp
```diff
@@ -881,6 +901,13 @@ void ClassDeclaration::compile_accept_json_visitor(CodeGenerator &W, ClassPtr kl
   W << END << NL;
 }
 
+void ClassDeclaration::compile_accept_json_visitor(CodeGenerator &W, ClassPtr klass) {
+  for (auto[encoder, to_encode] : klass->json_encoders) {
```
mkornaukhov03, line 905:
```suggestion
  for (auto [encoder, to_encode] : klass->json_encoders) {
```

compiler/code-gen/naming.h
```diff
@@ -166,6 +171,7 @@ class FunctionSignatureGenerator {
   bool final_ = false;
   bool pure_virtual_ = false;
   bool inline_ = false;
+  bool definition_ = false;
```
mkornaukhov03, line 174: Maybe change to `enum`?
Just it's strange a little bit cognitively complex when you call some functions and add the last parameter true or false

compiler/code-gen/declarations.cpp
```diff
@@ -810,7 +819,7 @@ static void compile_json_visitor_call(CodeGenerator &W, ClassPtr json_encoder, C
   }
 }
 
-void ClassDeclaration::compile_accept_json_visitor(CodeGenerator &W, ClassPtr klass, bool to_encode, ClassPtr json_encoder) {
+static void do_compile_accept_json_visitor(CodeGenerator &W, ClassPtr klass, bool to_encode, ClassPtr json_encoder, bool compile_declaration_only) {
```
mkornaukhov03, line 822: As I mentioned, there is even 2 `bool` parameters here. Maybe change `to_encode\to_decode` to `enum`, too?

mkornaukhov03:
Please, validate my reasoning about `template <class T> generic_accept` function.
For example, we have a class:
```php
class User {
  public string $name;
  public int $age;

  public function __construct(string $name, int $age) {
    $this->name = $name;
    $this->age = $age;
  }
}
```

And we have a function:
```php
function foo() {
  var_dump(to_array_debug(new User("Vasiliy", 42)));
}
```
The source code in function requires to generate `generic_accept` for `User`.
Transpiled file for `foo()` will(approximately) look like 
```C++
// foo.cpp

#include <User.h>

void f$foo() {
  f$to_array_debug(User.create(...));
}
```
`User.h` contains template declaration for `User::generic_accept`, so it allows compiler to generate to generate mangled symbol for this function during compilation of `foo.cpp` translation unit.
There is an explicit template instantiating of this method in `User.cpp` translation unit.
And in linking stage, linker ordinary will find required mangled (with template) symbol for `User::generic_accept` in `User.cpp` resulting object file.
Am I right?

compiler/code-gen/declarations.cpp
drdzyk: We use bool params almost everywhere for such cases

unserialize (approved):

PR #841, opened 2023-06-06
enable shutdown functions on several timeout situations

Ramzeeset:
General
-----------
Enable shutdown functions if timeout was in script context and script call swap context or start resumable

Details
---------
Remove longjmp from sigalrm handler. Instead of this remember timeout fact and set up second timeout that is deadline to call shutdown functions from runtime

### DIFF
--- a/server/php-runner.cpp
+++ b/server/php-runner.cpp
@@ -10,7 +10,6 @@
 #include <cstdlib>
 #include <cstring>
 #include <exception>
-#include <execinfo.h>
 #include <sys/mman.h>
 #include <sys/time.h>
 #include <unistd.h>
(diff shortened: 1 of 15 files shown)


server/php-runner.cpp
```diff
@@ -84,17 +58,37 @@ void PhpScript::error(const char *error_message, script_error_t error_type) noex
...
+
+  if (get_shutdown_functions_count() != 0 && get_shutdown_functions_status() == shutdown_functions_status::not_executed) {
+    run_shutdown_functions_from_timeout();
+  }
+  perform_error_if_running("timeout exit\n", script_error_t::timeout);
+}
+
+void PhpScript::check_net_context_errors() noexcept {
+  php_assert(PhpScript::in_script_context);
+  if (memory_limit_exceeded ) {
```
DrDet, line 87:
```suggestion
  if (memory_limit_exceeded) {
```

DrDet (approved):

PR #843, opened 2023-06-07
improving statshouse metrics

troy4eg:
replaced host with cluster name for statshouse

### DIFF
--- a/server/statshouse/worker-stats-buffer.cpp
+++ b/server/statshouse/worker-stats-buffer.cpp
@@ -3,10 +3,11 @@
 // Distributed under the GPL v3 License, see LICENSE.notice.txt
 
 #include "worker-stats-buffer.h"
+#include "common/resolver.h"
 #include "common/tl/constants/statshouse.h"
 #include "runtime/critical_section.h"
+#include "server/server-config.h"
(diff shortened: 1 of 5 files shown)


server/statshouse/worker-stats-buffer.cpp
```diff
@@ -79,9 +80,11 @@ void WorkerStatsBuffer::flush() {
...
   std::vector<StatsHouseMetric> metrics;
   const char *hostname = kdb_gethostname();
+  const char *cluster_name = vk::singleton<ServerConfig>::get().get_cluster_name();
 
   for (size_t i = 0; i < static_cast<size_t>(WorkerType::types_count); ++i) {
     std::vector<tag> tags;
+    tags.emplace_back("cluster", cluster_name);
     tags.emplace_back("worker_type", i == static_cast<size_t>(WorkerType::general_worker) ? "general" : "job");
     if (hostname != nullptr) {
       tags.emplace_back("host", hostname);
```
DrDet, line 90: We wanted to replace `host` with `cluster_name`, but `host` is still present

DrDet (changes requested):

PR #892, opened 2023-09-01
Removing the dependency on the libucontext library

troy4eg:
Removing the dependency on the libucontext library makes it easier to build for macOS on arm processors


DrDet (commented):
Maybe it's better to move all of ucontext stuff to `common/ucontext` folder?

astrophysik (approved):

unserialize (approved):

DrDet (approved):

PR #901, opened 2023-09-16
Added errors for unimplemented features

Tsygankov-Slava:
Fixed `catch` location detection.

Added errors for unimplemented features and tests:
- `finally` syntax
- `readonly` classes syntax
- `first class callable` syntax
- `intersection types`
- `explicit octal integer literal` notation
- `constants in traits`
- `standalone true type`
- `catching multiple exception types`
- `true type`
- `iterable pseudo-type`
- `links in list()`
- `yield` & `yield from`


compiler/code-gen/vertex-compiler.cpp
```diff
@@ -297,6 +297,12 @@ void compile_throw(VertexAdaptor<op_throw> root, CodeGenerator &W) {
 }
 
 void compile_try(VertexAdaptor<op_try> root, CodeGenerator &W) {
+  bool isExistFinally = root->finally_cmd_ref()->type() != Operation::op_empty;
```
Danil42Russia, line 300: change it to `snake_case`

compiler/code-gen/vertex-compiler.cpp
```diff
@@ -297,6 +297,12 @@ void compile_throw(VertexAdaptor<op_throw> root, CodeGenerator &W) {
...
+  if (isExistFinally) {
+    stage::set_location(root->finally_cmd()->location);
+    kphp_error(0, "  construct not implemented");
```
Danil42Russia, line 303: there's a `finally` missing here

compiler/gentree.cpp
```diff
@@ -1704,6 +1752,10 @@ VertexPtr GenTree::get_class(const PhpDocComment *phpdoc, ClassType class_type)
   StackPushPop<FunctionPtr> f_alive(functions_stack, cur_function, cur_class->gen_holder_function(full_class_name));
 
   cur_class->modifiers = modifiers;
+  if (is_readonly) {
+    cur_class->is_readonly = is_readonly;
+    kphp_error(0, "  `readonly` classes is not supported");
```
Danil42Russia, line 1757: Why are there two spaces at the beginning?

compiler/lexer.cpp
```diff
@@ -425,6 +427,12 @@ bool TokenLexerNum::parse(LexerData *lexer_data) const {
   } else if (s[0] == '0' && s[1] == 'b') {
     t += 2;
     state = binary;
+  } else if (s[0] == '0' && (s[1] == 'o' || is_digit(s[1]))) { // 0o123 or 0123
```
Danil42Russia, line 430: There is a missing test for this change

compiler/gentree.h
```diff
@@ -152,4 +154,3 @@ class GenTree {
   FunctionPtr cur_function;         // = functions_stack.back()
   SrcFilePtr processing_file;
 };
-
```
Danil42Russia, line 155: Unnecessary change

compiler/lexer.cpp
```diff
@@ -559,15 +578,15 @@ bool TokenLexerNum::parse(LexerData *lexer_data) const {
...
-  if (!is_float) {
-    if (s[0] == '0' && s[1] != 'x' && s[1] != 'b') {
-      for (int i = 0; i < t - s; i++) {
-        if (s[i] < '0' || s[i] > '7') {
-          return TokenLexerError("Bad octal number").parse(lexer_data);
-        }
-      }
-    }
-  }
+//  if (!is_float) {
```
Danil42Russia, line 581: It's definitely not supposed to be

compiler/gentree.cpp
```diff
@@ -2064,6 +2142,9 @@ VertexPtr GenTree::get_statement(const PhpDocComment *phpdoc) {
     case tok_public:
     case tok_private:
       if (std::next(cur, 1)->type() == tok_const) {
+        if (cur_class->class_type == ClassType::trait) {
+          kphp_error(0, "\"const\" member is not supported in trait");
```
Danil42Russia, line 2146: You have an inconsistent type of quotation marks (**`**, **"**, **'**)

compiler/gentree.cpp
```diff
@@ -1991,10 +2043,19 @@ GenericsInstantiationPhpComment *GenTree::parse_php_commentTs(vk::string_view st
...
+  auto location = auto_location();
+
   CE (expect(tok_catch, "'catch'"));
   CE (expect(tok_oppar, "'('"));
   auto exception_class = cur->str_val;
   CE (expect(tok_func_name, "type that implements Throwable"));
+  bool is_multiple_exception_types = false;
+  while (cur->type() == tok_or) { // processing catching multiple exception types (ExceptionType1 | ExceptionType2 | ...)
+    expect(tok_or, "union types");
+    CE (expect(tok_func_name, "type that implements Throwable"));
```
Danil42Russia, line 2055: I think there's an incorrect error message

Tsygankov-Slava: I think this message is one of the possible.

For example, we have the code:
```php
try {
  // ...
} catch (Exception1 | string $err) {
  // ...
}
```

Output:
```
Expected type that implements Throwable, found 'string'

```

compiler/gentree.h
Tsygankov-Slava: This file has two empty line in the end file - it's bad. Therefore I delete extra empty line.

compiler/data/class-data.h
```diff
@@ -64,6 +64,7 @@ class ClassData : public Lockable {
   bool has_custom_constructor{false};
   bool is_serializable{false};
   bool has_job_shared_memory_piece{false};
+  bool is_readonly{false};
```
mkornaukhov03, line 67: Do we really need it?
is this property not expressed through `is_immutable`?

compiler/pipes/calc-rl.cpp
```diff
@@ -115,6 +115,12 @@ void rl_common_calc(VertexPtr root, RLValueType expected_rl_type) {
       break;
     case op_ffi_new:
       break;
+    case op_yield:
+      kphp_error(false, "yield isn't supported");
+      break;
+    case op_yield_from:
+      kphp_error(false, "yield from isn't supported");
+      break;
```
mkornaukhov03, line 123: Is this code reachable?
I suppose it isn't true.

compiler/pipes/cfg.cpp
```diff
@@ -656,6 +656,10 @@ void CFG::create_cfg(VertexPtr tree_node, Node *res_start, Node *res_finish, boo
       *res_finish = Node();
       break;
     }
+    case op_yield:
+    case op_yield_from:
+      kphp_error(false, "yield isn't supported");
+      break;
```
mkornaukhov03, line 662: Is this code reachable?

compiler/gentree.cpp
```diff
@@ -1024,10 +1042,34 @@ VertexAdaptor<op_return> GenTree::get_return() {
...
   }
+
   CE (expect(tok_semicolon, "';'"));
   return ret.set_location(location);
 }
 
+VertexAdaptor<op_yield> GenTree::get_yield() {
+  auto location = auto_location();
+  next_cur();
+  skip_phpdoc_tokens();
```
mkornaukhov03, line 1053: Why we should skip phpdoc here?

tests/phpt/exceptions/with_finally/02_finally_with_throw.php
```diff
@@ -0,0 +1,18 @@
...
+
+function inverse($x) {
+    if (!$x) {
+        throw new Exception('Division by zero.');
+    }
+    return 1/$x;
+}
+
+try {
+    echo inverse(0);
```
mkornaukhov03, line 13: There's no difference between 02_finally_with_throw.php and 03_finally_without_throw.php
These compiler tests are negative, so compiler doesn't know anything about runtime, that is why no matter which number is argument of `inverse` call

tests/phpt/typehints/params/09_fail_intersection_type.php
```diff
@@ -0,0 +1,12 @@
+@kphp_should_fail
+\Intersection types is not supported\
```
mkornaukhov03, line 2: s/is/are

mkornaukhov03:
I cannot find any test on octal numbers

compiler/pipes/cfg.cpp
Tsygankov-Slava: Yes, it's at this stage that we inform that we don't support `yield`

compiler/pipes/calc-rl.cpp
Tsygankov-Slava: You're right. I'll delete it.

compiler/gentree.cpp
Tsygankov-Slava: Yes, that's an extra action. I'll delete it.

compiler/data/class-data.h
Tsygankov-Slava: I agree, it's better to use `is_immutable `.

PR #911, opened 2023-09-28
StatsHouse metrics improvements

DrDet:
## Host tag toggle function
Now we can dynamically enable `host` tag in all StatsHouse metrics in all workers and master processes.
All we need to do is to call new special toggle function from script execution inside any worker.
It enables host tag for the next `30 sec` in all workers and master for all metrics. This toggle shares between all workers via shared memory with max `5 sec` lag in the worst case.
This makes possible embedding it into vkcom toggles mechanism and manage StatsHouse `host` tag in a very convenient way. 
## New `kphp_by_host_*` metrics
Essential dedicated metrics to detect problem on separate hosts. All of them always use `host` tag. Here're four such metrics:
- **request time**: script and net time
- **workers cpu usage**: sys and user
- **workers heap memory usage**: rss_peak and vm_peak from `/proc/self/status`
- **script errors**: count metric for `timeout`, `memory_limit`, `php_assert` and other errors

They are intentionally made as separate metrics because of very high cardinality, although they copy some other metrics.

## Others
- all of the above changes became possible after StatsHouse related refactoring:
  - separated StatsHouseClient and StatsHouseMetrics entities
  - implemented setting up common tags 
  - refactored singleton implementation details
- added `status` tag for `kphp_request_time` metric to make it possible calculating spent time before script termination with an error
- added `kphp_request_errors` count metric to calculate errors only
- refactored worker and master crons
- other minor refactoring


astrophysik (dismissed):

server/statshouse/statshouse-metrics.h
```diff
@@ -0,0 +1,87 @@
...
+#include "common/mixin/not_copyable.h"
+#include "runtime/memory_resource/memory_resource.h"
+#include "server/job-workers/job-stats.h"
+#include "server/statshouse/statshouse-client.h"
+#include "server/workers-control.h"
+#include "server/workers-stats.h"
+
+enum class script_error_t : uint8_t;
+
+class StatsHouseMetrics : vk::not_copyable {
```
troy4eg, line 19: What about renaming this to StatshouseManager or something else?

server/php-engine.cpp
```diff
@@ -1664,6 +1667,7 @@ void init_all() {
     }
     log_server_warning(deprecation_warning);
   }
+  StatsHouseMetrics::get().init_common_tags(vk::singleton<ServerConfig>::get().get_cluster_name(), kdb_gethostname());
```
troy4eg, line 1670: ServerConfig will have mutable state in the future. For example get_cluster_name() function will be able to return a different value each time it is called.

server/php-engine.cpp
```diff
@@ -2089,7 +2093,7 @@ int main_args_handler(int i, const char *long_option) {
         host = "127.0.0.1";
       }
 
-      StatsHouseClient::init(host, port);
+      StatsHouseMetrics::init_client(host, port);
```
troy4eg, line 2096: I think it is not good to divide initialization into several methods, because potentially we can get a non-valid state

server/statshouse/statshouse-metrics.h
DrDet: done

server/php-engine.cpp
DrDet: done, now we update common tags in cron

server/php-engine.cpp
DrDet: after the above changes here is single initialization, renamed to just `init`

troy4eg (approved):

PR #915, opened 2023-10-12
maintained RpcReqError with headers in vkext

DrDet:
## Context
TL error serialization is historically totally broken and doesn't fit the scheme.
By according to TL scheme it must be like this:
https://github.com/VKCOM/kphp/blob/c4031c748a563f44cc8f58ab07aab90b33261a5c/common/tl-files/common.tl#L79

But actually it comes as:
https://github.com/VKCOM/kphp/blob/c4031c748a563f44cc8f58ab07aab90b33261a5c/common/tl-files/common.tl#L85

`rpcReqError#7ae432f5` is outer error, but it's fetched as inner one instead of `reqError#b527877d`, while latter is never used at all.

## Problem
This weird deserialization is done via workaround as `tl_fetch_lookup` right before the answer fetching. And it's done inconsistently: if we require extra response headers via:
https://github.com/VKCOM/kphp/blob/c4031c748a563f44cc8f58ab07aab90b33261a5c/common/tl-files/common.tl#L47
we will get the error on parsing them in case of the error, although in case of normal response everything is fine.

## Changes

Now we can successfully fetch error with extra fields (header), if we require them.
In typed mode it's fetched as `rpcResponseHeader` object, containing `rpcResponseError` in its `$result` field.
It doesn't fit php-doc types, but it's fair, because it doesn't fit TL scheme at all.
This approach seems for me the best option, as it helps to avoid extra related workarounds in `tl2php`, `tlo-parser`, `kphp`, etc.
In `kphp` headers will just be skipped and `rpcResponseError` will be returned as usual.

----
In the future this workaround will be fixed, and serialization will fit the TL scheme and work correctly in `kphp` and `vkext` the same way.


KokorinIlya (commented):
Парсинг выглядит ок

Hidanjkezor (approved):

vkext/vkext-rpc-tl-serialization.cpp
```diff
@@ -1356,19 +1432,17 @@ void vk_memcache_query_result1(INTERNAL_FUNCTION_PARAMETERS) {
   auto *T = reinterpret_cast<tl_tree *>(q->extra);
   INC_REF (T);
   if (do_rpc_get_and_parse(qid, timeout - precise_now) < 0) {
-    zval *r = make_query_result_or_error(NULL, "Response not found, probably timed out",
-                                         TL_ERROR_RESPONSE_NOT_FOUND);
+    zval *r = make_query_result_or_error(NULL, {TL_ERROR_RESPONSE_NOT_FOUND, "Response not found, probably timed out"});
```
troy4eg, line 1435: may be refactor all NULL to nullptr?

troy4eg (dismissed):

troy4eg (approved):

vkext/vkext-rpc-tl-serialization.cpp
DrDet: I think it's better to do it in separate PR

PR #922, opened 2023-10-27
refactor net-level alarm handling

astrophysik:


server/php-worker.cpp
```diff
@@ -127,7 +139,7 @@ void PhpWorker::state_try_start() noexcept {
 }
 
 void PhpWorker::state_init_script() noexcept {
-  double timeout = finish_time - precise_now - 0.01;
```
astrophysik, line 130: Here is potentially dangerous place for job workers. Before that pr every timeout was less on 10ms

PR #927, opened 2023-11-01
Install h3 from cmake

andreylzmw:


troy4eg:
Can you add about -DOWNLOAD_MISSING_LIBRARIES to the Mac OS build documentation? Also, as far as I know,if we use this cmake flag, we no longer need to clone the epoll-shim repository. May be remove this part?

https://github.com/VKCOM/kphp/pull/927/files#diff-efd602afe9fcc34ab6c7f2fa70ee154e91c0e0adf1b3ff73d9c51da1e1541b8cR91

troy4eg (approved):

PR #938, opened 2023-11-13
prohibition of parallel rpc requests 

astrophysik:


server/php-engine.cpp
```diff
@@ -828,13 +829,16 @@ int rpcx_func_wakeup(connection *c) {
   } else {
     if (c->pending_queries < 0 || c->status != conn_wait_net) {
       std::array<char, 1024> message{'\0'};
-      int id = rand();
+      std::random_device rd;
+      std::mt19937 gen(rd());
```
DrDet, line 833: It seems you don't need to use `std::mt19937` here

server/php-engine.cpp
```diff
@@ -975,6 +979,11 @@ int rpcx_execute(connection *c, int op, raw_message *raw) {
       return 0;
     }
     case TL_RPC_INVOKE_REQ: {
+      if (php_worker.has_value()) {
+        //check that another rpc request has already been processed
+        vkprintf(1, "get new RPC_INVOKE_REQ while another rpc query is running\n");
```
DrDet, line 984: let's write error to sentry via `log_server_warning`

server/php-engine.cpp
```diff
@@ -975,6 +962,11 @@ int rpcx_execute(connection *c, int op, raw_message *raw) {
       return 0;
     }
     case TL_RPC_INVOKE_REQ: {
+      if (php_worker.has_value()) {
+        //check that another rpc request has already been processed
+        log_server_warning( "get new RPC_INVOKE_REQ while another rpc query is running\n");
```
DrDet, line 967: I think we need to send RPC error response via `send_rpc_error` like below

DrDet (approved):

PR #939, opened 2023-11-14
add support of polymorphic types in instance cache

astrophysik:
# Add support of polymorphic types in instance cache
A polymorphic type `can be stored` in `instance cache` if the following conditions are met
1. Type is `immutable` or it is `interface`
2. All `fields` of user types `can be stored` (use dfs for searching this solves the problems with cycles)
3. If type has `derivatives` all of them `can be stored`

Example of use
```php
/** @kphp-immutable-class */
Master extends Simple {}
/** @kphp-immutable-class */
Worker extends Simple {}

...

function save_type(string $key, Simple $obj) {
    instance_cache_store($key, $obj);
}
save_type("key", new Worker());

...

$obj = instance_cache_store(Simple:class, "key");
if ($obj instanceof Worker) {
  // special logic
}
if ($obj instanceof Master) {
  // special logic
}
```

## Compiler
Since we can't add more stringent conditions to existing checks because it would stop working in the old code, all new checks are called from the `check_instance_cache_store_call` and `check_instance_cache_fetch_call`.
Few details 

* Fields checking can lead to cycles for this purpose the flag `process_fields_ic_compatibility` is used inside `check_fields_ic_compatibility`
* In order not to process the same subtrees several times flag `is_subtree_immutable` is used for optimization. It can take the values ` immutable`,  `not_immutable`, `not_visited`

## Runtime
For objects copying used methods `virtual_builtin_functions` that generated on compilation 

## phpt tests
Add tests to compile different types of code
## python tests
Add tests for errors related to interaction in the shared memory


mkornaukhov03:
Wouldn't it be better to explicitly store and load to atomic fields via `load()` and `store()` member function? I suppose it will signal code readers that these variables are atomic, for now it's kind of muddle.

compiler/pipes/final-check.cpp
```diff
@@ -41,6 +41,62 @@ void check_class_immutableness(ClassPtr klass) {
...
+    } else if (derived->is_subtree_immutable == SubtreeImmutableType::not_immutable) {
+      // do not save them because they have already been processed
+      has_mutable_subtree = true;
+    } else if (!derived->is_immutable && !derived->is_interface()) {
+      mutable_children.push_back(derived);
+    } else {
+      check_fields_ic_compatibility(derived);
+      std::list<ClassPtr> descendant = find_not_ic_compatibility_derivatives(derived);
+      mutable_children.splice(mutable_children.end(), descendant);
+    }
```
mkornaukhov03, line 92: I suppose it could be rewritten in iterative-style DFS without extra allocations for `std::list`

compiler/pipes/final-check.cpp
```diff
@@ -41,6 +41,62 @@ void check_class_immutableness(ClassPtr klass) {
                         TermStringFormat::paint(klass->parent_class->name, TermStringFormat::red)));
 }
 
+std::list<ClassPtr> find_not_ic_compatibility_derivatives(ClassPtr klass);
+
+void check_fields_ic_compatibility(ClassPtr klass) {
+  if (klass->process_fields_ic_compatibility) {
+    return;
+  }
+  klass->process_fields_ic_compatibility = true;
```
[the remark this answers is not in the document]
astrophysik: fix

compiler/pipes/final-check.cpp
astrophysik: Rewrite dfs and change buffer to vector

compiler/pipes/final-check.cpp
```diff
@@ -64,36 +66,38 @@ void check_fields_ic_compatibility(ClassPtr klass) {
...
-      // do not save them because they have already been processed
-      has_mutable_subtree = true;
-    } else if (!derived->is_immutable && !derived->is_interface()) {
-      mutable_children.push_back(derived);
-    } else {
-      check_fields_ic_compatibility(derived);
-      std::list<ClassPtr> descendant = find_not_ic_compatibility_derivatives(derived);
-      mutable_children.splice(mutable_children.end(), descendant);
+  std::vector<ClassPtr> mutable_children{};
+  std::vector<ClassPtr> queue = {klass};
```
mkornaukhov03, line 82: I think that's a stack, not a queue

compiler/pipes/final-check.cpp
```diff
@@ -4,6 +4,8 @@
 
 #include "compiler/pipes/final-check.h"
 
+#include <stack>
```
mkornaukhov03, line 7: Redundant include

mkornaukhov03 (changes requested):

compiler/data/class-data.h
```diff
@@ -23,6 +23,12 @@ struct FFIClassDataMixin;
 struct FFIScopeDataMixin;
 namespace kphp_json { class KphpJsonTagList; }
 
+enum class SubtreeImmutableType {
+  immutable,
+  not_immutable,
```
mkornaukhov03, line 28: I would change at `mutable`, but up to you.

mkornaukhov03 (approved):

compiler/data/class-data.h
astrophysik: mutable is a key word

DrDet (approved):

PR #946, opened 2023-12-03
Make master outlive confdata out of shared memory errors

DrDet:
Here're two new approaches:
## Events throttling
To prevent OOM errors on confdata overflows we start to throttle events, when too little shared memory left. There're two modes:
- Soft OOM (85%) - replay only existing keys related events until server restart. Ignore new keys.
- Hard OOM (95%) - replay nothing. Ignore all events, don't read binlog at all.

If server enters any OOM modes, it can't go back out of it. Manual server restart is needed for that.

## External preventive allocation heuristic checks
Before any action, that internally invokes allocations, we explicitly check that allocator has enough memory for that.
To achieve this we need to somehow calculate upper-bound for memory that will be required by the action in each particular case.
There're two 'kinds' of allocations:
1. Allocation made by our runtime (`array<mixed>::mutate_if_shared()`, etc..). It's not too complicated, because it's our own code and we can calculate it easily.
2. Allocation made by STL, out of `std::map<>` which uses our shared memory allocator.  To handle this we will use common sense, some knowledge about search trees internal structure and compile time upper-bound for tree node size from our allocator: https://github.com/VKCOM/kphp/blob/d685b6cce3cc351d48816d7bc982800c93b18d52/runtime/memory_resource/resource_allocator.h#L48

## New stats for confdata 

This PR also includes writing confdata metrics to statshouse:
- `kphp_confdata_memory` 
- `kphp_confdata_events`
- `kphp_confdata_sections_by_count`
- `kphp_confdata_update_fails`

## New option for blacklisting confdata events

`--confdata-force-ignore-keys-prefix` — an emergency option, e.g. 'highload.vid*', to forcibly drop keys from snapshot/binlog; may be used multiple times.

## Timeouts for reading binlog

To prevent master hanging in `master_cron()`, a timeout for reading a next binlog chunk is added, controlled by `--confdata-update-timeout`. Every second (every cron iteration) we allow binlog reader to consume next event no more than timeout seconds. That allows master to perform other work.

Note, that an option for reading binlog in a separate thread can't be implemented, since the master process is also fundamentally single-threaded (lots of globals, global allocator, global allocator switches between heap/shared memory, etc.).

### DIFF
--- a/server/confdata-binlog-replay.cpp
+++ b/server/confdata-binlog-replay.cpp
@@ -10,6 +10,7 @@
 #include <cmath>
 #include <forward_list>
 #include <map>
+#include <optional>
 #include <string_view>
 
 #include "common/binlog/binlog-replayer.h"
@@ -36,6 +37,7 @@ struct {
   size_t memory_limit{2u * 1024u * 1024u * 1024u};
   double soft_oom_threshold_ratio = 0.85;
   double hard_oom_threshold_ratio = 0.95;
+  double confdata_update_timeout_sec = 0.3;
   std::unique_ptr<re2::RE2> key_blacklist_pattern;
   std::forward_list<vk::string_view> force_ignore_prefixes;
   std::unordered_set<vk::string_view> predefined_wildcards;
@@ -50,6 +52,7 @@ public:
   enum class OperationStatus {
     no_update,
     throttled_out,
+    timed_out,
     blacklisted,
     ttl_update_only,
     full_update
@@ -170,6 +173,7 @@ public:
     for (int i = 0; i < nrecords; i++) {
       if (index_offset[i] >= 0) {
         auto res = store_element(reinterpret_cast<const entry_type &>(index_binary_data[index_offset[i]]));
+        assert(res != OperationStatus::timed_out);
         if (res == OperationStatus::throttled_out) {
           ret_code = -1;
           raise_confdata_oom_error("Can't read confdata snapshot on start");
@@ -280,7 +284,7 @@ public:
       }
       assert(head->second.size() <= std::numeric_limits<short>::max());
       auto res = delete_element(head->second.c_str(), static_cast<short>(head->second.size()));
-      if (res == OperationStatus::throttled_out) {
+      if (res == OperationStatus::throttled_out || res == OperationStatus::timed_out) {
         return;
       }
       assert(expired_elements == expiration_trace_.size() + 1);
@@ -295,7 +299,24 @@ public:
   const ConfdataStats::EventCounters &get_event_counters() const noexcept {
     return event_counters_;
   }
+  
+  void on_start_update_cycle(double update_timeout_sec) noexcept {
+    binlog_update_start_time_point_ = std::chrono::steady_clock::now();
+    if (update_timeout_sec > 0) {
+      update_timeout_sec_ = std::chrono::duration<double>{update_timeout_sec};
+    }
+  }
 
+  bool on_finish_update_cycle() noexcept {
+    bool ok = !is_update_timeout_expired();
+    update_timeout_sec_.reset();
+    return ok;
+  }
+  
+  bool is_update_timeout_expired() const noexcept {
+    return update_timeout_sec_.has_value() &&
+           std::chrono::steady_clock::now() - binlog_update_start_time_point_ > update_timeout_sec_;
+  }
 private:
   ConfdataBinlogReplayer() noexcept:
     garbage_from_previous_confdata_sample_(new(&garbage_mem_) GarbageList{}),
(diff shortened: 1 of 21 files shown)


server/confdata-binlog-replay.cpp
```diff
@@ -370,6 +469,11 @@ class ConfdataBinlogReplayer : vk::binlog::replayer {
       return OperationStatus::no_update;
     }
 
+    if (array_for_second_key.get_reference_counter() > 1) {
```
unserialize, line 472: Please, add a comment, what this condition means

server/confdata-binlog-replay.cpp
```diff
@@ -44,6 +65,40 @@ class ConfdataBinlogReplayer : vk::binlog::replayer {
...
+    if (soft_oom_reached_) return MemoryStatus::SOFT_OOM;
+    return MemoryStatus::NORMAL;
+  }
+
+  bool check_has_enough_memory(size_t need_bytes, std::string_view msg) noexcept {
+    if (memory_resource_->is_enough_memory_for(need_bytes)) {
+      return true;
+    }
+    log_server_critical("Not enough confdata shared memory. Processing key with first part = '%s'. %zu bytes needed by estimation for %s",
+                        processing_key_.get_first_key().c_str(), need_bytes, msg.data());
```
unserialize, line 89: In general, `.data()` is incorrect for `string_view`, since it may not end with `\0`. Probably, the easiest way is just to use `const char *`, as below

server/confdata-binlog-replay.cpp
```diff
@@ -114,28 +169,44 @@ class ConfdataBinlogReplayer : vk::binlog::replayer {
 
     // disable the blacklist because we checked the keys during the previous step
     blacklist_enabled_ = false;
+    int ret_code = 0;
     for (int i = 0; i < nrecords; i++) {
       if (index_offset[i] >= 0) {
-        store_element(reinterpret_cast<const entry_type &>(index_binary_data[index_offset[i]]));
+        auto res = store_element(reinterpret_cast<const entry_type &>(index_binary_data[index_offset[i]]));
+        assert(res != OperationStatus::timed_out);
+        if (res == OperationStatus::throttled_out) {
```
unserialize, line 177: Probably, instead of doing conditions here, it's easier to check for `MemoryStatus::NORMAL` in the end. (but I'm not sure that calling `current_memory_status()` inside this function is correct, you can skip this comment).

DrDet: fixed

server/confdata-binlog-replay.cpp
DrDet: replaced to `const char *`

server/confdata-binlog-replay.cpp
```diff
@@ -44,6 +65,40 @@ class ConfdataBinlogReplayer : vk::binlog::replayer {
...
+    if (memory_resource_->is_enough_memory_for(need_bytes)) {
+      return true;
+    }
+    log_server_critical("Not enough confdata shared memory. Processing key with first part = '%s'. %zu bytes needed by estimation for %s",
+                        processing_key_.get_first_key().c_str(), need_bytes, msg.data());
+    update_memory_status(true);
+    return false;
+  }
+
+  void raise_confdata_oom_error(const char *msg) const noexcept {
```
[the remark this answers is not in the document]
DrDet: fixed

server/confdata-binlog-replay.cpp
```diff
@@ -25,13 +28,31 @@
...
 #include "server/confdata-stats.h"
 #include "server/server-log.h"
+#include "server/statshouse/statshouse-manager.h"
 
 namespace {
 
+struct {
+  const char *binlog_mask{nullptr};
+  size_t memory_limit{2u * 1024u * 1024u * 1024u};
+  double soft_oom_threshold_ratio = CONFDATA_DEFAULT_SOFT_OOM_RATIO;
```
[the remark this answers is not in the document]
DrDet: fixed

server/confdata-stats.h
```diff
@@ -21,6 +21,7 @@ struct ConfdataStats : private vk::not_copyable {
 
   size_t total_updates{0};
   size_t ignored_updates{0};
+  size_t timed_out_updates;
```
unserialize, line 24: {0}

DrDet: done

server/confdata-binlog-replay.cpp
unserialize: I meant the following:
```
    // no changes here
    for (int i = 0; i < nrecords; i++) {
      if (index_offset[i] >= 0) {
        store_element(reinterpret_cast<const entry_type &>(index_binary_data[index_offset[i]]));
      }
    }
    blacklist_enabled_ = true;
    size_hints_.clear();

    // the only change here
    if (current_memory_status() != MemoryStatus::NORMAL) {
      raise_confdata_oom_error("Can't read confdata snapshot on start");
      return -1;
    }
    return 0;
```

unserialize (dismissed):

server/confdata-binlog-replay.cpp
DrDet: fixed

unserialize (approved):

troy4eg (approved):

PR #951, opened 2023-12-06
add curlopt_accept_encoding

astrophysik:


Danil42Russia:
> I can't find option `CURLOPT_ACCEPT_ENCODING` in php. It seems like it has another name

This is a documentation problem (it is not there), in the php sources this constant is present

https://github.com/php/php-src/blob/6f569b42fa05462c835f7389ba3af1653e5a9bef/ext/curl/curl_arginfo.h#L680

Danil42Russia (approved):

PR #956, opened 2024-01-10
Support of ML runtime using KML-format

mkornaukhov03:
## Common
Add new subsystem into runtime, it is inference of machine learning models. We have developed new format -- KML, this PR add a support of models in such a format. For now, it only supports xgboost and catboost models. 

## API
PHP side API is following:
```
function kml_xgb_predict(string $model_name, array $features): array< float >;
function kml_catboost_predict(string $model_name, array $float_feats, array $cat_feats) : array< float >;
function kml_catboost_predict_multi(string $model_name, array $float_feats, array $cat_feats) : array< array< float > >;
```
To load KML-models (.kml files), you should specify `kml-dir` option.

## Internals
Master-process iterates specified directory *recursively* and load all .kml files inside it simultaneously calculating the least sufficient amount of memory that is used to predict a model. The calculated value is used to allocate buffer of `value` size (one buffer per worker). This buffer is going to be used as arena for each call of `kml_xgboost_predict, kml_catboost_predict, kml_catboost_predict_multi`, that is why there are no heap allocation during script working time.


mkornaukhov03:
See https://github.com/VKCOM/kphp/pull/983

PR #957, opened 2024-01-10
small improvements

troy4eg:
- added s/m/h/d prefix for --time-limit option and updated docs
- removed the 2 GB upper limit for the --hard-memory-limit option


server/php-engine.cpp
```diff
@@ -1873,7 +1877,7 @@ int main_args_handler(int i, const char *long_option) {
       return vk::singleton<ServerConfig>::get().init_from_config(optarg);
     }
     case 't': {
-      script_timeout = static_cast<int>(normalize_script_timeout(atoi(optarg)));
+      script_timeout = static_cast<int>(normalize_script_timeout(parse_time_limit(optarg)));
```
DrDet, line 1880: Let's check that default specifier for `parse_time_limit` is seconds

troy4eg: I checked it out
--time-limit 5 and --time-limit 5s work the same way

DrDet (approved):

PR #959, opened 2024-01-22
fix getdate sigsegv on overflow

astrophysik:


runtime/datetime/datetime_functions.cpp
```diff
@@ -387,7 +387,11 @@ array<mixed> f$getdate(int64_t timestamp) {
   }
   tm t;
   time_t timestamp_t = timestamp;
-  localtime_r(&timestamp_t, &t);
+  tm * tp = localtime_r(&timestamp_t, &t);
+  if (tp == nullptr) {
+    php_warning("getdate overflow with timestamp %" PRId64, timestamp);
```
DrDet, line 392: let's log `errno` as well

DrDet (approved):

PR #963, opened 2024-02-01
Refactor stream_functions to C++ style

NikOsint:
In this PR `stream_functions` struct is refactored as a base class with virtual functions.

Classes derived from `stream_functions`:
- `php_stream_functions`
- `ssl_stream_functions`
- `tcp_stream_functions`
- `udp_stream_functions`
- `file_stream_functions`

Base class functions implementation just aborts KPHP - it is unexpected to call them.
Functions implementations in derived classes either do the same if corresponding protocol does not support particular function, or are valid implementations if the protocol supports it.
An `array<string, bool>` was added to the `stream_functions` struct to store supported functions and avoid throwing and handling exceptions.


NikOsint:
TODO Implement tests

PR #968, opened 2024-02-09
Fix issue with get_global_vars_memory_stats()

andreylzmw:
```
kphp_source/runtime/memory_usage.h:152:16: error: ‘struct C$KphpSpan’ has no member named ‘accept’
value.get()->accept(visitor);
pid [11216] failed or terminated : return code 1
```


DrDet (commented):
The same proplem can happen with other visitors. Let's use `DummyVisitorMethods::accept` for that

andreylzmw:
> The same proplem can happen with other visitors. Let's use `DummyVisitorMethods::accept` for that

ok, done

DrDet (approved):

PR #664, opened 2022-10-31
Allow trailing comma for function use and parameter lists

mkornaukhov03:
Previous attempt: https://github.com/VKCOM/kphp/pull/308
I didn't rebase because of merge conflict.

PR #971, opened 2024-03-04
Trailing comma for function use and parameter lists

mkornaukhov03:
See https://github.com/VKCOM/kphp/pull/664
Just change `compare` branch


compiler/gentree.cpp
```diff
@@ -97,6 +99,9 @@ VertexPtr GenTree::get_foreach_value() {
 }
 
 VertexAdaptor<op_var> GenTree::get_function_use_var_name_ref() {
+  if (cur->type() == tok_clpar) { // for trailing comma
+    return {};
+  }
```
unserialize, line 104: Seems strange, because `get_func_param()` and other list-item parsing function don't handle ')' separately

PR #975, opened 2024-04-01
Reformat old code using clang-format and add .git-blame-ignore-revs

mkornaukhov03:
* Reformat all files using `find . -name '*.cpp' -or -name '*.h' -or -name '*.inl' | xargs clang-format -i` from root project directory.
* Add .git-blame-ignore-revs. It's a black list for commits that are not desired.


mkornaukhov03:
Done in https://github.com/VKCOM/kphp/pull/1281

PR #977, opened 2024-04-04
Add support of 'random_int' and 'random_bytes' builtins

apolyakov:
This PR adds support of `random_int` and `random_bytes` functions added in `PHP8`. Also, it adds `Random\RandomException` class into runtime.

### DIFF
--- a/runtime/math_functions.cpp
+++ b/runtime/math_functions.cpp
@@ -219,11 +219,10 @@ int64_t f$getrandmax() noexcept {
   return f$mt_getrandmax();
 }
 
-int64_t f$random_int(int64_t l, int64_t r) noexcept {
+Optional<int64_t> f$random_int(int64_t l, int64_t r) noexcept {
   if (unlikely(l > r)) {
-    string errMsg{"Argument #1 ($min) must be less than or equal to argument #2 ($max)"};
-    THROW_EXCEPTION(make_throwable<C$ValueError>(string(__FILE__), __LINE__, 0, errMsg));
-    return {};
+    php_warning("Argument #1 ($min) must be less than or equal to argument #2 ($max)");
+    return false;
   }
 
   if (unlikely(l == r)) {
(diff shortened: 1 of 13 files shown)


runtime/math_functions.cpp
```diff
@@ -209,6 +219,49 @@ int64_t f$getrandmax() noexcept {
   return f$mt_getrandmax();
 }
 
+int64_t f$random_int(int64_t l, int64_t r) noexcept {
+  if (unlikely(l > r)) {
+    string errMsg{"Argument #1 ($min) must be less than or equal to argument #2 ($max)"};
+    THROW_EXCEPTION(make_throwable<C$ValueError>(string(__FILE__), __LINE__, 0, errMsg));
```
DrDet, line 225: I think kphp-way is to use `php_warning` here instead of `ValueError` exception

runtime/math_functions.cpp
```diff
@@ -209,6 +219,49 @@ int64_t f$getrandmax() noexcept {
...
+    return -1;
+  } catch (...) {
+    php_critical_error("Unhandled exception");
+  }
+}
+
+string f$random_bytes(int64_t length) noexcept {
+  if (unlikely(length < 1)) {
+    string errMsg{"Argument #1 ($length) must be greater than 0"};
+    THROW_EXCEPTION(make_throwable<C$ValueError>(string(__FILE__), __LINE__, 0, errMsg));
```
DrDet, line 250: And here the same. I think we can actually change the php signature and return `Optional<string>`
Just like here:
https://github.com/VKCOM/kphp/blob/82f3514c53acde7a1baa50f98d5f2cd31ca02a6e/runtime/openssl.cpp#L600

runtime/math_functions.cpp
```diff
@@ -209,6 +219,49 @@ int64_t f$getrandmax() noexcept {
   return f$mt_getrandmax();
 }
 
+int64_t f$random_int(int64_t l, int64_t r) noexcept {
+  if (unlikely(l > r)) {
+    string errMsg{"Argument #1 ($min) must be less than or equal to argument #2 ($max)"};
```
DrDet, line 224: we usually use snake case for variables as well

runtime/math_functions.cpp
```diff
@@ -209,6 +219,49 @@ int64_t f$getrandmax() noexcept {
...
+}
+
+string f$random_bytes(int64_t length) noexcept {
+  if (unlikely(length < 1)) {
+    string errMsg{"Argument #1 ($length) must be greater than 0"};
+    THROW_EXCEPTION(make_throwable<C$ValueError>(string(__FILE__), __LINE__, 0, errMsg));
+    return {};
+  }
+
+  string str{static_cast<string::size_type>(length), true};
```
DrDet, line 254: Actually, now it creates string with size=0, e.g. empty string for php developer

DrDet (changes requested):
Actually, `random_bytes()` always creates string with size=0, e.g. empty string for php developer

you can try run this:
```php

function foo() {
    $str = random_bytes(15);
    var_dump($str);
}

foo();
```

We can add phpt test for it, and check that actual result string length is equal to the requested one

runtime/math_functions.cpp
```diff
@@ -209,6 +219,49 @@ int64_t f$getrandmax() noexcept {
...
+    THROW_EXCEPTION(make_throwable<C$ValueError>(string(__FILE__), __LINE__, 0, errMsg));
+    return {};
+  }
+
+  if (unlikely(l == r)) {
+    return l;
+  }
+
+  try {
+    std::random_device rd{"/dev/urandom"};
```
DrDet, line 234: let's also add `#include <radnom>` above

tests/phpt/exceptions/spl/02_builtin_exceptions_1.php
DrDet, line 1: let's change the name of file :)

PR #985, opened 2024-04-22
Fix SIGSEGV in openssl_random_pseudo_bytes

apolyakov:
SIGSEGV was caused by using wrong string's constructor. The constructor we used is an optimization for single-character strings.

### DIFF
--- a/tests/phpt/openssl/9_openssl_random_pseudo_bytes.php
+++ b/tests/phpt/openssl/9_openssl_random_pseudo_bytes.php
@@ -1,26 +1,33 @@
 @ok
 <?php
 
+function is_kphp() {
+#ifndef KPHP
+    return false;
+#endif
+    return true;
+}
+
 function test_openssl_random_pseudo_bytes() {
   for ($i = -300; $i < 2048; $i += 273) {
-  /** @var string */
-  $res = "";
-#ifndef KPHP
-    try {
+    $res = "";
+
+    if (is_kphp()) {
       $res = openssl_random_pseudo_bytes($i);
-    } catch (Exception $e) {
-      if ($i < 1) {
+      if ($res === false && $i < 1) {
         print("error");
-      } else {
-        critical_error("unexpected exception");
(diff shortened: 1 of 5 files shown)


tests/phpt/openssl/9_openssl_random_pseudo_bytes.php
```diff
@@ -0,0 +1,12 @@
+@ok
+<?php
+
+function test_openssl_random_pseudo_bytes() {
```
Danil42Russia, line 4: how to use `for` and the range from negative to positive number?

DrDet: Maybe it worths doing `for` with positive numbers

DrDet (commented):
I've just made some code lookup and found absolutely the same problem with `xor_strings("1", "2");`

It seems like root cause is that cached char string optimization breaks some invariants. Especially, `string(int, char)` constructor behaves quite weiridly with cached string optimization. Sometimes it returns shared strings, sometimes not, and it breaks everything. I think we need to try fixing it

tests/phpt/openssl/9_openssl_random_pseudo_bytes.php
apolyakov: Please check the current solution

tests/phpt/string_functions/010_xor_strings.php
```diff
@@ -0,0 +1,24 @@
...
+
+function is_kphp() {
+#ifndef KPHP
+    return false;
+#endif
+    return true;
+}
+
+function test_xor_strings() {
+  $strings = ["handling and manipulating", "ings, take a look", "ctory Access Protocol", "0000000dwxasd", "1111sadsd11"];
```
DrDet, line 12: let's add a single char string to the array, because the current test succeeds even without a bug fix

DrDet (approved):

Issue #997, opened 2024-05-09
Incorrect library directories when building on RPM-based Linux distributions


lRespublica:
I'm building kphp for ALT Linux (which is an rpm-based distribution), and after building the libraries are not installed correctly.

libtlo_parsing is installed into /usr/lib directory. libvk-flex-data is installed into /usr/usr/lib (I do not user standart CMAKE_INSTALL_PREFIX. Its value is /usr, in my case.)

Both options do not meet the standards for installing software on the system.
Libraries must be installed in the lib64 directory. In this case, preferably in ${CMAKE_INSTALL_PREFIX}/lib64

As far as I understand from the source code, the names of the installation directories are simply hardcoded in the cmake files. For my build I changed these values, but doing this every time is not good practice.

This problem is relevant not only for ALT Linux, but also for any other RPM-based distribution (Fedora, RHEL, OpenSuse).

lRespublica:
PR with fix is merged.

PR #998, opened 2024-05-09
Added GNUInstallDirs CMake module

lRespublica:
The problem I'm trying to solve is described in the corresponding issue #997.

The patch replaces hardcoded installation paths with paths from GNUInstallDirs. This should help ensure cross-distribution builds, at least at the directory compatibility level.

### DIFF
--- a/common/tlo-parsing/tlo-parsing.cmake
+++ b/common/tlo-parsing/tlo-parsing.cmake
@@ -25,8 +25,8 @@ set_target_properties(tlo_parsing_static PROPERTIES
 
 install(TARGETS tlo_parsing_static
         COMPONENT TLO_PARSING_DEV
-        ARCHIVE DESTINATION lib
-        PUBLIC_HEADER DESTINATION usr/include/tlo-parsing)
+        ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
+        PUBLIC_HEADER DESTINATION /usr/${CMAKE_INSTALL_INCLUDEDIR}/tlo-parsing)
 
 vk_add_library(tlo_parsing_shared SHARED $<TARGET_OBJECTS:tlo_parsing_src>)
 set_target_properties(tlo_parsing_shared PROPERTIES OUTPUT_NAME tlo_parsing)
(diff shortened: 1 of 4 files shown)


common/tlo-parsing/tlo-parsing.cmake
```diff
@@ -25,8 +25,8 @@ set_target_properties(tlo_parsing_static PROPERTIES
 
 install(TARGETS tlo_parsing_static
         COMPONENT TLO_PARSING_DEV
-        ARCHIVE DESTINATION lib
-        PUBLIC_HEADER DESTINATION usr/include/tlo-parsing)
```
mkornaukhov03, line 29: Let's leave current behavior on debian-like systems.
```
usr/${CMAKE_INSTALL_INCLUDEDIR}/tlo-parsing
```

flex/flex.cmake
```diff
@@ -35,8 +35,8 @@ set_target_properties(flex_data_shared flex_data_static
 
 install(TARGETS flex_data_shared flex_data_static
         COMPONENT FLEX
-        LIBRARY DESTINATION usr/lib
-        ARCHIVE DESTINATION usr/lib)
```
mkornaukhov03, line 39: Same here.
```
usr/${CMAKE_INSTALL_LIBDIR}
```

mkornaukhov03 (changes requested):

common/tlo-parsing/tlo-parsing.cmake
lRespublica: I don't quite understand the reason for this decision. By specifying an absolute path, we are sure that the headers will be installed in `/usr/${CMAKE_INSTALL_INCLUDEDIR}/*`.

If we use a relative path, problems may arise if `CMAKE_INSTALL_PREFIX` is not defaulted. Some build systems replace it with cmake options.  (I'm talking specifically about [rpmbuild](https://docs.fedoraproject.org/en-US/packaging-guidelines/CMake/#_available_macros) macros, but perhaps others systems may also have similar behavior).

In this case, duplication will occur, and libraries will begin to be installed in `/usr/usr`. This doesn't seem like desirable behavior.

UPD: Now during the check I discovered that in the `install_dir/usr/usr/share/vkontakte`  "usr" directory is duplicated when `CMAKE_INSTALL_PREFIX` is set to `/usr`.

It seems to me rational to replace this part with an absolute path too. What do you think?

mkornaukhov03: OK, you're right

mkornaukhov03 (approved):

PR #1008, opened 2024-05-28
Single-stage build

mkornaukhov03:
Previously, we built runtime in a separate stage. It was just linked to transpiled and compiled code. This PR add possibility to rebuild runtime along with PHP-source code.


cmake/popular-common.cmake
```diff
@@ -36,6 +41,9 @@ endif()
 
 vk_add_library(light_common OBJECT ${LIGHT_COMMON_SOURCES})
```
astrophysik, line 42: Maybe add here something like `if (COMPILE_RUNTIME_LIGHT)`? Because it starts require libc++ to comiple project with old runtime

astrophysik (approved):

PR #1010, opened 2024-05-30
Classes in mixed

mkornaukhov03:
# This PR introduce a possibility to store objects inside a `mixed`

## Reason
For a long time in KPHP, the mixed type included only primitive types, which was a significant difference from Zend PHP. This approach was taken to ensure that classes and primitive types do not mix, thereby achieving high performance. However, this is not always convenient. Additionally, this functionality, together with ArrayAccess support, will help to decouple code built on mixed and hash tables (e.g., `mixed $user; $user['age'] => User $user; $user->age`). This PR adds the ability to explicitly convert objects to the mixed type (`to_mixed`). Access to objects in mixed, you should use smart casts:
```php
mixed $m = foo(...);
if ($m instanceof User)) {
  // use $m as if it has type User
}
```

## Implementation
Mixed layout consists of two integer values, first one for tag and the second one for the payload. In case of objects, we store pointer to an object. But there is a caveat: this pointer point to object that was casted into `may_be_mixed_base` abstract class that virtually inherited by every class in `extends`-chain and every interface in `implements`-chain. It allows to use RTTI that helps to cast an object into parent classes inside C++ (in codegen, class like `class A extends B implements I1, I2, I3` will be converted into 
```c++
class A : inherit_virtually<B, I1, I2, I3> {}
```
 ).
That's why such a snippet
```php
class A extends B implements I1, I2 { ... }
mixed $a = create(...);
if ($a instanceof B) { ... }
if ($a instanceof I1) { ... }
if ($a instanceof I2) { ... }
```
works now.


runtime/mixed.inl
```diff
@@ -44,22 +57,39 @@ void mixed::copy_from(mixed &&other) {
...
 void mixed::init_from(T &&v) {
-  auto type_and_value_ref = get_type_and_value_ptr(v);
-  type_ = type_and_value_ref.first;
-  auto *value_ptr = type_and_value_ref.second;
-  using ValueType = std::decay_t<decltype(*value_ptr)>;
-  new(value_ptr) ValueType(std::forward<T>(v));
+  if constexpr (inner_type_of_class_instance<std::decay_t<T>>::has) {
+    may_be_mixed_base * ptr = dynamic_cast<may_be_mixed_base*>(v.get());
+    if (unlikely(!ptr)) {
+      php_error("Internal compiler error. Trying to set invalid object to mixed");
```
apolyakov, line 63: It's not a compiler error

runtime/mixed.inl
```diff
@@ -551,6 +598,9 @@ int64_t mixed::to_int() const {
     case type::ARRAY:
       php_warning("Wrong conversion from array to int");
       return as_array().to_int();
+    case type::OBJECT:
+      php_warning("Wrong conversion from %s to int", get_type_or_class_name());
+      return (as_object() ? 1 : 0);
```
apolyakov, line 603: Shouldn't we just return `1` as it's specified for `intval`?
 https://www.php.net/manual/en/function.intval.php

runtime/mixed.inl
```diff
@@ -571,29 +621,57 @@ double mixed::to_float() const {
     case type::ARRAY:
       php_warning("Wrong conversion from array to float");
       return as_array().to_float();
+    case type::OBJECT: {
+      php_warning("Wrong conversion from %s to float", get_type_or_class_name());
+      return (as_object() ? 1.0 : 0.0);
```
apolyakov, line 626: Same as for conversion to `int`

runtime/mixed.inl
```diff
@@ -1579,6 +1745,10 @@ inline bool mixed::is_reference_counter(ExtraRefCnt ref_cnt_value) const noexcep
       return as_string().is_reference_counter(ref_cnt_value);
     case type::ARRAY:
       return as_array().is_reference_counter(ref_cnt_value);
+    case type::OBJECT:
+      // TODO fix later to remove const cast
+      // Have to change codegen such that get_refcnt has const qualifier
+      return const_cast<may_be_mixed_base*>(as_object())->get_refcnt() == ref_cnt_value;
```
apolyakov, line 1751: It's very dangerous to have `const_cast`. I'd prefer to get rid of it in this PR unless it takes too much efforts

runtime/mixed_decl.inl
```diff
@@ -154,6 +161,38 @@ public:
...
+  };
+  template <typename T>
+  struct inner_type_of_class_instance<class_instance<T>> {
+    static constexpr bool has = true;
+    using value_type = T;
+  };
+
+  // TODO is it ok to return pointer to mutable from const method?
+  // I need it just to pass such a pointer into class_instance. Mutability is needed because
+  // class_instance do ref-counting
```
apolyakov, line 179: Why don't we just store `class_instance` instead of a raw pointer?

apolyakov (commented):
Overall, I have two questions:

1. How hard is it to get rid of `may_be_mixed_base` class?
2. Why don't you use `class_instance` instead of a raw pointer?

runtime/comparison_operators.inl
```diff
@@ -46,6 +46,10 @@ inline bool eq2(const string &lhs, const string &rhs) {
 }
 
 inline bool eq2(const mixed &lhs, const mixed &rhs) {
+  if (lhs.is_object() || rhs.is_object()) {
+    php_warning("operators ==, != are not supported for %s and %s", lhs.get_type_or_class_name(), rhs.get_type_or_class_name());
+    return false;
```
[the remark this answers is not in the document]
mkornaukhov03: We don't know the type info of both lhs and rhs and so we don't know the layouts and that's why we cannot compare it.

runtime/mixed.inl
mkornaukhov03: Logically, it's compiler error. Compiler shouldn't build such a code, so it's kind of assert in runtime. 
Invariant is like "If compiler code written correctly then this error should never happen".

runtime/mixed.inl
mkornaukhov03: In case of null object, `intval` returns 0

runtime/mixed.inl
mkornaukhov03: For null object it should be 0.0

runtime/mixed.inl
mkornaukhov03: fixed

mkornaukhov03: Fixed

mkornaukhov03:
1. The whole this PR is based on `may_be_mixed_base` =)
I use virtual tables here to implement `is_a()` function. I suppose it's good approach. I don't know how else I could implement it.
2. `class_instance` is just a wrapper for instance of a class. Variable of type `mixed` may outlive this. Besides, it has more abstract methods than just `inc_ref/dec_ref/get_ref`, that's the only thing I need for this.

runtime/mixed.inl
apolyakov: This is what written in `intval`'s docs:
`intval() should not be used on objects, as doing so will emit an [E_WARNING](https://www.php.net/manual/en/errorfunc.constants.php#constant.e-warning) level error and return 1.`

runtime/mixed.inl
apolyakov: It's either a compiler error that happens at compile-time, or runtime error that happens in runtime. It may mislead a programmer that their program crashes in runtime with `compiler error` message.

mkornaukhov03:
@apolyakov 
I've implemented the things we discussed
1) `intrusive_ptr` instead of manual ref-counting and `class_instance`
2) get rid of `may_be_mixed_base`

runtime/class_instance_decl.inl
```diff
@@ -33,6 +33,7 @@ public:
   class_instance() = default;
   class_instance(const class_instance &) = default;
   class_instance(class_instance &&) noexcept = default;
+  class_instance(T *ptr) noexcept : o(ptr, true) {}
```
apolyakov, line 36: Is it possible to get rid of this constructor and use `class_instance<T>::create_from_base_raw_ptr`?

mkornaukhov03: Yes, sure, my fault

apolyakov (dismissed):
LGTM

apolyakov (approved):

astrophysik (approved):

DrDet:
Let's check out patterns like that a bit more carefully
https://github.com/VKCOM/kphp/blob/08e4a29839860e4283efce775c52fca2360b98d2/runtime/memory_usage.h#L52-L59

mkornaukhov03:
> Let's check out patterns like that a bit more carefully
> 
> https://github.com/VKCOM/kphp/blob/08e4a29839860e4283efce775c52fca2360b98d2/runtime/memory_usage.h#L52-L59

As I see everything all right

DrDet (approved):

PR #1011, opened 2024-06-03
Codegen globals as a linear memory piece, not as C++ variables

unserialize:
Before this MR, all constants / globals were codegenerated as separate C++ variables.

This MR splits constants and globals into different storages and pipelines and targets to **fully avoid mutable state** in codegenerated C++ code. 

The purpose of this approach is to be able to compile a script into .so, load it multiple times, and execute concurrently.

# Constants

Constants are auto-extracted variables from PHP code (const strings, const arrays, etc.).

They are initialized on master process start and are therefore accessible from all workers for reading. Every const has a special refcount: if PHP code tries to mutate it, it'll be copied to script memory. Moreover, every constant has dependency_level (`var-data.h`). For instance, an array of strings should be initialized after all strings have been initialized.

Since their count is huge, they are batched, and initialization is done (codegenerated) per-batch (more concrete, all level0 for every batch, then level1, etc.).

When codegen starts, `ConstantsBatchedMem::prepare_mem_and_assign_offsets()` is called. It splits variables into batches and calculates necessary properties used in codegen later. Note, that finally each constant is represented as an independent C++ variable (unlike mutable globals, see below, which are placed in a single memory piece). We tested different ways to use the same approach for constants also (including batching them randomly into linear memory or generating per-file / per-folder structures). It works, but either leads to lots of incremental re-compilation or inconsistent compilation times in vkcom.

See `const-globals-batched-mem.h`, `const-vars-init.cpp` and `collect-const-vars.cpp`.

# Mutable globals

While constants are initialized once in master process, mutable globals exists in each script (and are initialized on script start, placed in script memory).

Opposed to constants, mutable globals are NOT C++ variables: instead, they all are placed in a single linear memory piece (`char *g_linear_mem`), every var has offset (`VarData::offset_in_linear_mem`), and we use `(&reinterpret_cast<T*>(g_linear_mem+offset))` to access a variable (`GlobalVarInPhpGlobals`).

This memory piece is inited once on worker start, see `php-script-globals.h`.

When codegen starts, `GlobalsBatchedMem::prepare_mem_and_assign_offsets()` is called. Its does pretty the same work as for constants.

Note, that for correct offset calculation, the compiler must be aware of `sizeof()` of any possible type. If a global infers as `std::tuple<bool,int>`, earlier g++ determined its size. Now, we need to compute sizes and offsets at the moment of code generation, and to do it exactly the same as g++ would. 

Another thing to point is that we also split globals into batches, but leave "spaces" in linear memory:
```
[batch1, ...(nothing, rounded up to 1KB), batch2, ...(nothing), ...]
```
It's done to achieve less incremental re-compilation: when PHP code changes introducing a new global, offsets will be shifted only inside one batch, but not throughout the whole project.

See `const-globals-batched-mem.cpp`, `globals-vars-reset.cpp` and (runtime) `php-script-globals.h`.

# Superglobals

All superglobals, that earlier were represented as C++ variables hardcoded at runtime (`v$_SERVER`, etc.) are now moved to `PhpScriptBuiltInSuperGlobals`. No variables related to superglobal handling are left at runtime.

The only "global" C++ variable left is in `php-script-globals.cpp`. It will later be a field in a script context.

# Other changes

This MR provides all necessary changes to
* handle `require_lib` and globals within child libs at codegeration / linkage
* handle `KPHP_ENABLE_GLOBAL_VARS_MEMORY_STATS` which could be used for dev domains
* handle various corner cases, for example foreach by reference in global scope should still present a C++ variable, actually referencing somewhere to previously allocated memory
* setting correct locations for all constants (previously some locations were missing)
* and some other minor changes

# How to review this MR?

It's better to review changes as a whole, not per-commit, because even though each commit is working and indepentent, some of proposed solutions were finally rolled back or rewritten.


compiler/data/var-data.h
```diff
@@ -44,7 +44,10 @@ class VarData {
   bool marked_as_const = false;
   bool is_read_only = true;
   bool is_foreach_reference = false;
-  int dependency_level = 0;
+  bool is_builtin_runtime = false;        // $_SERVER, $argv, etc., see PhpScriptBuiltInSuperGlobals in runtime
+  int dependency_level = 0;               // for constants only (c_str$, c_arr$, etc)
```
mkornaukhov03, line 48: Why do we need to store `dependency_level` here?
Why can't we just calculate it once in codegen?

For now this class becomes very overloaded semantically

compiler/pipes/analyzer.cpp
```diff
@@ -92,6 +92,11 @@ VertexPtr CommonAnalyzerPass::on_enter_vertex(VertexPtr vertex) {
     if (var->is_constant()) {
       run_function_pass(var->init_val, this);
     }
+    if (var->is_in_global_scope()) {
+      // even if current_function->global_var_ids is not empty, has_global_vars_inside may still be unset,
+      // for example if a function declares `global $v` but not uses it, or its uses are dropped off after cfg pass
```
mkornaukhov03, line 97: But why in this case we should set this flag?

compiler/pipes/calc-empty-functions.cpp
```diff
@@ -49,6 +48,7 @@ FunctionData::body_value get_vertex_body_type(VertexPtr vertex) {
     case op_seq:
       return calc_seq_body_type(vertex.as<op_seq>());
     case op_empty:
+    case op_global:
```
mkornaukhov03, line 51: Why do we need it?

compiler/pipes/calc-locations.cpp
```diff
@@ -5,20 +5,24 @@
...
+      f.root->location = Location{current_function->file_id, current_function, f.root->location.line};
+      if (f.var->init_val) {
+        f.var->init_val.set_location_recursively(f.root->location);
+      }
     });
-    current_function->class_id->members.for_each([](ClassMemberConstant &c) {
-      stage::set_line(c.value->location.line);
-      c.value.set_location(stage::get_location());
+    current_function->class_id->members.for_each([this](ClassMemberConstant &c) {
+      c.value.set_location_recursively(Location{current_function->file_id, current_function, c.value->location.line});
```
mkornaukhov03, line 25: Why don't you use `stage::get_location`?
If it has some cons, may be better to fix them than using `Location` constructor?

compiler/pipes/gen-tree-postprocess.cpp
```diff
@@ -230,9 +230,9 @@ VertexPtr GenTreePostprocessPass::on_enter_vertex(VertexPtr root) {
 }
 
 VertexPtr GenTreePostprocessPass::on_exit_vertex(VertexPtr root) {
-  if (root->type() == op_var) {
-    if (VertexUtil::is_superglobal(root->get_string())) {
-      root->extra_type = op_ex_var_superglobal;
+  if (auto as_var = root.try_as<op_var>()) {
+    if (as_var->str_val[0] == '_' && VarData::does_name_eq_any_language_superglobal(as_var->str_val)) {
```
mkornaukhov03, line 234:
```suggestion
    if (VarData::does_name_eq_any_language_superglobal(as_var->str_val)) {
```

compiler/pipes/generate-virtual-methods.cpp
```diff
@@ -423,7 +423,7 @@ void generate_body_of_virtual_method(FunctionPtr virtual_function) {
     }
   }
   if (!cases.empty()) {
-    auto case_default_warn = generate_critical_error_call(fmt_format("call method({}) on null object", virtual_function->as_human_readable()));
+    auto case_default_warn = generate_critical_error_call(fmt_format("call method({}) on null object", virtual_function->as_human_readable(false)));
```
mkornaukhov03, line 426: Why?

mkornaukhov03 (changes requested):

runtime/php-script-globals.cpp
```diff
@@ -6,6 +6,19 @@
...
 
+// actually, g_linear_mem is allocated and never freed (since workers live forever),
+// but to prevent leak sanitizer errors, free memory manually (descructor is called of worker's stop)
+PhpScriptMutableGlobals::~PhpScriptMutableGlobals() {
+  if (g_linear_mem != nullptr) {
+    delete g_linear_mem;
+    g_linear_mem = nullptr;
+  }
+  for (const auto &[k, _] : libs_linear_mem) {
+    delete libs_linear_mem[k];
```
mkornaukhov03, line 17:
```c++
for (const auto &[_, lib_mem] : libs_linear_mem) {
  delete lib_mem;
```

compiler/pipes/generate-virtual-methods.cpp
unserialize: For codegeneration stability. "true" outputs instantiation location, which is ok for human-readable (for compilation errors, etc.), but is unstable and leads to occasional changes C++ code.

compiler/pipes/calc-locations.cpp
unserialize: Because this is the pass where we calculate locations) `stage::get_location()` works incorrectly there.

compiler/pipes/calc-empty-functions.cpp
unserialize: It was just missing. Such a file:
```
global $a;
function f() { }
```
was earlier considered as non-empty.

compiler/code-gen/const-globals-batched-mem.cpp
```diff
@@ -0,0 +1,334 @@
...
+ConstantsBatchedMem constants_batched_mem;
+GlobalsBatchedMem globals_batched_mem;
+
+int calc_sizeof_tuple_shape(const TypeData *type);
+
+[[gnu::always_inline]] inline int calc_sizeof_in_bytes_runtime(const TypeData *type) {
+  switch (type->get_real_ptype()) {
+    case tp_int:
+    case tp_float:
+      return type->use_optional() ? SIZEOF_OPTIONAL + 8 : 8;
```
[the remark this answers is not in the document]
unserialize: +

compiler/code-gen/const-globals-batched-mem.cpp
```diff
@@ -0,0 +1,334 @@
...
+      kphp_assert(!type->use_optional());
+      return SIZEOF_MIXED;
+    case tp_bool:
+      return type->use_optional() ? 2 : 1;
+    case tp_tuple:
+    case tp_shape:
+      return calc_sizeof_tuple_shape(type);
+    case tp_future:
+    case tp_future_queue:
+      return type->use_optional() ? SIZEOF_OPTIONAL + 8 : 8;
```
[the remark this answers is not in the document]
unserialize: done

compiler/code-gen/const-globals-batched-mem.cpp
```diff
@@ -0,0 +1,334 @@
...
+  } 
+}
+
+int GlobalsBatchedMem::detect_globals_batch_count(int n_globals) {
+  if (n_globals > 10000) return 256;
+  if (n_globals > 5000) return 128;
+  if (n_globals > 1000) return 32;
+  if (n_globals > 500) return 16;
+  if (n_globals > 100) return 4;
+  if (n_globals > 50) return 2;
```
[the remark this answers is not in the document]
unserialize: done

compiler/data/var-data.h
unserialize: It's calculated earlier, in collect const vars. Codegen uses already calculated values.
Yes, this class if quite complicated, but all these fields are really needed for variable description.

compiler/pipes/analyzer.cpp
unserialize: I suppose, you misunderstood the comment. I've rewritten it to be more reliable

compiler/pipes/gen-tree-postprocess.cpp
unserialize: Though this suggestion is correct, it would affect performance: all $variables would be checked, but we have a guarantee, that superglobals start from _

runtime/php-script-globals.cpp
unserialize: At first, I've written exactly the same code, but CLion (or clang-tidy?) highlights `delete lib_mem` as a warning. Maybe, it's false positive, but nevertheless, I've rewritten in that way

mkornaukhov03: > At first, I've written exactly the same code, but CLion (or clang-tidy?) highlights `delete lib_mem` as a warning. Maybe, it's false positive, but nevertheless, I've rewritten in that way

Oh, I suppose it's the problem with const- and reference-ness of structured binding.
Just `for (auto [_, lib_mem] : libs_linear_mem)` will be without warnings I believe.

compiler/data/var-data.cpp
```diff
@@ -32,13 +32,21 @@ const ClassMemberInstanceField *VarData::as_class_instance_field() const {
...
+  return superglobal_names.find(name) != superglobal_names.end();
+}
+
+bool VarData::does_name_eq_any_builtin_runtime(const std::string &name) {
+  // these vars are runtime built-ins, see PhpScriptBuiltInSuperGlobals
+  static const std::unordered_set<std::string> runtime_names = {
+    "_SERVER", "_GET", "_POST", "_ENV", "_FILES", "_COOKIE", "_REQUEST",
+    "argc", "argv", "d$PHP_SAPI"
+  };
+  return runtime_names.find(name) != runtime_names.end();
```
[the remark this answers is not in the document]
mkornaukhov03: Up to you

mkornaukhov03 (approved):

PR #1012, opened 2024-06-04
Shorten too long file names at code generation

unserialize:
In vkcom, we sometimes see "file name too long".

Provided solution is to replace `very\long\namespace\classname` with `{namespace_hash}\classname` at code generation.

We do such replacement only if a resulting filename is too long. Most filenames are left untouched.

Note, that we do this only at the moment of codegen (not at the moment of function name generation), because original (long) function names are sometimes parsed to show proper compilation errors (for instance, from `baseclass$$method$$contextclass` static=`contextclass` is parsed).


mkornaukhov03 (commented):
Why can't we do this for (almost) all files, not only very long?

compiler/pipes/code-gen.cpp
```diff
@@ -176,6 +194,25 @@ void CodeGenF::prepare_generate_function(FunctionPtr func) {
   }
 }
 
+std::string CodeGenF::shorten_occurence_of_class_in_file_name(ClassPtr occuring_class, const std::string &file_name) {
```
Danil42Russia, line 197: Don't you want to add a label at the beginning of the cpp file, with its original name? Or save it to some file.

This is information to be useful for `kphp-tools`.

unserialize: (discussed offline)

PR #1016, opened 2024-06-07
Fix compilation error for lambdas in unused functions

### LINKED ISSUE #457: Passing a closure from unused methods results in errors
```php
<?php

class Example {
  /**
   * @param callable(int):int $fn
   */
  public function test(callable $fn): int {
    return $this->test2(function () use ($fn) {
      return $fn(1);
    });
  }
  
  /**
   * @param callable():int $fn
   */
  public function test2(callable $fn): int {
    return $fn();
  }
}

$example = new Example();
$example->test2(function() { return 0; });
```
In this example, `test` is unused, it calls used method `test2()` with a closure.
If `test()` is used, there are not errors.
If `test()` itself is unused, we get this:
```
Compilation error at stage: Check classes, gen by check-classes.cpp:78
  test.php:0

var Lambda$u794a37164ccb5c4d_0::$fn is declared but never written; please, provide a default value
```
This is probably an oversight and we should fix that.

unserialize:
Sometimes, one could face the following compilation error:
```
var Lambda$ufa6f526451637135_0::$tmp_var is declared but never written; please, provide a default value
```

That happened in such a scenario:
1. A lambda with `use` capturing was in an instance function 
2. That instance function was actually unused
3. That lambda was passed to a typed callable

The reason of an error was that a parent (instance) function was unused, but a lambda itself was used (from a typed callable switch-case dispatching). And since `use` capturing actually declares a field in a lambda class, a type of that field couldn't be inferred (since a type inferer never visits its parent function, it's unused).

The provided solution is to detect such cases (after marking used/unused functions) and to manually
1. Manually delete `case` of that lambda from a typed callable dispatching
2. Manually delete that lambda from reachable functions

Of course, this looks like a hack, but as of current compiler architecture (when generating virtual methods is performed prior to calculating used call graph, and instance functions are pre-marked used apriori), nothing could be done better.

Fixes #457

### DIFF
--- a/compiler/pipes/filter-only-actually-used.cpp
+++ b/compiler/pipes/filter-only-actually-used.cpp
@@ -9,6 +9,75 @@
 #include "compiler/data/src-file.h"
 #include "compiler/compiler-core.h"
 #include "compiler/threading/profiler.h"
+#include "compiler/vertex-util.h"
+
+// having a typed callable __invoke(), which is a virtual function with switch-case dispatching,
+// replace body of `case {lambda_class_to_remove.hash()}:` (which is a lambda invoke call)
+// to just `break`; see below, why this is important
+class RemoveLambdaCallFromTypedCallablePass final : public FunctionPassBase {
+  std::string case_hash;
+
+public:
+  std::string get_description() override {
+    return "Remove lambda call from typed callable";
+  }
+
+  explicit RemoveLambdaCallFromTypedCallablePass(ClassPtr lambda_class_to_remove)
+    : case_hash(std::to_string(lambda_class_to_remove->get_hash())) {}
+
+  VertexPtr on_enter_vertex(VertexPtr root) override {
+    if (auto as_case = root.try_as<op_case>()) {
+      if (auto as_int_const = as_case->expr().try_as<op_int_const>()) {
+        if (as_int_const->str_val == case_hash) {
+          auto level1 = VertexUtil::create_int_const(1);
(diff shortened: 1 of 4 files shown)


compiler/pipes/filter-only-actually-used.cpp
```diff
@@ -9,6 +9,75 @@
 #include "compiler/data/src-file.h"
 #include "compiler/compiler-core.h"
 #include "compiler/threading/profiler.h"
+#include "compiler/vertex-util.h"
+
+// having a typed callable __invoke(), which is a virtual function with switch-case dispatching,
+// replace body of `case {lambda_class_to_remove.hash()}:` (which is a lambda invoke call)
+// to just `break`; see below, why this is important
+class RemoveLambdaCallFromTypedCallablePass final : public FunctionPassBase {
```
Danil42Russia, line 17: When calling a function lambda from a function lambda, the problem is returned:
```php
class Example {
  public function unused_function(): void {
    $tmp_var = "";
    $this->call_function_first(function() use ($tmp_var): void {
      $this->call_function_second(function() use ($tmp_var) {});
    });
  }

  /** @param callable():void $fn */
  public function call_function_first(callable $fn): void {
    $fn();
  }

  /** @param callable():void $fn */
  public function call_function_second(callable $fn): void {
    $fn();
  }
}
```

unserialize: Nice catch! BTW, the bug existed only in multi-threaded mode, because of a race condition. Fixed in a separate commit, please check

mkornaukhov03 (approved):
lgtm

PR #1019, opened 2024-06-17
Added environment to `StatsHouse` and `engine.stat`

DrDet:
Supposed to use for distinguishing instances on baremetall and different cloud environments

### DIFF
--- a/server/server-config.cpp
+++ b/server/server-config.cpp
@@ -55,6 +55,9 @@ int ServerConfig::init_from_config(const char *config_path) noexcept {
     if (auto err_msg = set_cluster_name(cluster_name.data(), false)) {
       throw std::runtime_error(err_msg);
     }
+    if (auto environment = node["environment"]) {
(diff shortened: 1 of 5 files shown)


server/server-config.cpp
```diff
@@ -55,6 +55,9 @@ int ServerConfig::init_from_config(const char *config_path) noexcept {
     if (auto err_msg = set_cluster_name(cluster_name.data(), false)) {
       throw std::runtime_error(err_msg);
     }
+    if (auto environment = node["environment"]) {
+      environment_ = environment.as<std::string>();
```
Danil42Russia, line 59: Maybe set it to some default value or something? Most likely we want to find servers for which `environment` is not specified

DrDet: We can do this by filtering by `empty` environment tag

Danil42Russia (approved):

PR #1024, opened 2024-06-20
Fix incorrect parse of integer values for double vars

soloth:
Values like `{"a": 101}` were parsed incorrectly.

They're internally interpreted as being `integer` and during assignment to `double` variable it were bit-casted from `integer` to `double`, which is wrong.

This PR fixes this behavior.


runtime/from-json-processor.h
```diff
@@ -79,7 +79,7 @@ class FromJsonVisitor {
       on_input_type_mismatch(json);
       return;
     }
-    value = json.as_double();
+    value = json.is_int() ? json.as_int() : json.as_double();
```
Danil42Russia, line 82: It looks good, but let's add tests to it?

tests/phpt/json/116_json_float_read.php
Danil42Russia, line 1: All good but:
1. Test numbers greater than 100 are negative tests.
2. Maybe move this test to `23_float_precision.php`?

mkornaukhov03: > All good but:
> 
>     1. Test numbers greater than 100 are negative tests (checking for some kind of error).
> 
>     2. Maybe move this test to `23_float_precision.php`?

I've heard a rumor that negative tests just starts with 1, not grater than 100

tests/phpt/json/37_json_float_read.php
```diff
@@ -0,0 +1,23 @@
...
+}
+
+function test_json_float_read(): void {
+  $raw = '{"amount":101}';
+
+  $res = JsonEncoder::decode($raw, MoneyRequest::class);
+  var_dump($res->getAmount());
+}
+
+test_json_float_read();
```
mkornaukhov03, line 23: Please, add a new line in the end of the file

mkornaukhov03 (changes requested):

Danil42Russia (approved):

PR #1025, opened 2024-06-24
Change exit codes

mkornaukhov03:
Introduce special exit codes for` PHP -> CPP`, `CPP -> Object files` and `Object files -> binary` stages.

Let's discuss values of exit codes and whether should we do something different from `abort()` for compiler asserts.

### DIFF
--- a/common/exit-codes.h
+++ b/common/exit-codes.h
@@ -1,0 +1,11 @@
+// Compiler for PHP (aka KPHP)
+// Copyright (c) 2024 LLC «V Kontakte»
+// Distributed under the GPL v3 License, see LICENSE.notice.txt
+
+#pragma once
+enum class ExitCode {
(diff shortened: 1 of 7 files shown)


common/exit-codes.h
```diff
@@ -0,0 +1,11 @@
+// Compiler for PHP (aka KPHP)
+// Copyright (c) 2024 LLC «V Kontakte»
+// Distributed under the GPL v3 License, see LICENSE.notice.txt
+
+#pragma once
+enum class ExitCode {
```
soloth, line 6: Please comment magic constants and logic behind it.

soloth (changes requested):

common/exit-codes.h
mkornaukhov03: Fixed

soloth (approved):

PR #1009, opened 2024-05-30
runtime light

astrophysik:

PR #1026, opened 2024-06-24
support runtime-light in compiler

astrophysik:
This pr is part of light integration https://github.com/VKCOM/kphp/pull/1009

### DIFF
--- a/compiler/compiler-settings.cpp
+++ b/compiler/compiler-settings.cpp
@@ -283,7 +283,7 @@ void CompilerSettings::init() {
   if (!no_pch.get()) {
     ss << " -Winvalid-pch -fpch-preprocess";
   }
-  if (dynamic_incremental_linkage.get()) {
+  if (mode.get() == "k2-component" ||  dynamic_incremental_linkage.get()) {
     ss << " -fPIC";
   }
   if (vk::contains(cxx.get(), "clang")) {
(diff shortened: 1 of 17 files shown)


compiler/make/make.cpp
```diff
@@ -177,6 +182,9 @@ static long long get_imported_header_mtime(const std::string &header_path, const
   return 0;
 }
 
+// todo:k2 review pch-related things for
+// k2-component mode
+
```
mkornaukhov03, line 187: Let's remove such TODOs. In the single stage build it's fully supported

compiler/compiler-settings.cpp
```diff
@@ -283,19 +290,23 @@ void CompilerSettings::init() {
...
-  if (dynamic_incremental_linkage.get()) {
+  if (mode.get() == "k2-component" ||  dynamic_incremental_linkage.get()) {
     ss << " -fPIC";
   }
   if (vk::contains(cxx.get(), "clang")) {
     ss << " -Wno-invalid-source-encoding";
+  } else {
+    if (mode.get() == "k2-component") {
+      ss << " -fcoroutines";
+    }
```
mkornaukhov03, line 301: I think we can remove it because `cxx.get()` must contain clang substring in case of `k2` mode.

compiler/compiler-settings.cpp
```diff
@@ -283,19 +290,23 @@ void CompilerSettings::init() {
...
-  #else
-    #error unsupported __cplusplus value
-  #endif
+#if __cplusplus <= 201703L
+  ss << " -std=c++17";
+#elif __cplusplus <= 202002L
+  ss << " -std=c++20";
+#else
+#error unsupported __cplusplus value
+#endif
```
mkornaukhov03, line 309: Just breakage of blame

mkornaukhov03 (changes requested):

compiler/make/make.cpp
astrophysik: remove

compiler/compiler-settings.cpp
astrophysik: fix

compiler/compiler-settings.cpp
astrophysik: fix

mkornaukhov03 (approved):
lgtm

PR #1028, opened 2024-06-25
Do not remove empty default values for string and array respectively

soloth:
Before this PR empty values are eliminated for string and arrays, so 

```php
class User {
  public string $name = '';
  public string $password = '';
}
``` 
and 
```php
class User {
  public string $name;
  public string $password;
}
```
are treated as same.

While it is true technically, this lead to unwanted consequences. Value absence marks such fields as `required` during JSON deserialization. Removed default value changes class semantics and has strong implication of `string` having default value of `''`.

Suggested changes are about storing fact user provided value was optimized and eliminated.
But the fact it was there is used by later steps to correctly process values.

### DIFF
--- a/compiler/data/kphp-json-tags.cpp
+++ b/compiler/data/kphp-json-tags.cpp
@@ -403,7 +403,7 @@ FieldJsonSettings merge_and_inherit_json_tags(const ClassMemberInstanceField &fi
 
   // for 'public int $id;' — no default and non-nullable type — set 'required' so that decode() would fire unless exists
   // it could be overridden with `@kphp-json required = false`
-  if (field.type_hint && !field.var->init_val && !does_type_hint_allow_null(field.type_hint)) {
+  if (field.type_hint && !field.var->init_val && !field.var->had_user_assigned_val && !does_type_hint_allow_null(field.type_hint)) {
     s.required = true;
   }
 
--- a/compiler/data/var-data.h
+++ b/compiler/data/var-data.h
@@ -32,6 +32,7 @@ public:
   std::string name;
   tinf::VarNode tinf_node;
   VertexPtr init_val;
+  bool had_user_assigned_val = false;
   FunctionPtr holder_func;
   ClassPtr class_id;
   std::unordered_set<VarPtr> *bad_vars = nullptr;
--- a/tests/phpt/json/38_optimized_init_val.php
+++ b/tests/phpt/json/38_optimized_init_val.php
@@ -1,0 +1,22 @@
+<?php
+
+class User {
+  public string $name = '';
+  public string $password = '';
+}
+
+function test_optimized_init_value(): void {
+  $raw = '{"name": "user"}';
+
+  $user = JsonEncoder::decode($raw, User::class);
+
+  $error = JsonEncoder::getLastError();
+  if (!empty($error)) {
+    echo "Error: " . $error;
+    exit();
+  }
+
+  var_dump(instance_to_array($user));
+}
+
+test_optimized_init_value();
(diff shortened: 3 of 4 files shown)


compiler/data/var-data.h
```diff
@@ -32,6 +32,7 @@ class VarData {
   std::string name;
   tinf::VarNode tinf_node;
   VertexPtr init_val;
+  bool had_user_assigned_val = false;
```
mkornaukhov03, line 35: Wouldn't `uninited_flag` be enough?
I had a quick look and I suppose it's what you need

compiler/data/kphp-json-tags.cpp
```diff
@@ -403,7 +403,7 @@ FieldJsonSettings merge_and_inherit_json_tags(const ClassMemberInstanceField &fi
 
   // for 'public int $id;' — no default and non-nullable type — set 'required' so that decode() would fire unless exists
   // it could be overridden with `@kphp-json required = false`
-  if (field.type_hint && !field.var->init_val && !does_type_hint_allow_null(field.type_hint)) {
+  if (field.type_hint && !field.var->init_val && !field.var->had_user_assigned_val && !does_type_hint_allow_null(field.type_hint)) {
```
mkornaukhov03, line 406: And here just
```suggestion
  if (field.type_hint && field.var->uninited_flag && !does_type_hint_allow_null(field.type_hint)) {
```
?

tests/phpt/json/38_optimized_init_val.php
```diff
@@ -0,0 +1,22 @@
+<?php
```
mkornaukhov03, line 1:
```suggestion
@ok
<?php
require_once 'kphp_tester_include.php';
```
It add some magic to enable this file for testing

mkornaukhov03 (changes requested):
Btw, why did you name `38_*` but not `37_*`?

compiler/data/var-data.h
soloth: No, it is not.
For this specific case it is important to know if user has provided value, this is the core root of a problem here.

compiler/data/kphp-json-tags.cpp
soloth: Technically speaking inited/not inited value dichotomy is not the problem. The problem is that we can't distinguish default type values (0 for int, 0.0 for float and so on) and same user-provided value.

tests/phpt/json/38_optimized_init_val.php
soloth: Thanks

compiler/data/var-data.h
mkornaukhov03: But `uninited_flag` is what we need. In always `true` if there was a provided `init_val` (even if it was optimized out) as I understand, isn't it?

soloth: I checked your proposal, no, it is not enough for this specific case. Value is wiped out and we get error. Let leave this case as is.

compiler/data/kphp-json-tags.cpp
soloth: I checked your proposal, it is not enough.
`uninited_flag` allows both `public string $password = '';` and `public string $password;` which is incorrect. Without ` = '';` there must be error.

soloth:
> Btw, why did you name `38_*` but not `37_*`?

`37_*` was added by another PR

compiler/data/kphp-json-tags.cpp
mkornaukhov03: I would change `had_user_assigned_val` into `init_val_was_optimised`, because if it wasn't optimised than this flag would be false.
But up to you

mkornaukhov03 (approved):

PR #1029, opened 2024-06-27
add fork support in runtime light

astrophysik:
Fork support in runtime light
-------------------------------------
This pr add `fork` and `wait` functions in runtime light. The main difference from forks in the master is the use of C++20 coroutines

### Overiew
The class that stores the current state of the forks is `KphpForkContext`. It contains `fork_scheduler`, which is responsible for executing forks. In the schedule function it tries to execute all possible coroutines, if the main fork runs out or there are no forks available for execution the component is blocked.

The fork stored in the `light_fork` class. It stores a pointer to the first coroutine frame and the last interrupt point. The `fork_result` type erasue class is used to store the result.


runtime-light/component/component.cpp
```diff
@@ -5,23 +5,18 @@
 #include "runtime-light/component/component.h"
 #include "runtime-light/core/globals/php-init-scripts.h"
 
-void ComponentState::resume_if_was_rescheduled() {
-  if (poll_status == PollStatus::PollReschedule) {
-    // If component was suspended by please yield and there is no awaitable streams
-    main_thread();
+bool ComponentState::is_descriptor_stream(uint64_t update_d) {
+  if (opened_descriptors.contains(update_d)) {
```
apolyakov, line 9: We can achieve the same effect with just single lookup in the map using `find`.

runtime-light/component/component.cpp
```diff
@@ -5,23 +5,18 @@
...
-}
-
-void ComponentState::resume_if_wait_stream(uint64_t stream_d, StreamStatus status) {
-  if (is_stream_timer(stream_d)) {
-    process_timer(stream_d);
-  } else {
-    process_stream(stream_d, status);
+bool ComponentState::is_descriptor_timer(uint64_t update_d) {
+  if (opened_descriptors.contains(update_d)) {
+    return opened_descriptors[update_d] == DescriptorRuntimeStatus::Timer;
```
apolyakov, line 17: Same: we can eliminate double lookup

runtime-light/core/small-object-storage.h
```diff
@@ -0,0 +1,38 @@
+#pragma once
```
apolyakov, line 1: Missing copyright header. Please, also check other files

runtime-light/core/small-object-storage.h
```diff
@@ -0,0 +1,38 @@
...
+
+#include "runtime-core/runtime-core.h"
+
+template<size_t limit>
+union small_object_storage {
+  char storage_[limit];
```
apolyakov, line 7: Don't you mind if we use `std::array` in such cases?

runtime-light/utils/timer.cpp
```diff
@@ -4,18 +4,17 @@
...
+uint64_t set_timer_without_callback(int64_t timeout_ns) {
+  const PlatformCtx & ptx = *get_platform_context();
+  ComponentState & ctx = *get_component_context();
+  uint64_t nanoseconds = timeout_ns;
   uint64_t timer_d = 0;
   SetTimerResult res = ptx.set_timer(&timer_d, nanoseconds);
   if (res != SetTimerOk) {
     php_warning("timer limit exceeded");
-    return;
+    return -1;
```
apolyakov, line 15: It'd also be great to get rid of magic constants. Let's name them :)

apolyakov (changes requested):

runtime-light/component/component.cpp
```diff
@@ -6,15 +6,17 @@
 #include "runtime-light/core/globals/php-init-scripts.h"
 
 bool ComponentState::is_descriptor_stream(uint64_t update_d) {
-  if (opened_descriptors.contains(update_d)) {
-    return opened_descriptors[update_d] == DescriptorRuntimeStatus::Stream;
+  auto it = opened_descriptors.find(update_d);
```
apolyakov, line 9: `if (const auto it = opened_descriptors.find(update_d); it != opened_descriptors.end()) ...`

Let's make no chance for errors :)

runtime-light/component/component.cpp
```diff
@@ -6,15 +6,17 @@
...
+  if (it != opened_descriptors.end()) {
+    return it->second == DescriptorRuntimeStatus::Stream;
   }
   return false;
 }
 
 bool ComponentState::is_descriptor_timer(uint64_t update_d) {
-  if (opened_descriptors.contains(update_d)) {
-    return opened_descriptors[update_d] == DescriptorRuntimeStatus::Timer;
+  auto it = opened_descriptors.find(update_d);
```
apolyakov, line 17: Same as above

runtime-light/streams/streams.h
```diff
@@ -16,6 +16,8 @@ enum class DescriptorRuntimeStatus {
   Timer
 };
 
+constexpr uint64_t BAD_PLATFORM_DESCRIPTOR = 0;
```
apolyakov, line 19: Are you sure that `0` is "bad" descriptor in the platform?

astrophysik: Yes

apolyakov (approved):
LGTM

apolyakov (dismissed):

runtime-light/coroutine/fork-scheduler.h
```diff
@@ -0,0 +1,74 @@
+// Compiler for PHP (aka KPHP)
+// Copyright (c) 2024 LLC «V Kontakte»
+// Distributed under the GPL v3 License, see LICENSE.notice.txt
+
+
+#pragma once
+
+#include "runtime-light/coroutine/light-fork.h"
```
apolyakov, line 8: I'd prefer to have explicit includes here. Such implicit includes may break in the future due to refactoring or something else

runtime-light/coroutine/runtime-future.h
```diff
@@ -0,0 +1,58 @@
+// Compiler for PHP (aka KPHP)
+// Copyright (c) 2024 LLC «V Kontakte»
+// Distributed under the GPL v3 License, see LICENSE.notice.txt
+
+#pragma once
+
+#include <cinttypes>
```
apolyakov, line 7: Let's use `<cstdint>` instead as types like `int64_t` are defined there

runtime-light/coroutine/runtime-future.h
```diff
@@ -0,0 +1,58 @@
...
+    , timeout_d(0) {}
+  stream_future()
+    : stream_d(0)
+    , timeout_d(0) {}
+};
+
+namespace std {
+template<>
+struct hash<fork_future> {
+  std::size_t operator()(const fork_future &f) const {
```
apolyakov, line 37: Let's not mix types with `std::` prefix and without it. We don't use that prefix in the project

runtime-core/memory-resource/resource_allocator.h
```diff
@@ -3,9 +3,11 @@
 // Distributed under the GPL v3 License, see LICENSE.notice.txt
 
 #pragma once
```
apolyakov, line 5: Add newline

apolyakov (changes requested):

runtime-light/coroutine/fork-scheduler.h
astrophysik: We use explicit includes everywhere in project

apolyakov: Missing includes
```
#include <coroutine>
#include <cstdint>

```

apolyakov (changes requested):

runtime-light/coroutine/fork-scheduler.h
astrophysik: You forget about
```
#include "runtime-core/memory-resource/resource_allocator.h"
#include "runtime-light/coroutine/runtime-future.h"
```

I also add them

apolyakov (approved):
LGTM!

runtime-light/coroutine/fork-scheduler.cpp
```diff
@@ -0,0 +1,237 @@
...
+  // free-associated timer
+  std::visit(
+    [this](auto &&arg) {
+      using T = std::decay_t<decltype(arg)>;
+      int64_t timer_d = 0;
+      if constexpr (std::is_same_v<T, fork_future>) {
+        timer_d = arg.timeout_d;
+      } else if constexpr (std::is_same_v<T, stream_future>) {
+        timer_d = arg.timeout_d;
+      } else {
```
apolyakov, line 27: Let's also add `static_assert(false, "non-exhaustive visitor!");` here

apolyakov (changes requested):

runtime-core/core-types/decl/optional.h
```diff
@@ -203,3 +203,17 @@ using enable_if_t_is_optional_t2 = std::enable_if_t<std::is_same<T, Optional<T2>
...
+  using type = T;
+};
+
+template<typename T>
+struct InternalOptionalType<Optional<T>> {
+  using type = T;
+};
+
+template<typename T>
+using internal_optional_type = typename InternalOptionalType<T>::type;
```
apolyakov, line 218: Let's name it `internal_optional_type_t` in a manner that standard lib names such traits

compiler/code-gen/vertex-compiler.cpp
```diff
@@ -1454,14 +1453,60 @@ void compile_function_resumable(VertexAdaptor<op_function> func_root, CodeGenera
...
+
+  FunctionPtr func = func_root->func_id;
+
+  const auto var_name_gen = [](CodeGenerator &W, VarPtr var) { W << VarName(var); };
+  //CALL FUNCTION
+  W << FunctionDeclaration(func, false) << " " << BEGIN;
+  if (func->has_global_vars_inside) {
+    W << PhpMutableGlobalsAssignCurrent() << NL;
+  }
+  W << "vk::final_action action([]{KphpForkContext::get().scheduler.mark_current_fork_as_ready();});" << NL;
```
DrDet, line 1466: Let's recheck that it's working in case when current function is called from other fork. It's a bit strange that we change codegen of function we want to fork. It should be redundant when we use C++-20 coroutines.

compiler/code-gen/vertex-compiler.cpp
```diff
@@ -1454,14 +1453,60 @@ void compile_function_resumable(VertexAdaptor<op_function> func_root, CodeGenera
...
+    auto params = func->get_params();
+    kphp_assert(!params.empty());
+    auto variadic_arg = std::prev(params.end());
+    auto name_of_variadic_param = VarName(variadic_arg->as<op_func_param>()->var()->var_id);
+    W << "if (!" << name_of_variadic_param << ".is_vector())" << BEGIN;
+    W << "php_warning(\"pass associative array(" << name_of_variadic_param << ") to variadic function: " << FunctionName(func) << "\");" << NL;
+    W << name_of_variadic_param << " = f$array_values(" << name_of_variadic_param << ");" << NL;
+    W << END << NL;
+  }
+  W << AsSeq{func_root->cmd()} << END << NL;
```
DrDet, line 1490: Also let's recheck that it's actually needed in our case

compiler/code-gen/vertex-compiler.cpp
DrDet: I think it's better to determine that fork is ready outside of its logic, i.e. somewhere in scheduler

runtime-light/coroutine/light-fork.h
```diff
@@ -0,0 +1,47 @@
+// Compiler for PHP (aka KPHP)
+// Copyright (c) 2024 LLC «V Kontakte»
+// Distributed under the GPL v3 License, see LICENSE.notice.txt
+
+#pragma once
+
```
DrDet, line 6: let's add midding includes like `coroutine`, `type_traits` etc.

runtime-light/coroutine/fork-scheduler.h
```diff
@@ -0,0 +1,78 @@
...
+  /* map from runtime_future {stream_future, fork_future} to waiting fork*/
+  unordered_map<runtime_future, int64_t> future_to_blocked_fork;
+  /* reverse of future_to_blocked_fork */
+  unordered_map<int64_t, runtime_future> fork_to_awaited_future;
+  /* timeout timer_d to waiting fork */
+  unordered_map<uint64_t, int64_t> timer_to_blocked_fork;
+  /* set of forks that wait to accept incoming stream */
+  unordered_set<int64_t> wait_incoming_query_forks;
+
+  deque<int64_t> forks_ready_to_resume;
```
DrDet, line 77: I suggest to add detailed comment about typical fork lifecycle and its transitions between these hash tables

PR #1030, opened 2024-06-27
Runtime light: RPC support

apolyakov:
This PR adds preliminary support for RPC into runtime-light.

Also, the PR fixes a bug in `unsycnhronized_memory_resource`. It prevents defragmentation from merging two memory pieces from distinct memory arenas, e.g. main memory resource and extra memory pool.


runtime-light/stdlib/rpc/rpc-api.cpp
```diff
@@ -0,0 +1,522 @@
...
+  return {};
+}
+
+bool rpc_fetch_remaining_enough(size_t len) noexcept {
+  return RpcComponentContext::get().fetch_state.remaining() >= len;
+}
+
+array<mixed> make_fetch_error(string &&error_msg, int32_t error_code) {
+  array<mixed> res;
+  res.set_value(string{"__error", 7}, std::move(error_msg));
```
astrophysik, line 55: why string(char *, size_t) instead of just string(char *)?

runtime-light/stdlib/rpc/rpc-api.cpp
```diff
@@ -0,0 +1,522 @@
...
+    buffer << static_cast<uint8_t>(string_len);
+    size_len = 1;
+  } else if (string_len <= rpc_impl_::MEDIUM_STRING_MAX_LEN) {
+    buffer << static_cast<uint8_t>(rpc_impl_::MEDIUM_STRING_MAGIC) << static_cast<uint8_t>(string_len & 0xff) << static_cast<uint8_t>((string_len >> 8) & 0xff)
+           << static_cast<uint8_t>((string_len >> 16) & 0xff);
+    size_len = 4;
+  } else {
+    buffer << static_cast<uint8_t>(rpc_impl_::LARGE_STRING_MAGIC) << static_cast<uint8_t>(string_len & 0xff) << static_cast<uint8_t>((string_len >> 8) & 0xff)
+           << static_cast<uint8_t>((string_len >> 16) & 0xff) << static_cast<uint8_t>((string_len >> 24) & 0xff);
+//    buffer << static_cast<uint8_t>(rpc_impl_::LARGE_STRING_MAGIC) << static_cast<uint8_t>(string_len & 0xff) << static_cast<uint8_t>((string_len >> 8) & 0xff)
```
astrophysik, line 363: Why this comment here? What does it mean?

runtime-light/stdlib/rpc/rpc-api.h
```diff
@@ -0,0 +1,113 @@
...
+task_t<RpcQueryInfo> typed_rpc_tl_query_one_impl(const string &actor, const RpcRequest &rpc_request, double timeout, bool collect_responses_extra_info,
+                                                 bool ignore_answer) noexcept;
+
+task_t<class_instance<C$VK$TL$RpcResponse>> typed_rpc_tl_query_result_one_impl(int64_t query_id, const RpcErrorFactory &error_factory) noexcept;
+
+} // namespace rpc_impl_
+
+// === Rpc Store ==================================================================================
+
+bool f$store_int(int64_t v) noexcept;
```
astrophysik, line 39: Why functions.txt does not include below functions?

runtime-light/stdlib/rpc/rpc-api.cpp
```diff
@@ -0,0 +1,522 @@
...
+task_t<array<mixed>> rpc_tl_query_result_one_impl(int64_t query_id) noexcept {
+  if (query_id < RPC_VALID_QUERY_ID_RANGE_START) {
+    co_return make_fetch_error(string{"wrong query_id"}, TL_ERROR_WRONG_QUERY_ID);
+  }
+
+  auto &rpc_ctx{RpcComponentContext::get()};
+  class_instance<RpcTlQuery> rpc_query{};
+  class_instance<C$ComponentQuery> component_query{};
+
+  {
```
astrophysik, line 230: Is nested scope here necessary?

cmake/init-compilation-flags.cmake
```diff
@@ -105,6 +105,9 @@ endif()
 
 add_compile_options(-Werror -Wall -Wextra -Wunused-function -Wfloat-conversion -Wno-sign-compare
                     -Wuninitialized -Wno-redundant-move -Wno-missing-field-initializers)
+if(COMPILE_RUNTIME_LIGHT)
```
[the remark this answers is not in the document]
apolyakov: I can only say that runtime-light doesn't compile without that option, at least on my machine with clang++18

compiler/code-gen/files/tl2cpp/tl2cpp.cpp
```diff
@@ -147,13 +148,20 @@ void write_tl_query_handlers(CodeGenerator &W) {
...
     FunctionSignatureGenerator(W) << "void fill_tl_storers_ht() " << BEGIN;
+    std::stringstream ss{}; // TODO: use std::transform_reduce or something like that
+    int32_t tl_storers_nums = 0;
     for (const auto &module_name : modules_with_functions) {
+      tl_storers_nums += modules[module_name].target_functions.size();
       for (const auto &f : modules[module_name].target_functions) {
-        W << "gen$tl_storers_ht.set_value(" << register_tl_const_str(f->name) << ", " << "&" << cpp_tl_struct_name("f_", f->name) << "::store, "
-          << hash_tl_const_str(f->name) << "L);" << NL;
+        ss << "gen$tl_storers_ht.set_value(" << register_tl_const_str(f->name) << ", " << "&" << cpp_tl_struct_name("f_", f->name) << "::store, "
+           << hash_tl_const_str(f->name) << "L);\n";
```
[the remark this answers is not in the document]
apolyakov: I work with `std::stringstream`, and it doesn't support `NL`

runtime-light/stdlib/rpc/rpc-api.cpp
apolyakov: Since it's more efficient. We don't need to call `strlen` when we use that constructor

runtime-light/stdlib/rpc/rpc-api.cpp
apolyakov: I'm gonna uncomment it as soon as we support large strings in TL

runtime-light/stdlib/rpc/rpc-api.h
apolyakov: We don't need these function right now. That's why I didn't add them into `functions.txt` at the moment

runtime-light/stdlib/rpc/rpc-api.cpp
apolyakov: It's as necessary as `RAII` is necessary :) Sure, we can go without that nested scope. But I use it to make sure that there will be no error when someone tries to use iterators to internal maps, e.g. `rpc_ctx.pending_rpc_queries`

astrophysik (approved):

PR #1031, opened 2024-06-28
fixed clang --asm-operand-widths warnings for aarch

troy4eg:


mkornaukhov03:
Are there any sources to read about problem?

troy4eg:
> Are there any sources to read about problem?

I didn't quite understand the question, but 

[clang docs](https://clang.llvm.org/docs/DiagnosticsReference.html#wasm-operand-widths) warning: value size does not match register size specified by the constraint and modifier

[arm instruction set reference guide](https://www.heenes.de/ro/material/arm/arm_instruction_set_reference_guide.pdf), paragraph 3.5, page 76. W is for 32-bit, X is for 64-bit.

mkornaukhov03 (approved):

PR #1034, opened 2024-07-01
fix string buffer MAX and MIN default values

astrophysik:


runtime-core/core-types/decl/string_buffer_decl.inl
```diff
@@ -63,8 +63,9 @@ public:
   friend inline bool operator!=(const string_buffer &lhs, const string_buffer &rhs);
 };
 
+
 struct string_buffer_lib_context {
-  string::size_type MIN_BUFFER_LEN;
-  string::size_type MAX_BUFFER_LEN;
+  string::size_type MIN_BUFFER_LEN = 266175;
```
apolyakov, line 68: Can you add a description about what does this number mean?)

DrDet (approved):

PR #1037, opened 2024-07-08
Make RPC in light runtime more robust

apolyakov:
Fix a bug in store_string caused by incorrect usage of string_buffer.

Add RpcBuffer class that prevents us from similar bugs in the future.

### DIFF
--- a/runtime-light/stdlib/rpc/rpc-extra-headers.cpp
+++ b/runtime-light/stdlib/rpc/rpc-extra-headers.cpp
@@ -8,7 +8,7 @@
 
 #include "common/algorithms/find.h"
 #include "common/tl/constants/common.h"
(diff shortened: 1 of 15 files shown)


runtime-light/stdlib/rpc/rpc-extra-headers.cpp
```diff
@@ -8,7 +8,7 @@
 
 #include "common/algorithms/find.h"
 #include "common/tl/constants/common.h"
-#include "runtime-light/utils/php_assert.h"
```
astrophysik, line 11: Why you change include? I Think we want to minimize count of runtime-core include

astrophysik (commented):
Add please component for rpc test

astrophysik (approved):

PR #1038, opened 2024-07-09
Add light runtime mode tests in CI

andarut:


runtime-light/header.h
Danil42Russia, line 1: And why are these changes here?

cmake/init-compilation-flags.cmake
```diff
@@ -134,9 +135,10 @@ if(COMPILE_RUNTIME_LIGHT)
...
             "${PROJECT_BINARY_DIR}/check_coroutine_include.cpp"
             COMPILE_DEFINITIONS "${TRY_COMPILE_COMPILE_OPTIONS}"
+            CXX_STANDARD 20
     )
     if(NOT HAS_COROUTINE)
         message(FATAL_ERROR "Compiler or libstdc++ does not support coroutines")
     endif()
     file(REMOVE "${PROJECT_BINARY_DIR}/check_coroutine_include.cpp")
-endif()
+endif()
```
Danil42Russia: `\n` got lost

cmake/init-compilation-flags.cmake
```diff
@@ -134,9 +135,10 @@ if(COMPILE_RUNTIME_LIGHT)
             "${PROJECT_BINARY_DIR}/tmp"
             "${PROJECT_BINARY_DIR}/check_coroutine_include.cpp"
             COMPILE_DEFINITIONS "${TRY_COMPILE_COMPILE_OPTIONS}"
+            CXX_STANDARD 20
```
Danil42Russia, line 138: It's a strange place, I would move it here:
https://github.com/VKCOM/kphp/blob/32765c1ebdd1598236a7233f424b3f02de3522d6/cmake/init-compilation-flags.cmake#L21

Danil42Russia: And at the same time, it is better to use the cmake style of specifying parameters for the property.

```cmake
set(CMAKE_CXX_STANDARD 20)
```

.github/workflows/Dockerfile.focal
```diff
@@ -6,13 +6,17 @@ COPY tests/python/requirements.txt /tmp/
 RUN apt-get update && \
     apt-get install -y --no-install-recommends apt-utils ca-certificates gnupg wget pkg-config software-properties-common && \
     wget -qO /etc/apt/trusted.gpg.d/vkpartner.asc https://artifactory-external.vkpartner.ru/artifactory/api/gpg/key/public && \
+    wget -qO - https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add - && \
+    add-apt-repository ppa:ubuntu-toolchain-r/test && \
+    echo "deb http://apt.llvm.org/focal/ llvm-toolchain-focal-18 main" >> /etc/apt/sources.list && \
+    echo "deb-src http://apt.llvm.org/focal/ llvm-toolchain-focal-18 main" >> /etc/apt/sources.list && \
```
Danil42Russia, line 12: Are we sure we need the source code?

cmake/init-compilation-flags.cmake
andreylzmw: ok, fixed

cmake/init-compilation-flags.cmake
andreylzmw: it's not a cmake global parameter, it's for try_compile function. By default cmake do not pass c++ standart variable to COMPILE_DEFINITIONS, which is passed to try_compile function

runtime-light/header.h
Danil42Russia: Like all `c++` files

.github/workflows/Dockerfile.focal
andreylzmw: what do you mean? we are not building llvm from source here. https://apt.llvm.org/ says this is the way

runtime-light/header.h
andreylzmw: to bypass clang-format, these changes provided by @astrophysik and merged to by branch for testing checking step

.github/workflows/Dockerfile.focal
andreylzmw: also `ppa:ubuntu-toolchain-r` needed to install g++-11, which is first version of gcc which supports coroutins in c++20 standard without `-fcoroutins` flag

runtime-light/header.h
Danil42Russia: It shouldn't be in this MR anyway. There should be a separate MR for `clang-format`.

compiler/compiler-settings.cpp
```diff
@@ -312,6 +312,7 @@ void CompilerSettings::init() {
     ss << " -std=c++17";
   #elif __cplusplus <= 202002L
     ss << " -std=c++20";
+    ss << " -Wno-type-limits -Wno-attributes -Wno-ignored-attributes";
```
astrophysik, line 315: Can we avoid of passing this parameters?

cmake/init-compilation-flags.cmake
```diff
@@ -106,7 +106,8 @@ endif()
 add_compile_options(-Werror -Wall -Wextra -Wunused-function -Wfloat-conversion -Wno-sign-compare
                     -Wuninitialized -Wno-redundant-move -Wno-missing-field-initializers)
 if(COMPILE_RUNTIME_LIGHT)
-    add_compile_options(-Wno-vla-cxx-extension)
+    add_compile_options(-Wno-type-limits -Wno-attributes -Wno-ignored-attributes)
```
astrophysik, line 109: Are this options necessary? Which compiler doesn't compile without them?

astrophysik: It might be better to fix the places in the code that cause the errors

vladdunchik (dismissed):

.github/workflows/Dockerfile.focal
andreylzmw: ok

cmake/init-compilation-flags.cmake
andreylzmw: ok

compiler/compiler-settings.cpp
andreylzmw: sure

.github/workflows/Build.yml
```diff
@@ -13,6 +13,17 @@ env:
   kphp_build_dir: /home/kitten/kphp/build
 
 jobs:
+  # clang-format:
```
astrophysik, line 16: You don't need that code? Why don't you delete it?

runtime-light/header.h
astrophysik: I don't think that this is important

astrophysik: > It shouldn't be in this MR anyway. There should be a separate MR for `clang-format`.

I don't think that this is important

.github/workflows/Build.yml
astrophysik: Oh. As I understand it is for clang-format check. Let's turn on it

cmake/init-compilation-flags.cmake
astrophysik: I see that you add if for GCC in this place. In my opinion this check should't exist, but it's a different task

.github/workflows/Build.yml
Danil42Russia: @astrophysik, yes, it should definitely be enabled. But let's do it in another MR.

To do it right, you have to at least do the preparation to make it work the same locally and on CI :)

andarut: removed

aleksola (approved):

aleksola (approved):

PR #1048, opened 2024-07-23
Fix CI docker caching images

andarut:
Fixing caching images by using `--cache-from` docker option and tagging them

### DIFF
--- a/.github/workflows/Build.yml
+++ b/.github/workflows/Build.yml
@@ -63,7 +63,10 @@ jobs:
         docker build -f $GITHUB_WORKSPACE/.github/workflows/Dockerfile.${{matrix.os}} $GITHUB_WORKSPACE \
               -t kphp-build-img-${{matrix.os}} \
               --cache-from=type=local,src=kphp-build-img-${{matrix.os}}-cache
-        docker tag kphp-build-img-${{matrix.os}} kphp-build-img-${{matrix.os}}-cache && mkdir -p /tmp/docker-save-${{matrix.os}} && docker save kphp-build-img-${{matrix.os}}-cache -o /tmp/docker-save-${{matrix.os}}/kphp-build-env-${{matrix.os}}.tar && ls -lh /tmp/docker-save-${{matrix.os}}
+        docker tag kphp-build-img-${{matrix.os}} kphp-build-img-${{matrix.os}}-cache \
+        mkdir -p /tmp/docker-save-${{matrix.os}} \
+        docker save kphp-build-img-${{matrix.os}}-cache -o /tmp/docker-save-${{matrix.os}}/kphp-build-env-${{matrix.os}}.tar \
+        ls -lh /tmp/docker-save-${{matrix.os}}
         
     - name: Load docker image from cache
       if: steps.docker-image-cache.outputs.cache-hit == 'true'
--- a/.github/workflows/Build.yml
+++ b/.github/workflows/Build.yml
@@ -60,7 +60,9 @@ jobs:
     - name: Build and save docker image
       if: steps.docker-image-cache.outputs.cache-hit != 'true'
       run: |
-        docker build -f $GITHUB_WORKSPACE/.github/workflows/Dockerfile.${{matrix.os}} $GITHUB_WORKSPACE -t kphp-build-img-${{matrix.os}} --cache-from=type=local,src=kphp-build-img-${{matrix.os}}-cache
+        docker build -f $GITHUB_WORKSPACE/.github/workflows/Dockerfile.${{matrix.os}} $GITHUB_WORKSPACE \
+              -t kphp-build-img-${{matrix.os}} \
+              --cache-from=type=local,src=kphp-build-img-${{matrix.os}}-cache
         docker tag kphp-build-img-${{matrix.os}} kphp-build-img-${{matrix.os}}-cache && mkdir -p /tmp/docker-save-${{matrix.os}} && docker save kphp-build-img-${{matrix.os}}-cache -o /tmp/docker-save-${{matrix.os}}/kphp-build-env-${{matrix.os}}.tar && ls -lh /tmp/docker-save-${{matrix.os}}
         
     - name: Load docker image from cache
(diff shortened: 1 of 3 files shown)


.github/workflows/Build.yml
```diff
@@ -54,22 +54,22 @@ jobs:
...
-        key: docker-image-cache-${{matrix.os}}-${{ hashFiles('.github/workflows/Dockerfile.*', 'tests/python/requirements.txt') }}
+        path: /tmp/docker-save-${{matrix.os}}
+        key: docker-save-${{matrix.os}}-${{ hashFiles('.github/workflows/Dockerfile.*', 'tests/python/requirements.txt') }}
 
     - name: Build and save docker image
       if: steps.docker-image-cache.outputs.cache-hit != 'true'
       run: |
-        docker build -f $GITHUB_WORKSPACE/.github/workflows/Dockerfile.${{matrix.os}} $GITHUB_WORKSPACE -t kphp-build-img-${{matrix.os}}
-        docker save --output kphp-build-env-${{matrix.os}}.tar kphp-build-img-${{matrix.os}}
+        docker build -f $GITHUB_WORKSPACE/.github/workflows/Dockerfile.${{matrix.os}} $GITHUB_WORKSPACE -t kphp-build-img-${{matrix.os}} --cache-from=type=local,src=kphp-build-img-${{matrix.os}}-cache
```
Danil42Russia, line 63: let's add line breaks

```sh
docker build -f $GITHUB_WORKSPACE/.github/workflows/Dockerfile.${{matrix.os}} $GITHUB_WORKSPACE \
              -t kphp-build-img-${{matrix.os}} \
              --cache-from=type=local,src=kphp-build-img-${{matrix.os}}-cache
```

.github/workflows/Build.yml
```diff
@@ -54,22 +54,22 @@ jobs:
...
+        docker tag kphp-build-img-${{matrix.os}} kphp-build-img-${{matrix.os}}-cache && mkdir -p /tmp/docker-save-${{matrix.os}} && docker save kphp-build-img-${{matrix.os}}-cache -o /tmp/docker-save-${{matrix.os}}/kphp-build-env-${{matrix.os}}.tar && ls -lh /tmp/docker-save-${{matrix.os}}
```
Danil42Russia, line 64: why is there a `&&` if you're in multiline command mode?

.github/workflows/Build.yml
```diff
@@ -54,22 +54,22 @@ jobs:
       uses: actions/cache@v3
       id: docker-image-cache
       with:
-        path: kphp-build-env-${{matrix.os}}.tar
-        key: docker-image-cache-${{matrix.os}}-${{ hashFiles('.github/workflows/Dockerfile.*', 'tests/python/requirements.txt') }}
+        path: /tmp/docker-save-${{matrix.os}}
```
Danil42Russia, line 57: why do you need an additional subfolder if it will eventually contain one tar file?

.github/workflows/Build.yml
andarut: ok, done

.github/workflows/Build.yml
andarut: fixed now

.github/workflows/Build.yml
```diff
@@ -55,21 +55,24 @@ jobs:
       id: docker-image-cache
       with:
         path: kphp-build-env-${{matrix.os}}.tar
-        key: docker-image-cache-${{matrix.os}}-${{ hashFiles('.github/workflows/Dockerfile.*', 'tests/python/requirements.txt') }}
+        key: kphp-build-env-${{matrix.os}}-${{ hashFiles('.github/workflows/Dockerfile.*', 'tests/python/requirements.txt') }}.tar
```
Danil42Russia, line 58: why does it have the `.tar` suffix?

.github/workflows/Build.yml
```diff
@@ -55,21 +55,24 @@ jobs:
...
 
     - name: Build and save docker image
       if: steps.docker-image-cache.outputs.cache-hit != 'true'
       run: |
-        docker build -f $GITHUB_WORKSPACE/.github/workflows/Dockerfile.${{matrix.os}} $GITHUB_WORKSPACE -t kphp-build-img-${{matrix.os}}
-        docker save --output kphp-build-env-${{matrix.os}}.tar kphp-build-img-${{matrix.os}}
+        docker build -f $GITHUB_WORKSPACE/.github/workflows/Dockerfile.${{matrix.os}} $GITHUB_WORKSPACE \
+              -t kphp-build-img-${{matrix.os}} \
+              --cache-from=type=local,src=kphp-build-img-${{matrix.os}}-cache \
+        docker tag kphp-build-img-${{matrix.os}} kphp-build-img-${{matrix.os}}-cache \
```
Danil42Russia, line 66: The `\` is only needed to move **ONE** command across multiple lines.

.github/workflows/Build.yml
andarut: there is no need, i was thinking that is more convinient way. backed to file now

.github/workflows/Build.yml
andarut: ok, remove

.github/workflows/Build.yml
andarut: no need, removed

.github/workflows/Build.yml
```diff
@@ -60,16 +60,19 @@ jobs:
...
+        docker save kphp-build-img-${{matrix.os}}-cache -o kphp-build-env-${{matrix.os}}.tar
         
     - name: Load docker image from cache
       if: steps.docker-image-cache.outputs.cache-hit == 'true'
       run: docker load --input kphp-build-env-${{matrix.os}}.tar
 
     - name: Start docker container
       run: |
-        docker run -dt --name kphp-build-container-${{matrix.os}} kphp-build-img-${{matrix.os}}
+        docker run -dt --name kphp-build-container-${{matrix.os}} kphp-build-img-${{matrix.os}}-cache
```
Danil42Russia, line 75: This will break the pipes on the `master`. It would be necessary to do the same in other yaml.

Danil42Russia (approved):

efazleev (approved):

PR #1054, opened 2024-08-03
add get_running_fork_id and more logs

astrophysik:


runtime-light/coroutine/awaitable.h
```diff
@@ -187,9 +206,11 @@ class wait_fork_t {
   void await_suspend(std::coroutine_handle<> coro) noexcept {
     fork_awaiter.await_suspend(coro);
     timer_awaiter.await_suspend(coro);
+    suspended_fork_id_ = ForkComponentContext::get().get_running_fork_id();
```
apolyakov, line 209: Did you try it? You'll get multiple equal logs. There will be `3` equal log lines. Let's remove this and keep just `2` :)

runtime-light/scheduler/scheduler.cpp
```diff
@@ -57,6 +57,7 @@ ScheduleStatus SimpleCoroutineScheduler::scheduleOnYield() noexcept {
 }
 
 ScheduleStatus SimpleCoroutineScheduler::schedule(ScheduleEvent::EventT event) noexcept {
+  php_debug("Scheduler: schedule event type %lu", event.index());
```
apolyakov, line 60: Unnecessary log. It's enough to keep just log in `process_platform_updates`

runtime-light/stdlib/fork/fork-api.cpp
```diff
@@ -11,6 +11,10 @@
 #include "runtime-light/coroutine/awaitable.h"
 #include "runtime-light/coroutine/task.h"
 
+int64_t f$get_running_fork_id() noexcept {
```
apolyakov, line 14: Let's define it in header, so the compiler can inline it.

P.S.: please don't add `inline` keyword in header

runtime-light/stdlib/fork/fork-context.h
```diff
@@ -17,19 +17,29 @@ class ForkComponentContext {
   template<hashable Key, typename Value>
   using unordered_map = memory_resource::stl::unordered_map<Key, Value, memory_resource::unsynchronized_pool_resource>;
 
-  static constexpr auto FORK_ID_INIT = 1;
+  static constexpr auto FORK_ID_INIT = 0;
 
   unordered_map<int64_t, task_t<fork_result>> forks_;
   int64_t next_fork_id_{FORK_ID_INIT};
+  ino64_t running_fork_id_{FORK_ID_INIT};
```
apolyakov, line 24: `int64_t`?

runtime-light/stdlib/fork/fork-context.h
```diff
@@ -17,19 +17,29 @@ class ForkComponentContext {
...
 
 public:
   explicit ForkComponentContext(memory_resource::unsynchronized_pool_resource &memory_resource) noexcept
     : forks_(unordered_map<int64_t, task_t<fork_result>>::allocator_type{memory_resource}) {}
 
   static ForkComponentContext &get() noexcept;
 
+  int64_t get_running_fork_id() const noexcept {
```
apolyakov, line 32: Actually, I don't see any benefits of having such methods. We don't have any logic that should wrap access/modify of current fork id. So let's just have a public member `current_fork_id`

runtime-light/coroutine/awaitable.h
astrophysik: Yeah, there is a bug, when awaitable is await_ready ForkComponentContext set default int64_t{} as current running. Fix it

runtime-light/stdlib/fork/fork-context.h
astrophysik: Can't agree. I don't want to add `php_debug(...)` in all places where code modify current_fork_id.

apolyakov: Your approach has several drawbacks:

1. it's a responsibility of `ForkComponentContext` to determine what fork is currently running. `ForkComponentContext` is just a storage to save current fork id and all forks that were started and not yet awaited by some other fork;
2. putting logs into `ForkComponentContext` you actually loose log accuracy. For example, you'll get unnecessary logs in case of awaitable that is `await_ready`.

astrophysik: ok

runtime-light/stdlib/fork/fork-api.h
```diff
@@ -42,6 +42,10 @@ requires(is_optional<T>::value) task_t<T> f$wait(Optional<int64_t> fork_id_opt,
   co_return co_await f$wait<T>(fork_id_opt.has_value() ? fork_id_opt.val() : INVALID_FORK_ID, timeout);
 }
 
+inline int64_t f$get_running_fork_id() noexcept {
```
apolyakov, line 45: To make it a little neater, let's visually divide code into blocks. For example, here we can easily divide it to two blocks: blocking and non-blocking API

runtime-light/coroutine/awaitable.h
apolyakov, line 1: Can you think about implementing these logs with mixins or something similar? I'd like to keep awaitables as clear and minimal as possible since they are already quite complicated.

apolyakov (changes requested):

runtime-light/coroutine/awaitable.h
```diff
@@ -47,9 +47,21 @@ namespace awaitable_impl_ {
 
 enum class State : uint8_t { Init, Suspend, Ready, End };
 
+struct awaitable_base {
```
apolyakov, line 50: Something like this is totally enough for now
```cpp
class fork_id_watcher_t {
  int64_t fork_id{ForkComponentContext::get().running_fork_id};

protected:
  void await_resume() const noexcept { ForkComponentContext::get().running_fork_id = fork_id; }
};

// usage:
class wait_for_update_t : private awaitable_impl_::fork_id_watcher_t {
  // ...
  void await_suspend() { awaitable_impl_::fork_id_watcher_t::await_resume(); ... }
};
```

runtime-light/coroutine/awaitable.h
```diff
@@ -47,21 +47,18 @@ namespace awaitable_impl_ {
...
 
-  void await_suspend_prologue() noexcept {
-    suspended_fork_id = ForkComponentContext::get().running_fork_id;
-  }
-
-  void await_resume_prologue() const noexcept {
-    ForkComponentContext::get().running_fork_id = suspended_fork_id;
-  }
+protected:
+  void await_suspend() noexcept {fork_id = ForkComponentContext::get().running_fork_id;}
```
apolyakov, line 54: No need for that. You've already got running fork ID on line 51.

apolyakov (changes requested):

PR #1058, opened 2024-08-07
Initial implementation of crypto for runtime-light

mkornaukhov03:
This PR provides initial implementation of `openssl_*` builtins. It includes:
* Implementation of `openssl_random_pseudo_bytes()` and `openssl_x509_parse()` using crypto component that is provided by K2-platform
* Initial version of updating TL-schema describing interface to interact with crypto component 
* Fix old and add new tests


runtime-light/crypto/crypto-builtins.cpp
```diff
@@ -0,0 +1,28 @@
...
+//  Copyright (c) 2024 LLC «V Kontakte»
+//  Distributed under the GPL v3 License, see LICENSE.notice.txt
+
+#include "runtime-light/crypto/crypto-builtins.h"
+
+#include "runtime-light/tl/tl-core.h"
+#include "runtime-light/streams/interface.h"
+
+// Crypto-Specific TL magics
+constexpr uint32_t TL_GET_CRYPTOSECURE_PSEUDORANDOM_BYTES = 0x2491'b81d;
```
apolyakov, line 11: Please, put in anonymous namespace. It's not gonna be used by other parts of runtime and we don't want this symbol to be exported.

runtime-light/crypto/crypto-builtins.cpp
```diff
@@ -0,0 +1,28 @@
...
+
+task_t<Optional<string>> f$openssl_random_pseudo_bytes(int64_t length) {
+  if (length <= 0 || length > string::max_size()) {
+    co_return false;
+  }
+
+  tl::TLBuffer buffer;
+  buffer.store_trivial<uint32_t>(TL_GET_CRYPTOSECURE_PSEUDORANDOM_BYTES);
+  buffer.store_trivial<int32_t>(length);
```
apolyakov, line 20: From `openssl_random_pseudo_bytes` docs: `The length of the desired string of bytes. Must be a positive integer less than or equal to 2147483647`. I think we need to check `length` here as we do in `rpc`.

runtime-light/crypto/crypto-builtins.cpp
```diff
@@ -0,0 +1,28 @@
...
+
+
+  string request_buf{};
+  request_buf.append(buffer.data(), buffer.size());
+
+  auto query = f$component_client_send_query(string("crypto"), request_buf);
```
apolyakov, line 26: Up to you: use the same initialization syntax. For example, braced init as for line 23.

runtime-light/crypto/crypto-builtins.h
```diff
@@ -0,0 +1,10 @@
+//  Compiler for PHP (aka KPHP)
+//  Copyright (c) 2024 LLC «V Kontakte»
+//  Distributed under the GPL v3 License, see LICENSE.notice.txt
+
+#pragma once
+
+#include "runtime-core/runtime-core.h"
+#include "runtime-light/coroutine/task.h"
+
+task_t<Optional<string>> f$openssl_random_pseudo_bytes(int64_t length);
```
apolyakov, line 10: Let's mark it `noexcept`.

runtime-light/crypto/crypto_schema.tl
```diff
@@ -0,0 +1,34 @@
+---types---
+
+// Boolean
+boolFalse#bc799737 = Bool;
```
apolyakov, line 4: Why do we need to define basic TL types one more time?

apolyakov (changes requested):

runtime-light/crypto/crypto-builtins.cpp
mkornaukhov03: `string::max_size() < 2147483647`. In old runtime we had the same restrictions, basically we cannot satisfy the docs fully because of current `string` implementation

runtime-light/crypto/crypto_schema.tl
mkornaukhov03: For now this TL-schema is self-sufficient, all common types are copy-pasted here. It's parsed by hands and it's easier to have the single source of truth

runtime-light/crypto/crypto-builtins.h
mkornaukhov03: `string::append` is not marked as `noexcept`, so I don't think I should

apolyakov: We will be compiling with `-fno-exception` in the future. Basically `noexcept` for us is just a note to the compiler that it can generate better code with respect to our decision not to care about exceptions.

runtime-light/crypto/crypto-builtins.cpp
```diff
@@ -19,10 +38,118 @@ task_t<Optional<string>> f$openssl_random_pseudo_bytes(int64_t length) {
...
+  for (uint32_t i = 0; i < size; ++i) {
+    const auto key_view = buffer.fetch_string();
+    if (key_view.empty()) {
+      co_return false;
+    }
+
+    string key(key_view.data(), key_view.length());
+
+    const std::optional<uint32_t> magic = buffer.fetch_trivial<uint32_t>();
+    if (!magic) {
```
apolyakov, line 99: Do you check that optional is empty or magic is not equal to zero? Why not just call `has_value` method?

runtime-light/crypto/crypto-builtins.h
mkornaukhov03: I think it's strange to mark function that call no-`noexcept` functions as `noexcept`, may be we should check all the functions once in another PR and paste `noexcept`? Most of other runtime functions that are dealing with streams are not `noexcept`, too.

runtime-light/crypto/crypto_schema.tl
mkornaukhov03: How does it makes life harder?
Which problems we are gonna to face?
It doesn't use codegen yet, so there is the same file in `k2` repo, I think we should copy-paste primitives till the merging to the one schema

runtime-light/crypto/crypto-builtins.cpp
mkornaukhov03: Previously I used `has_value`, but it seemed to me that `operator!()` is better. 
I changed auto on optional to emphasize that casting to `bool` is for `std::optional` semantic.
But `has_value()` is better, I agree now

runtime-light/crypto/crypto-builtins.h
apolyakov: They are now actually :) Also, `noexcept` does nothing except it tells the compiler it's required to generate specific code that helps in exception handling.

apolyakov (dismissed):
😞

apolyakov (approved):

PR #1060, opened 2024-08-09
Split TL and RPC stuff 

mkornaukhov03:
We may want use TL not only for RPC things, but also for communication via k2 streams. This PR is refactoring to simplify this use case

### DIFF
--- a/runtime-light/tl/tl-core.h
+++ b/runtime-light/tl/tl-core.h
@@ -1,0 +1,137 @@
+// Compiler for PHP (aka KPHP)
+// Copyright (c) 2024 LLC «V Kontakte»
+// Distributed under the GPL v3 License, see LICENSE.notice.txt
+
+#pragma once
+
+#include <cstddef>
+#include <optional>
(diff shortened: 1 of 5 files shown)


runtime-light/tl/tl-core.h
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  template<standard_layout T, standard_layout U>
+    requires std::convertible_to<U, T>
+  void store_trivial(const U &t) noexcept {
+    // Here we rely on that endianness of architecture is Little Endian
+    store_bytes(reinterpret_cast<const char *>(std::addressof(t)), sizeof(T));
+  }
+
+  void store_string(const char *str_buf, size_t str_len) noexcept;
```
apolyakov, line 83: What do you think about changing the signature: replace these two parameters with a single `std::string_view`?

apolyakov (approved):
LGTM

apolyakov (approved):

PR #1061, opened 2024-08-09
implement wait queue

astrophysik:
General
-----------

add buildin functions from runtime realted to wait fork queue
```php
function wait_queue_create (array<future<any> | false> $request_ids = []) ::: future_queue<^1[*][*]>;
function wait_queue_push (future_queue<any> &$queue_id, future<any> | false $request_ids) ::: void;
function wait_queue_empty (future_queue<any> $queue_id) ::: bool;
/** @kphp-extern-func-info interruptible */
function wait_queue_next (future_queue<any> $queue_id, $timeout ::: float = -1.0) ::: future<^1[*]> | false;
```

The storage for waiting queues is `WaitQueueContext`. That class is responsible for queues managing. Class `WaitQueue` store `tast_t`, `tast_t::awaitable_t` of awaited forks and queue of coroutine handles that was suspended on this queue.

Waiting inside `wait_queue_next` done by awaitable `wait_queue_next_t`

Details
---------
When fork is added to queue it pop associated `task_t` from `ForkComponentContext` to `WaitQueue`. This is done in such way that  to prevent the added fork from being available for waiting in other functions by `wait`. When fork become ready awaitable push it back to `ForkComponentContext`.

Also `WaitQueue` store queue of coroutine handles that suspended in `wait_queue_next_t`

Unobvious subtleties
----------------------------
1. Now fork that become ready in wait_queue has different fork id. Example:
```php
$id = fork(foo());
$queue = wait_queue_create([$id]);

$ready_fork = wait_queue_next($queue); // $ready_fork != $id
```
2. If fork is not obtained from wait_queue_next his `task_t` stored in that wait_queue, not in `ForkComponentContext`


runtime-core/memory-resource/resource_allocator.h
```diff
@@ -46,7 +46,7 @@ class resource_allocator {
   }
 
   static constexpr size_t max_value_type_size() {
-    return 128U;
+    return 256U;
```
apolyakov, line 49: Why is it changed?

runtime-light/coroutine/awaitable.h
```diff
@@ -199,3 +200,41 @@ class wait_fork_t {
...
+    , timer_awaiter(timeout_) {}
+
+  bool await_ready() const noexcept {
+    if (auto queue = WaitQueueContext::get().get_queue(queue_id); queue.has_value()) {
+      return queue.val()->empty() || queue.val()->has_ready_fork();
+    }
+    return true;
+  }
+
+  void await_suspend(std::coroutine_handle<> coro) noexcept {
```
apolyakov, line 222: There is something wrong with the design: `wait_queue_next_t` always waits for timeout before trying to check fork readiness.  So, for example, `wait_queue_next($id, 1sec)` is going to wait 1 second anyway, even if next fork is ready in 10ms.

runtime-light/coroutine/awaitable.h
```diff
@@ -199,3 +200,41 @@ class wait_fork_t {
...
+  }
+
+  Optional<int64_t> await_resume() noexcept {
+    auto queue = WaitQueueContext::get().get_queue(queue_id);
+    queue.val()->pop_coro_handle();
+    auto ready_fork = queue.val()->pop_fork();
+    if (ready_fork.has_value()) {
+      timer_awaiter.cancel();
+      // todo set here info about prev fork_id
+      return ForkComponentContext::get().push_fork(std::move(ready_fork.val().second));
```
apolyakov, line 235: Future ID should not be changed.

runtime-light/coroutine/awaitable.h
```diff
@@ -199,3 +200,41 @@ class wait_fork_t {
     }
   }
 };
+
+// ================================================================================================
+
+class wait_queue_next_t {
```
apolyakov, line 206: Will it work with job workers? Job workers are also represented as futures, so we can wait on them as well.

runtime-light/stdlib/fork/wait-queue.h
```diff
@@ -0,0 +1,96 @@
...
+#include <coroutine>
+#include <cstdint>
+
+#include "runtime-core/memory-resource/resource_allocator.h"
+#include "runtime-core/memory-resource/unsynchronized_pool_resource.h"
+#include "runtime-light/coroutine/task.h"
+#include "runtime-light/stdlib/fork/fork.h"
+#include "runtime-light/utils/concepts.h"
+
+class WaitQueue {
```
apolyakov, line 17: It looks quite strange. We should not work with raw coroutine handlers unless we develop and awaitable. In my opinion, `WaitQueue` can be completely removed. `WaitQueueContext` can be simplified a lot. Also, keep in mind that wait queues can be used to wait for job workers

runtime-light/stdlib/interface.cpp
```diff
@@ -14,6 +14,7 @@
 
 int64_t f$rand() {
```
apolyakov, line 15: Why is it defined in that file?

apolyakov (changes requested):
In general, I think that PR is based on wrong assumptions. It doesn't actually use cancellation, it's not tolerant to extensions in the future, e.g. job workers, RPC futures etc.

runtime-light/coroutine/awaitable.h
astrophysik: remove class `wait_queue_next_t`

runtime-light/coroutine/awaitable.h
astrophysik: fix

runtime-light/coroutine/awaitable.h
astrophysik: fix it, check `wait_queue_t`

runtime-light/stdlib/fork/wait-queue.h
astrophysik: merge `wait_next_t` and `WaitQueue`

PR #1067, opened 2024-08-16
Serialize inheritance

soloth:
Support serialization of inherited classes: 

```php
<?php

/**
 * @kphp-serializable
 */
class A {
  /**
   * @var int
   * @kphp-serialized-field 1
   */
  public $id = 0;
  /**
   * @var int[]
   * @kphp-serialized-field 2
   */
  public $admin_user_ids = [];
}

/**
 * @kphp-serializable
 */
class B extends A {
  /**
   * @var string
   * @kphp-serialized-field 3
   */
  public $str = "hello";
}


function demo() {
    $b = new B();
    $str = instance_serialize($b);
    $x = instance_deserialize($str, B::class);
    var_dump(instance_to_array($x));
}

demo();

```

### DIFF
--- a/compiler/code-gen/declarations.cpp
+++ b/compiler/code-gen/declarations.cpp
@@ -1151,8 +1151,8 @@ void ClassMembersDefinition::compile_msgpack_serialize(CodeGenerator &W, ClassPt
   uint16_t cnt_fields = 0;
 
   ClassPtr the_klass = klass;
-  W << "// " << the_klass->src_name << NL;
 
+  // Put class' fields along with all parent's fields.
   while (the_klass) {
     std::vector<std::string> body_inner;
     the_klass->members.for_each([&](ClassMemberInstanceField &field) {
@@ -1198,7 +1198,8 @@ void ClassMembersDefinition::compile_msgpack_deserialize(CodeGenerator &W, Class
   cases.emplace_front("default: break;");
 
   ClassPtr the_klass = klass;
-  W << "// " << the_klass->src_name << NL;
+
+  // Put class' fields along with all parent's fields.
   while (the_klass) {
     std::vector<std::string> cases_inner;
     the_klass->members.for_each([&](ClassMemberInstanceField &field) {
--- a/compiler/code-gen/declarations.cpp
+++ b/compiler/code-gen/declarations.cpp
@@ -1151,25 +1151,25 @@ void ClassMembersDefinition::compile_msgpack_serialize(CodeGenerator &W, ClassPt
   std::vector<std::string> body;
   uint16_t cnt_fields = 0;
 
+  std::vector<ClassPtr> klasses;
   ClassPtr the_klass = klass;
 
-  // Put class' fields along with all parent's fields.
   while (the_klass) {
-    std::vector<std::string> body_inner;
-    the_klass->members.for_each([&](ClassMemberInstanceField &field) {
-      if (field.serialization_tag != -1) {
-        auto func_name = fmt_format("vk::msgpack::packer_float32_decorator::pack_value{}", field.serialize_as_float32 ? "_float32" : "");
-        body_inner.emplace_back(fmt_format("packer.pack({}); {}(packer, ${});", field.serialization_tag, func_name, field.var->name));
-        cnt_fields += 2;
-      }
-    });
-    for (auto &i : body_inner) {
-      body.emplace_back(std::move(i));
-    }
+    klasses.push_back(the_klass);
     the_klass = the_klass->parent_class;
   }
 
-  std::reverse(body.begin(), body.end());
+  std::reverse(klasses.begin(), klasses.end());
+
+  for (auto &k : klasses) {
+    k->members.for_each([&](ClassMemberInstanceField &field) {
+      if (field.serialization_tag != -1) {
+        auto func_name = fmt_format("vk::msgpack::packer_float32_decorator::pack_value{}", field.serialize_as_float32 ? "_float32" : "");
+        body.emplace_back(fmt_format("packer.pack({}); {}(packer, ${});", field.serialization_tag, func_name, field.var->name));
+        cnt_fields += 2;
+      }
+    });
+  }
 
   FunctionSignatureGenerator(W).set_const_this()
     << "void " << klass->src_name << "::msgpack_pack(vk::msgpack::packer<string_buffer> &packer)" << BEGIN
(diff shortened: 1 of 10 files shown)


compiler/code-gen/declarations.cpp
```diff
@@ -1141,16 +1141,26 @@ void ClassMembersDefinition::compile_msgpack_serialize(CodeGenerator &W, ClassPt
   //   packer.pack(field_1);
   //   ...
   //}
-  std::vector<std::string> body;
+  std::list<std::string> body;
```
mkornaukhov03, line 1144: Why not a `std::deque`?

compiler/code-gen/declarations.cpp
```diff
@@ -1177,14 +1187,25 @@ void ClassMembersDefinition::compile_msgpack_deserialize(CodeGenerator &W, Class
   //}
   //
 
-  std::vector<std::string> cases;
-  klass->members.for_each([&](ClassMemberInstanceField &field) {
+  std::list<std::string> cases;
```
mkornaukhov03, line 1190: Same here

compiler/code-gen/declarations.cpp
soloth: Deque is excessive here. Moreover, I believe `std::forward_list` is cheaper, one might try.

apolyakov: Almost the only case where `std::list` is preferable over `std::vector` or `std::deque` is when you need to insert/remove elements in middle positions.

compiler/code-gen/declarations.cpp
```diff
@@ -1141,16 +1141,26 @@ void ClassMembersDefinition::compile_msgpack_serialize(CodeGenerator &W, ClassPt
...
+  while (the_klass) {
+    std::vector<std::string> body_inner;
+    the_klass->members.for_each([&](ClassMemberInstanceField &field) {
+      if (field.serialization_tag != -1) {
+        auto func_name = fmt_format("vk::msgpack::packer_float32_decorator::pack_value{}", field.serialize_as_float32 ? "_float32" : "");
+        body_inner.emplace_back(fmt_format("packer.pack({}); {}(packer, ${});", field.serialization_tag, func_name, field.var->name));
+        cnt_fields += 2;
+      }
+    });
+    for (auto it = body_inner.rbegin(); it != body_inner.rend(); ++it) {
```
apolyakov, line 1159: Is it mandatory to iterate in reverse order here?

compiler/code-gen/declarations.cpp
```diff
@@ -1141,16 +1141,26 @@ void ClassMembersDefinition::compile_msgpack_serialize(CodeGenerator &W, ClassPt
...
+      body.push_front(*it);
```
apolyakov, line 1160: As you don't further need these values, it'd be nice to move them into `body`. E.g., `body.emplace_front(std::move(*it));`

compiler/code-gen/declarations.cpp
soloth: > Is it mandatory to iterate in reverse order here?

No, it is not mandatory. But the reason to reverse values here is to save natural order of cases in resulting code. Take a look to generated code: cases are ordered naturally same as `@kphp-serialized-field` indices.
Insertion of parent's fields is done in _reverse_ order, that's why values are inserted backwards.

`std::list` (or even `forward_list`) is a good choice if you need to insert in reverse order with further usage in direct order.

apolyakov: It's not actually. It requires a memory allocation on each element addition. At the same time `std::deque` allows you to each insert/delete elements from both its ends without needing to allocate memory each time you insert an element.

Also, it's recommended to avoid explicit loops in C++. For example, here you could use, for example, `std::move` algorithm from `<algorithm>`. It makes code easier to read and could also increase performance due to heavy optimization deep inside standard library.

compiler/code-gen/declarations.cpp
soloth: > As you don't further need these values, it'd be nice to move them into `body`. E.g., `body.emplace_front(std::move(*it));`

Surprisingly code logic got broken, compilation goes successful where should fail.

compiler/code-gen/declarations.cpp
soloth: > It's not actually. It requires a memory allocation on each element addition. At the same time `std::deque` allows you to each insert/delete elements from both its ends without needing to allocate memory each time you insert an element.

Sure, you're right, but it is done in a cost of preliminary allocation. According to `std::deque` [documentation](https://en.cppreference.com/w/cpp/container/deque), it is significantly more expensive than vector:

```
On the other hand, deques typically have large minimal memory cost; 
a deque holding just one element has to allocate its full internal array 
(e.g. 8 times the object size on 64-bit libstdc++; 16 times the object size 
or 4096 bytes, whichever is larger, on 64-bit libc++).
```

In fact, we have one-way memory allocation for small amount of fields.

Most (30%) classes has no more than 6 fields, deque usage will be memory overkill.
Also, it should be noted memory is allocated (and immediately deallocated) only during code generation routines on per-class bases in `compile_msgpack_serialization` and `compile_msgpack_deserialization`, so, not think it is a problem to use `list` here. We have about 2000 classes with serialization, not that much.

soloth: > Also, it's recommended to avoid explicit loops in C++. For example, here you could use, for example, `std::move` algorithm from `<algorithm>`. It makes code easier to read and could also increase performance due to heavy optimization deep inside standard library.

I found no usage of `std::move` from `<algorithm>` in our codebase.
Don't want to be first :)

compiler/code-gen/declarations.cpp
```diff
@@ -1177,14 +1190,30 @@ void ClassMembersDefinition::compile_msgpack_deserialize(CodeGenerator &W, Class
...
 
-  cases.emplace_back("default: break;");
+    for (auto &i : cases_inner) {
+      cases.emplace_back(std::move(i));
+    }
+
+    the_klass = the_klass->parent_class;
+  }
+
+  std::reverse(cases.begin(), cases.end());
```
mkornaukhov03, line 1216: Why do we need reverse?

compiler/pipes/final-check.cpp
```diff
@@ -1063,3 +1065,49 @@ void FinalCheckPass::check_magic_clone_method(FunctionPtr fun) {
...
+
+      // Inheritance with serialization is allowed if
+      // * parent class has ho instance field
+      // * if there are instance fields, class should be marked with kphp-serializable
+      if (klass->is_serializable && !instance_fields_error_reported) {
+
+        bool hierarchy_condition = (the_klass->members.has_any_instance_var() && the_klass->is_serializable) || !the_klass->members.has_any_instance_var();
+        if (!hierarchy_condition) {
+          instance_fields_error_reported = true;
+        }
```
mkornaukhov03, line 1088: I don't understand what is `instance_fields_error_reported` for?
We can just break the loop once `hierarchy_condition == false`

compiler/pipes/final-check.h
```diff
@@ -44,4 +44,6 @@ class FinalCheckPass final : public FunctionPassBase {
   static void check_instanceof(VertexAdaptor<op_instanceof> instanceof_vertex);
   static void check_indexing(VertexPtr array, VertexPtr key) noexcept;
   static void check_array_literal(VertexAdaptor<op_array> vertex) noexcept;
+
```
mkornaukhov03, line 47: Let make it consistent

compiler/pipes/final-check.cpp
soloth: I have eliminated this check at all.

compiler/pipes/check-classes.cpp
```diff
@@ -118,6 +113,7 @@ void CheckClassesPass::check_serialized_fields(ClassPtr klass) {
       }
       f.serialization_tag = kphp_serialized_field;
       f.serialize_as_float32 = f.phpdoc->has_tag(PhpDocType::kphp_serialized_float32);
+
```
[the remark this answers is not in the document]
soloth: Done

compiler/pipes/final-check.h
soloth: Done

compiler/code-gen/declarations.cpp
soloth: I've refactored this part slightly since your review.

The reverse is required since classes are traversed in reverse order (from children to parents), but we want fields to be dumped in natural order -- from parents to children.

That's why at least one reverse is required.

compiler/code-gen/declarations.cpp
```diff
@@ -1141,16 +1141,29 @@ void ClassMembersDefinition::compile_msgpack_serialize(CodeGenerator &W, ClassPt
...
+  ClassPtr the_klass = klass;
+
+  while (the_klass) {
+    klasses.push_back(the_klass);
+    the_klass = the_klass->parent_class;
+  }
+
+  std::reverse(klasses.begin(), klasses.end());
+
+  for (auto &k : klasses) {
```
mkornaukhov03, line 1158: We can get rid of std::reverse and use `rbegin()` and `rend()`
But up to you

mkornaukhov03 (approved):
One minor note, but generally LGTM

PR #1070, opened 2024-08-19
Fix objects in mixed in light runtime

mkornaukhov03:
Fix UB with reaching `unreachable` builtin when do `var_dump()` when mixed stores object

### DIFF
--- a/runtime-light/stdlib/variable-handling.cpp
+++ b/runtime-light/stdlib/variable-handling.cpp
@@ -49,6 +49,11 @@ void do_print_r(const mixed &v, int depth) {
       *coub << shift << ")\n";
       break;
     }
+    case mixed::type::OBJECT: {
+      php_warning("print_r used on object");
+      *coub << v.as_object()->get_class();
+      break;
+    }
     default:
       __builtin_unreachable();
   }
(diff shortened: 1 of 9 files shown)


efazleev (dismissed):

runtime-light/stdlib/variable-handling.cpp
```diff
@@ -97,6 +102,12 @@ static void do_var_dump(const mixed &v, int depth) {
       *coub << shift << "}";
       break;
     }
+    case mixed::type::OBJECT: {
+      php_warning("var_dump used on object");
+      auto s = string(v.as_object()->get_class(), (string::size_type)strlen(v.as_object()->get_class()));
```
apolyakov, line 107: Let's not use C-style casts

runtime-light/stdlib/variable-handling.cpp
```diff
@@ -49,6 +49,11 @@ void do_print_r(const mixed &v, int depth) {
       *coub << shift << ")\n";
       break;
     }
+    case mixed::type::OBJECT: {
+      php_warning("print_r used on object");
+      *coub << v.as_object()->get_class();
+      break;
+    }
     default:
       __builtin_unreachable();
```
apolyakov, line 58: It's up to you, but why do we need a source of UB here? Isn't it better to replace `__builtin_unreachable` with, for example, `php_critical_error("non-exhaustive visitor")`?

apolyakov (changes requested):

PR #857, opened 2023-06-21
merging CI scripts between master and branches

Danil42Russia:

PR #1073, opened 2024-08-20
Fix docker cache DEMO

andarut:
This PR is for relocating docker cache on all build agents from Github Actions

I wrote script, so it will be rebuilding on different agents. Therefore all cache will be updated.


se-fedorov (approved):

Danil42Russia:
after #857, no longer relevant

PR #1075, opened 2024-08-21
 Fix objects in mixed in light runtime

mkornaukhov03:
Fix UB with reaching `unreachable` builtin when do var_dump() when mixed stores object and some fixes


apolyakov (approved):
LGTM

PR #1080, opened 2024-08-27
Add clang-format to CI

andarut:


pavelshulgin (dismissed):

astrophysik (commented):
Okay as for me. But let's wait approve from @Danil42Russia

runtime-light/stdlib/rpc/rpc-buffer.h
```diff
@@ -43,7 +43,7 @@ class RpcBuffer : private vk::not_copyable {
   }
 
   void reset(size_t pos) noexcept {
-    php_assert(pos >= 0 && pos <= size());
```
apolyakov, line 46: Why do you think it should be changed this way?

mkornaukhov03 (changes requested):
Am I right that it now formats with default rules? I think we should firstly decide about rules we need before merging into master

runtime-light/stdlib/rpc/rpc-buffer.h
astrophysik: As I remember clang-format gives an error that size_t can't be less than 0

Danil42Russia:
after #1283, no longer relevant

PR #1087, opened 2024-08-30
Update K2 component modes

apolyakov:
Rework K2 component modes. Previously we had two component modes: `oneshot`, `multshot`. In this PR I introduce following modes:

1. `k2-cli`: most simple mode. Semantically equivalent to regular `PHP` script;
2. `k2-server`: semantically equivalent to current `KPHP`-server;
3. `k2-oneshot`: K2 oneshot component;
4. `k2-multishot`: K2 multishot component.

### DIFF
--- a/runtime-light/component/component.cpp
+++ b/runtime-light/component/component.cpp
@@ -66,9 +66,9 @@ task_t<void> ComponentState::run_component_prologue() noexcept {
 
   component_kind_ = kind;
   if constexpr (kind == ComponentKind::CLI) {
-    output_stream_ = co_await init_kphp_cli_component();
+    standard_stream_ = co_await init_kphp_cli_component();
   } else if constexpr (kind == ComponentKind::Server) {
-    output_stream_ = co_await init_kphp_server_component();
+    standard_stream_ = co_await init_kphp_server_component();
   }
 }
 
@@ -171,8 +171,8 @@ uint64_t ComponentState::set_timer(std::chrono::nanoseconds duration) noexcept {
 }
 
 void ComponentState::release_stream(uint64_t stream_d) noexcept {
-  if (stream_d == output_stream_) {
-    output_stream_ = INVALID_PLATFORM_DESCRIPTOR;
+  if (stream_d == standard_stream_) {
+    standard_stream_ = INVALID_PLATFORM_DESCRIPTOR;
   }
   opened_streams_.erase(stream_d);
   pending_updates_.erase(stream_d); // also erase pending updates if exists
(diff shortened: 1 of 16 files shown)


runtime-light/component/component.cpp
```diff
@@ -51,12 +137,6 @@ void ComponentState::process_platform_updates() noexcept {
         }
       } else { // update on incoming stream
         php_debug("got new incoming stream %" PRIu64, stream_d);
-        if (standard_stream_ != INVALID_PLATFORM_DESCRIPTOR) {
```
astrophysik, line 54: Is there any guarantee that `k2-oneshot` component will not get second incoming stream?

runtime-light/component/component.cpp
```diff
@@ -6,18 +6,104 @@
...
+    standard_stream_ = co_await init_kphp_server_component();
+  }
+}
+
+template task_t<void> ComponentState::run_component_prologue<ComponentKind::CLI>();
+template task_t<void> ComponentState::run_component_prologue<ComponentKind::Server>();
+template task_t<void> ComponentState::run_component_prologue<ComponentKind::Oneshot>();
+template task_t<void> ComponentState::run_component_prologue<ComponentKind::Multishot>();
+
+task_t<void> ComponentState::run_component_epilogue() noexcept {
```
astrophysik, line 94: Why prologue is has compile time conditions. but epilogue has runtime?

runtime-light/component/component.cpp
```diff
@@ -6,18 +6,104 @@
...
+  } else if (magic == K2_INVOKE_JOB_WORKER_MAGIC) {
+    php_error("not implemented");
+  } else {
+    php_error("server got unexpected type of request: 0x%x", magic);
+  }
+
+  co_return stream_d;
+}
+
+int32_t merge_output_buffers(ComponentState &component_ctx) noexcept {
```
astrophysik, line 55: I think this function should be in another place. For example, output-buffer.h.

runtime-light/component/component.cpp
```diff
@@ -6,18 +6,104 @@
...
+#include "runtime-light/coroutine/task.h"
 #include "runtime-light/header.h"
 #include "runtime-light/scheduler/scheduler.h"
+#include "runtime-light/streams/streams.h"
 #include "runtime-light/utils/context.h"
+#include "runtime-light/utils/json-functions.h"
+
+namespace {
+
+constexpr uint32_t K2_INVOKE_HTTP_MAGIC = 0xd909efe8;
```
astrophysik, line 24: Maybe create one more file, that will contain init functions?

runtime-light/component/component.h
```diff
@@ -82,8 +97,9 @@ struct ComponentState {
   RegexComponentState regex_component_context;
 
 private:
-  task_t<void> main_task;
```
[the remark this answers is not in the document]
apolyakov: To get it a naming similar to other private members of ComponentContext

runtime-light/component/component.cpp
apolyakov: It's guaranteed by the K2 platform

runtime-light/component/component.cpp
apolyakov: Prologue is only called by the compiler, that's why we can use compile-time expressions. At the same time, epilogue is called not only by the compiler, but, for example, by exit functions too.

runtime-light/component/component.cpp
apolyakov: I'll do it in the next PR.

PR #1097, opened 2024-09-05
Add job workers support to K2 runtime

apolyakov:
This PR adds job workers support to K2 runtime. It has some things to think about in the future, for example, the number of available job workers.


builtin-functions/kphp-light/functions.txt
```diff
@@ -160,6 +160,53 @@ function component_server_fetch_request($query ::: ComponentQuery) ::: string;
 /** @kphp-extern-func-info interruptible */
 function component_server_send_response($query ::: ComponentQuery, $message ::: string) ::: void;
 
+// === Job Worker =================================================================================
```
astrophysik, line 163: Let's create file `job-worker.txt` with interface and include it in functions.txt

astrophysik (changes requested):

runtime-light/component/image.h
```diff
@@ -4,7 +4,6 @@
 
 #pragma once
 
-#include "runtime-light/header.h"
 #include "runtime-light/stdlib/rpc/rpc-context.h"
 
 struct ImageState {
```
astrophysik, line 9: Let's also delete `c_linear_mem` since it isn't used

runtime-light/component/component.h
```diff
@@ -63,6 +64,10 @@ struct ComponentState {
 
   task_t<void> run_component_epilogue() noexcept;
 
+  ComponentKind component_kind() const noexcept {
```
astrophysik, line 67: Is inline will be generated here automatically?

runtime-light/tl/tl-functions.h
```diff
@@ -0,0 +1,29 @@
...
+
+#include "runtime-core/runtime-core.h"
+#include "runtime-light/tl/tl-core.h"
+
+namespace tl {
+
+constexpr uint32_t K2_INVOKE_HTTP_MAGIC = 0xd909efe8;
+constexpr uint32_t K2_INVOKE_JOB_WORKER_MAGIC = 0x437d7312;
+
+struct K2InvokeJobWorker final {
```
astrophysik, line 17: Maybe rename it to make for clearer that it is request?

runtime-light/stdlib/fork/fork-context.h
```diff
@@ -12,7 +12,7 @@
 #include "runtime-light/coroutine/task.h"
 #include "runtime-light/utils/concepts.h"
 
-constexpr int64_t INVALID_FORK_ID = -1;
+inline constexpr int64_t INVALID_FORK_ID = -1;
```
astrophysik, line 15: Is inline make sense in such cases?

runtime-light/stdlib/job-worker/job-worker-api.h
```diff
@@ -0,0 +1,36 @@
...
+task_t<int64_t> f$kphp_job_worker_store_response(string response) noexcept;
+
+// === Misc =======================================================================================
+
+inline bool f$is_kphp_job_workers_enabled() noexcept {
+  return get_component_context()->component_kind() == ComponentKind::Server;
+}
+
+inline int64_t f$get_job_workers_number() noexcept {
+  return 50; // TODO
```
astrophysik, line 35: Maybe return zero or negative value?

runtime-light/stdlib/job-worker/job-worker-api.cpp
```diff
@@ -0,0 +1,153 @@
...
+#include "runtime-light/tl/tl-core.h"
+#include "runtime-light/tl/tl-functions.h"
+#include "runtime-light/tl/tl-types.h"
+#include "runtime-light/utils/context.h"
+
+namespace {
+
+constexpr const char *JOB_WORKER_COMPONENT_NAME = "_self";
+
+constexpr double MIN_TIMEOUT_S = 0.05;
```
astrophysik, line 32: `inline` like in case in fork-contex.h?

runtime-light/stdlib/job-worker/job-worker-api.cpp
```diff
@@ -0,0 +1,153 @@
...
+    php_warning("couldn't store job worker response: job workers are disabled");
+    co_return static_cast<int64_t>(JobWorkerError::store_response_incorrect_call_error);
+  } else if (jw_server_ctx.kind != JobWorkerServerComponentContext::Kind::Regular) { // we're in regular worker
+    php_warning("couldn't store job worker response: we are either in no reply job worker or not in a job worker at all");
+    co_return static_cast<int64_t>(JobWorkerError::store_response_incorrect_call_error);
+  } else if (jw_server_ctx.state == JobWorkerServerComponentContext::State::Replied) { // it's the first attempt to reply
+    php_warning("couldn't store job worker response: multiple stores are forbidden");
+    co_return static_cast<int64_t>(JobWorkerError::store_response_incorrect_call_error);
+  } else if (component_ctx.standard_stream() == INVALID_PLATFORM_DESCRIPTOR) { // we have a stream to write into
+    php_warning("couldn't store job worker response: no standard stream");
```
astrophysik, line 137: I think message `no standard stream` maybe quite unexpected for php developer. Maybe add here info that connection closed?

runtime-light/stdlib/stdlib.cmake
```diff
@@ -1,18 +1,21 @@
...
-        rpc/rpc-api.cpp
-        rpc/rpc-context.cpp
-        rpc/rpc-extra-headers.cpp
-        rpc/rpc-extra-info.cpp
-        rpc/rpc-tl-error.cpp
-        rpc/rpc-tl-query.cpp
-        rpc/rpc-tl-request.cpp
-        string/concat.cpp
-        regex/regex-functions.cpp)
+  RUNTIME_STDLIB_SRC
```
astrophysik, line 2: Why you format this file? Do you use IDE or what? My IDE format this file in different way. Let's come to a common decision

runtime-light/tl/tl-types.h
```diff
@@ -0,0 +1,25 @@
...
+#include <cstdint>
+
+#include "runtime-core/runtime-core.h"
+#include "runtime-light/tl/tl-core.h"
+
+namespace tl {
+
+constexpr uint32_t K2_JOB_WORKER_RESPONSE_MAGIC = 0x3afb3a08;
+
+struct K2JobWorkerResponse final {
```
astrophysik, line 16: Why K2JobWorkerResponse is a type, but K2InvokeJobWorker is a function?

runtime-light/tl/tl.cmake
```diff
@@ -1,4 +1,2 @@
-prepend(RUNTIME_TL_SRC tl/
-        tl-builtins.cpp
-        tl-core.cpp
-)
+prepend(RUNTIME_TL_SRC tl/ tl-builtins.cpp tl-core.cpp tl-functions.cpp
+        tl-types.cpp)
```
astrophysik, line 2: Here you use tab before tl-types.cpp, but in `runtime-light/stdlib/stdlib.cmake` don't

runtime-light/component/image.h
apolyakov: It's actually used in the compiler. If we want to remove this, let's do it in a separate PR.

runtime-light/component/component.h
apolyakov: Yep. See https://en.cppreference.com/w/cpp/language/inline

runtime-light/tl/tl-functions.h
apolyakov: It matches the name given this function in K2 platform.

runtime-light/stdlib/fork/fork-context.h
apolyakov: Yep, it guarantees that we only have one instance of that value.

runtime-light/stdlib/job-worker/job-worker-api.h
apolyakov: User code depends on that value, returning zero or negative value may lead us to a case where job workers are not working at all.

runtime-light/stdlib/job-worker/job-worker-api.cpp
apolyakov: `inline` is useless in `*.cpp` files since they are never included.

runtime-light/stdlib/job-worker/job-worker-api.cpp
apolyakov: We don't know the reason why `standard_stream` is unset, so I decided that such a general message is enough. At least for now.

runtime-light/stdlib/stdlib.cmake
apolyakov: It's a default formatting in my editor. Let's do this :)

runtime-light/tl/tl.cmake
apolyakov: Actually, there is not `tab` symbol, there are space symbols. But again, this is just a formatting used in my editor. I think we need to come up with a common decision for formatting of `cmake` files.

runtime-light/stdlib/job-worker/job-worker-api.h
astrophysik: Okay, but why was this constant chosen. Is it the average jw count on server?

astrophysik: And as soon we use k2 is any sense in this function?

runtime-light/stdlib/stdlib.cmake
astrophysik: Okay
Can you format modified files in the same way?

runtime-light/stdlib/job-worker/job-worker-api.h
apolyakov: Yep, in the future I'd like to come up with a decision made together with the platform team.

apolyakov: > Okay, but why was this constant chosen. Is it the average jw count on server?

It's just a constant that is similar to number of workers typically available in current runtime.

runtime-light/stdlib/stdlib.cmake
apolyakov: They already are.

runtime-light/tl/tl-types.h
apolyakov: Do you mean why not both of them are TL-functions?

astrophysik: I mean, that these entities have to have same type

apolyakov: `K2InvokeJobWorker` is a function cause it's assumed that it returns a result of type `K2JobWorkerResponse`. In this PR we don't always return `K2JobWorkerResponse` actually cause we do have no reply job workers. We could also return actual `K2JobWorkerResponse` even for no reply job workers, but it's useless.

astrophysik: okay, now it's clear

runtime-light/stdlib/job-worker/job-worker-api.cpp
```diff
@@ -0,0 +1,153 @@
...
+// ================================================================================================
+
+task_t<string> f$kphp_job_worker_fetch_request() noexcept {
+  if (!f$is_kphp_job_workers_enabled()) {
+    php_warning("couldn't fetch job worker request: job workers are disabled");
+    co_return string{};
+  }
+
+  auto &jw_server_ctx{JobWorkerServerComponentContext::get()};
+  if (jw_server_ctx.job_id == JOB_WORKER_INVALID_JOB_ID || jw_server_ctx.body.empty()) {
```
astrophysik, line 117: Is it good decision to rely on the string move constructor? Maybe better to set `jw_server_ctx.job_id` to `JOB_WORKER_INVALID_JOB_ID` when runtime got body?

astrophysik (dismissed):

runtime-light/stdlib/job-worker/job-worker-api.cpp
apolyakov: Hmm... Good question. Actually, the first version of `job_worker_fetch_request` used the approach you mentioned. But I decided to change it to the current one. Why? I think that `job_id` should be valid until job worker is done, at the same time it seems quite logical that we can fetch the request only once, so I change `jw_server_ctx.body` to empty string

runtime-light/tl/tl-functions.cpp
```diff
@@ -0,0 +1,48 @@
...
+#include <string_view>
+
+#include "runtime-light/tl/tl-core.h"
+
+namespace {
+
+// magic + flags + image_id + job_id + minimum string size length
+constexpr auto K2_INVOKE_JW_MIN_SIZE = 2 * sizeof(uint32_t) + sizeof(uint64_t) + sizeof(int64_t) + tl::SMALL_STRING_SIZE_LEN;
+
+constexpr auto K2_JOB_WORKER_IGNORE_ANSWER_FLAG = static_cast<uint32_t>(1U << 0U);
```
mkornaukhov03, line 17: I think it's good idea to represent TL-functions as structures.
In the spirit of this approach, it seems better to store this constants as `private static constexpr` constants

runtime-light/tl/tl-functions.cpp
```diff
@@ -0,0 +1,48 @@
...
+
+#include <cstdint>
+#include <string_view>
+
+#include "runtime-light/tl/tl-core.h"
+
+namespace {
+
+// magic + flags + image_id + job_id + minimum string size length
+constexpr auto K2_INVOKE_JW_MIN_SIZE = 2 * sizeof(uint32_t) + sizeof(uint64_t) + sizeof(int64_t) + tl::SMALL_STRING_SIZE_LEN;
```
mkornaukhov03, line 15: I suppose `sizeof(K2_INVOKE_JOB_WORKER_MAGIC) + sizeof(K2InvokeJobWorker::image_id)` is more expressive.
Imagine once you change the type of structure, `K2_INVOKE_JW_MIN_SIZE` will be recalculated automatically

runtime-light/tl/tl-types.cpp
```diff
@@ -0,0 +1,41 @@
...
+#include <cstdint>
+#include <string_view>
+#include <tuple>
+
+#include "runtime-light/tl/tl-core.h"
+
+namespace {
+
+// magic + flags + status_code + job_id + minimum string size length
+constexpr auto K2_JOB_WORKER_RESPONSE_MIN_SIZE = 2 * sizeof(uint32_t) + 2 * sizeof(int64_t) + tl::SMALL_STRING_SIZE_LEN;
```
mkornaukhov03, line 16: Same here

runtime-light/tl/tl-types.h
```diff
@@ -0,0 +1,25 @@
...
+#pragma once
+
+#include <cstdint>
+
+#include "runtime-core/runtime-core.h"
+#include "runtime-light/tl/tl-core.h"
+
+namespace tl {
+
+inline constexpr uint32_t K2_JOB_WORKER_RESPONSE_MAGIC = 0x3afb3a08;
```
mkornaukhov03, line 14: May be store it as `private static constexpr` field?

runtime-light/stdlib/job-worker/job-worker-api.cpp
```diff
@@ -0,0 +1,153 @@
...
+  }
+  if (request.empty()) {
+    php_warning("job worker request is empty");
+    co_return INVALID_FORK_ID;
+  }
+
+  // normalize timeout
+  const auto timeout_ns{
+    timeout >= MIN_TIMEOUT_S && timeout <= MAX_TIMEOUT_S
+      ? std::chrono::duration_cast<std::chrono::nanoseconds>(std::chrono::duration<double>{timeout})
```
mkornaukhov03, line 48: Why do we need this branch?
`std::clamp` does exactly this, doesn't it?

runtime-light/stdlib/job-worker/job-worker-context.h
```diff
@@ -0,0 +1,29 @@
...
+#include "runtime-light/stdlib/job-worker/job-worker.h"
+
+struct JobWorkerServerComponentContext final : private vk::not_copyable {
+  enum class Kind : uint8_t { Invalid, Regular, NoReply };
+  enum class State : uint8_t { Invalid, Working, Replied };
+
+  Kind kind{Kind::Invalid};
+  State state{State::Invalid};
+  int64_t job_id{JOB_WORKER_INVALID_JOB_ID};
+  string body;
```
mkornaukhov03, line 20: Why isn't it initialized as other fields?

mkornaukhov03 (changes requested):

runtime-light/tl/tl-functions.cpp
apolyakov: In general, it's a good idea, but for me it's good only in case of unification. For example, other TL-types like `string` should also have their constants defined in some namespace or class. Otherwise, we just add multiple ways to do the same thing

runtime-light/tl/tl-functions.cpp
apolyakov: I disagree. We shouldn't mix TL-types like `uint32_t`, `int64_t` and their c++ representation in particular case.

runtime-light/stdlib/job-worker/job-worker-context.h
apolyakov: Cause `string` is successfully default initialized here.

mkornaukhov03 (approved):

mkornaukhov03 (approved):

PR #1101, opened 2024-09-12
Add `header_register_callback` builtin support

PetrShumilov:
Add compatible with PHP 7.4.0 support of [header_register_callback](https://www.php.net/manual/en/function.header-register-callback.php) builtin.

Related issues: KPHP-1860

####  NOTE:
* After discussion with @DrDet we've decided to prohibit exceptions throwing from callback 
* `f$send_http_103_early_hints` shouldn't be a cause of callback invocation


runtime/interface.cpp
```diff
@@ -601,6 +605,10 @@ void f$fastcgi_finish_request(int64_t exit_code) {
       break;
     }
     case QUERY_TYPE_HTTP: {
+      // Run custom header handler before body processing
+      if (header_custom_handler_function) {
```
astrophysik, line 609: What if register function like this? Now it will break http response and call callback twice
```php
function foo() {
  echo "1";
  flush(); 
}
```

runtime/interface.cpp
```diff
@@ -562,7 +563,10 @@ static int ob_merge_buffers() {
 
 void f$flush() {
   php_assert(ob_cur_buffer >= 0 && php_worker.has_value());
-
+  // Run custom header handler before body processing
```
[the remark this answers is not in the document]
astrophysik: Okay, I see [here](https://www.php.net/manual/en/function.header-register-callback.php) that it should be called before any output

runtime/interface.cpp
```diff
@@ -2293,6 +2301,8 @@ static void reset_global_interface_vars(PhpScriptBuiltInSuperGlobals &supergloba
 }
 
 static void init_interface_lib() {
+  headers_sent = false;
```
astrophysik, line 2304: Why var `headers_sent` set false here?

astrophysik (changes requested):

runtime/interface.cpp
PetrShumilov: @astrophysik 
As I see it's an appropriate practice in case when some state releasing scenarios (e.g. `free_interface_lib()`) potentially may be not finished properly. 
You may see the same manipulation with `shutdown_functions_count` -- it is also double reinitializing in `init_interface_lib`. 

I can remove it if you want

PetrShumilov: @astrophysik Fixed

astrophysik (dismissed):

runtime/interface.cpp
```diff
@@ -206,6 +206,8 @@ static string http_status_line;
 static char headers_storage[sizeof(array<string>)];
 static array<string> *headers = reinterpret_cast <array<string> *> (headers_storage);
 static long long header_last_query_num = -1;
+bool headers_sent = false;
```
apolyakov, line 209: Why not `static`?

runtime/interface.cpp
```diff
@@ -206,6 +206,8 @@ static string http_status_line;
...
+headers_custom_handler_function_type headers_custom_handler_function;
```
apolyakov, line 210: Why not `static`?

apolyakov (changes requested):

runtime/interface.cpp
PetrShumilov: Fixed

runtime/interface.cpp
PetrShumilov: Fixed

astrophysik (approved):

tests/python/tests/http_server/php/index.php
```diff
@@ -119,6 +119,25 @@ public function work() {
...
+            flush();
+            sleep(2);
+            echo "Three ";
+        });
+        echo "One ";
+        return;
+     case 'flush_and_header_register_callback_callback_after_flush':
+        echo "Zero ";
+        header_register_callback(function () {
+        	echo "Unreachable ";
```
DrDet, line 135: As far as I can see it must be reachable here

DrDet (changes requested):
The main idea behind `header_register_callback` is to give opportunity to modify headers before any output is sent.
Output can be sent in the middle of script in some cases. Off the top of my head, I can remember at least 2 such cases:
- `f$flush`
There's a test for that case, but for some reason it's not correct
- `f$send_http_103_early_hints`

Now these cases are skipped, and we invoke callback only in the end of the script. Now it behaves just like `register_shutdown_function` and it's incorrect

For example: 
```php
<?php

function main() {
    header_register_callback(function() {
        fprintf(STDERR, "header_callback invoked\n");
        header("Qwerty: hello");
    });
    fprintf(STDERR, "before flush()\n");
    echo "111111111\n";
    flush();
    echo "222222222\n";
    fprintf(STDERR, "after flush()\n");
}

main();

```
Expected behaviour here is to have `Qwerty: hello` response header, but that's not really the case.

Let's do the following:
- find and fix all skipped cases like that
- add test with callback, which sends some RPC to check `swapcontext` correctness

tests/python/tests/http_server/php/index.php
PetrShumilov: @DrDet 
Yes, I agree, fixed

runtime/interface.cpp
```diff
@@ -601,6 +607,20 @@ void f$fastcgi_finish_request(int64_t exit_code) {
       break;
     }
     case QUERY_TYPE_HTTP: {
+      // Run custom headers handler before body processing
+      if (headers_custom_handler_function && !headers_sent) {
+        headers_sent = true;
+        headers_custom_handler_function();
+        // In `headers_custom_handler_function` might be happened `f$flush` invocation
+        // First of all, we have to accumulate response data from buffers
+        ob_total_buffer = ob_merge_buffers();
```
DrDet, line 616: here we call `ob_merge_buffers()` twice, even on usual execution path, is it idempotent?

runtime/interface.cpp
```diff
@@ -601,6 +607,20 @@ void f$fastcgi_finish_request(int64_t exit_code) {
...
+        // If a flushing is happened, send response without headers
+        if (php_worker.has_value() && php_worker->flushed_http_connection) {
+          string const raw_response = oub[ob_total_buffer].str();
+          http_set_result(nullptr, 0, raw_response.c_str(), raw_response.size(), static_cast<int32_t>(exit_code));
+          php_assert (0);
+        }
```
DrDet, line 622: Also I can't understand why this logic was changed.
As far as I can see, If flush was not invoked from header callback, we just go by usual path without changes.
If flush was invoked, it means header callback was invoked too, and headers already was sent, and then we go by usual old path too. Do I miss something?

compiler/pipes/final-check.cpp
```diff
@@ -318,20 +318,31 @@ static void check_kphp_tracing_func_enter_branch_call(FunctionPtr current_functi
...
+                               "But it throws {}\n"
+                               "Throw chain: {}",
+                               where.c_str(), vk::join(throws, ", "), callback->func_id->get_throws_call_chain()));
+}
+
 void check_register_shutdown_functions(VertexAdaptor<op_func_call> call) {
   auto callback = call->args()[0].as<op_callback_of_builtin>();
   if (!callback->func_id->can_throw()) {
     return;
   }
```
DrDet, line 336: Maybe move it inside `raise_error_if_throwable` to fit its name? Now this function always raises error assuming to take throwable callback

runtime/interface.cpp
PetrShumilov: Invocation moved into more appropriate place, fixed

runtime/interface.cpp
PetrShumilov: Callback invocation moved into more appropriate place, fixed

compiler/pipes/final-check.cpp
PetrShumilov: Fixed

DrDet (approved):

apolyakov (approved):

PR #1103, opened 2024-09-13
K2: add support for asynchronous callbacks

apolyakov:


tests/k2-components/test_async_builtins.php
```diff
@@ -0,0 +1,31 @@
+<?php
```
mkornaukhov03, line 1: Add here `@ok` to count it as a test

compiler/code-gen/vertex-compiler.cpp
```diff
@@ -1991,25 +1991,36 @@ void compile_callback_of_builtin(VertexAdaptor<op_callback_of_builtin> root, Cod
    * Will be transformed to:
    *   array_map([captured1 = $extern] (auto &&... args) {
    *       return lambda$xxx(captured1, std::forward<decltype(args)>(args)...);
-   *   }), const_array);
+   *   }, const_array);
+   * or in K2 mode:
+   *   co_await array_map([captured1 = $extern] (auto &&... args) [[-> task_t<lambda's return type>]]? {
+   *       [[return lambda$xxx(captured1, std::forward<decltype(args)>(args)...);
+   *       | co_return(co_await lambda$xxx(captured1, std::forward<decltype(args)>(args)...))]]
```
mkornaukhov03, line 1998: Why `[[, ]]`?
As I see output, there are single `[, ]`

compiler/code-gen/vertex-compiler.cpp
```diff
@@ -1991,25 +1991,36 @@ void compile_callback_of_builtin(VertexAdaptor<op_callback_of_builtin> root, Cod
...
   W << LockComments{};
 
   int idx = 0;
-  W << "[" << JoinValues(*root, ", ", join_mode::one_line, [&idx](CodeGenerator &W, VertexPtr cpp_captured) {
-    W << "captured" << (++idx) << " = " << cpp_captured;
-  }) << "]";
+  W << "["
+    << JoinValues(*root, ", ", join_mode::one_line, [&idx](CodeGenerator &W, VertexPtr cpp_captured) { W << "captured" << (++idx) << " = " << cpp_captured; })
+    << "]";
+
```
mkornaukhov03, line 2010: Let's leave it as is, to have a valid blame

runtime-light/stdlib/array/array-functions.h
```diff
@@ -60,9 +65,17 @@ array<T> f$array_filter_by_key(const array<T> &a, const T1 &callback) noexcept {
...
+task_t<array<R>> f$array_map(F f, array<A> arr) {
+  array<R> result{arr.size()};
+  for (const auto &it : arr) {
+    if constexpr (is_async_function_v<F, A>) {
+      result.set_value(it.get_key(), co_await std::invoke(f, it.get_value()));
+    } else {
+      result.set_value(it.get_key(), std::invoke(f, it.get_value()));
+    }
+  }
+  co_return result;
```
mkornaukhov03, line 78: But it forces `array_map` to be always async, isn't it?
Do we really want it to be async?

mkornaukhov03 (changes requested):

tests/k2-components/test_async_builtins.php
apolyakov: k2-component tests do not belong to any test system yet. As you may see, none of existing component tests have that tag

compiler/code-gen/vertex-compiler.cpp
apolyakov: I tried to use it as a syntax for block of code that support optionality via `?` and disjunction via `|`.

compiler/code-gen/vertex-compiler.cpp
apolyakov: Yep, fixed

runtime-light/stdlib/array/array-functions.h
apolyakov: Yes, `array_map` is async now according to its signature. So, in case of async callback it may stop its execution. And in case of a sync callback it will execute synchronously.

compiler/code-gen/vertex-compiler.cpp
mkornaukhov03: It's worse to read, I've spent about 5 mins to parse this syntax :smiley: 
Let's use just English `or` instead of disjunction

runtime-light/stdlib/array/array-functions.h
mkornaukhov03: But in any case it would be coroutine. Is it intended?

compiler/code-gen/vertex-compiler.cpp
apolyakov: It's hard to express this in English and keep comment small. So I used Antlr-like syntax :)

runtime-light/stdlib/array/array-functions.h
apolyakov: Yep, it is

mkornaukhov03: But it means we need extra allocation for coroutine frame each time, isn't it?
It's OK for `array_map` as it allocates result inside, but let's see at `array_reduce`.
It doesn't allocate extra memory, but if we gonna implement it like that, than we would have extra useless allocations. Is it possible to have a template overload such that returns `T` instead of `task_t<T>`?
If yes, how hard to implement this?

If it's hard, we should leave a comment about it at least

compiler/code-gen/vertex-compiler.cpp
mkornaukhov03: What `(-> task_t<lambda's return type>)?` stands for?

apolyakov: It means that it may have an explicit return type optionally

runtime-light/stdlib/array/array-functions.h
apolyakov: In general, we can't overload a function based on its return type. But it can be done with some hacks. I think it's better to have a clean implementation and apply those hacks only if we see that we have a performance gap

mkornaukhov03: After discussion we've decided to leave a comment

mkornaukhov03 (approved):

PR #1104, opened 2024-09-16
Refactor crypto for light runtime

mkornaukhov03:
Make TL-parsing logic as for JobWorkers

### DIFF
--- a/runtime-light/stdlib/crypto/crypto-functions.cpp
+++ b/runtime-light/stdlib/crypto/crypto-functions.cpp
@@ -7,29 +7,22 @@
 #include "common/tl/constants/common.h"
 #include "runtime-light/stdlib/component/component-api.h"
 #include "runtime-light/tl/tl-core.h"
-
-namespace {
-
-// Crypto-Specific TL magics
-
-constexpr uint32_t TL_CERT_INFO_ITEM_LONG = 0x533f'f89f;
-constexpr uint32_t TL_CERT_INFO_ITEM_STR = 0xc427'feef;
-constexpr uint32_t TL_CERT_INFO_ITEM_DICT = 0x1ea8'a774;
-
-constexpr uint32_t TL_GET_PEM_CERT_INFO = 0xa50c'fd6c;
-constexpr uint32_t TL_GET_CRYPTOSECURE_PSEUDORANDOM_BYTES = 0x2491'b81d;
-} // namespace
+#include "runtime-light/tl/tl-functions.h"
+#include "runtime-light/tl/tl-types.h"
 
 task_t<Optional<string>> f$openssl_random_pseudo_bytes(int64_t length) noexcept {
   if (length <= 0 || length > string::max_size()) {
     co_return false;
   }
 
-  tl::TLBuffer buffer;
-  buffer.store_trivial<uint32_t>(TL_GET_CRYPTOSECURE_PSEUDORANDOM_BYTES);
-  buffer.store_trivial<int32_t>(length);
-
   // TODO think about performance when transferring data
+
+  tl::GetCryptosecurePseudorandomBytes request {
+    .size = static_cast<int32_t>(length)
+  };
+
+  tl::TLBuffer buffer;
+  request.store(buffer);
 
   string request_buf;
   request_buf.append(buffer.data(), buffer.size());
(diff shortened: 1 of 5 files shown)


runtime-light/stdlib/crypto/crypto-functions.cpp
```diff
@@ -7,30 +7,23 @@
...
   // TODO think about performance when transferring data
 
+  tl::GetCryptosecurePseudorandomBytes request {
+    .size = static_cast<int32_t>(length)
+  };
+
+  tl::TLBuffer buffer;
+  request.store(buffer);
+
   string request_buf;
```
apolyakov, line 27: Not a blocker, but I think it's better to reduce number of names we introduce into a scope. For example, here we can just do following:
```
tl::TLBuffer buffer;
request.store(buffer);
auto query = f$component_client_send_request(string{"crypto"}, string{buffer.data(), buffer.size()});
```

runtime-light/tl/tl-types.cpp
```diff
@@ -38,4 +38,88 @@ void K2JobWorkerResponse::store(TLBuffer &tlb) const noexcept {
...
+  if (const auto magic = tlb.fetch_trivial<uint32_t>(); magic.value_or(TL_ZERO) != TL_DICTIONARY) {
+    return false;
+  }
+
+  const std::optional<uint32_t> size = tlb.fetch_trivial<uint32_t>();
+  if (!size.has_value()) {
+    return false;
+  }
+
+  auto response = array<mixed>::create();
```
apolyakov, line 55: Let's also add `reserve` here to reduce amount of reallocations.

runtime-light/tl/tl-types.h
```diff
@@ -22,4 +24,20 @@ struct K2JobWorkerResponse final {
   void store(TLBuffer &tlb) const noexcept;
 };
 
+// ===== CRYPTO =====
+
+enum CertInfoItem : uint32_t {
```
apolyakov, line 29: Should it really be declared in the header file? Is it necessary that clients know these magics?

apolyakov (changes requested):

runtime-light/tl/tl-types.cpp
```diff
@@ -53,6 +54,8 @@ bool GetPemCertInfoResponse::fetch(TLBuffer &tlb) noexcept {
   }
 
   auto response = array<mixed>::create();
+  constexpr size_t RESPONSE_SIZE = 10; // not more than
+  response.reserve(RESPONSE_SIZE, false);
```
apolyakov, line 58: Why not just use `size` from the line 51?

runtime-light/tl/tl-types.h
mkornaukhov03: It's actually one type, but with three magics
I think it's better than have three separate constexpr vars

apolyakov: No problem with having them as enumeration. I just think that they should not be a part of public interface of tl namespace

mkornaukhov03: What whould you suggest?

apolyakov: Move it to cpp file, for example

apolyakov: `K2_JOB_WORKER_RESPONSE_MAGIC` is used by clients outside of `tl` namespace.

mkornaukhov03: Where? I see it only in` tl-types.cpp`

apolyakov: Yep, you're right. Anyway, these magics should only be used to "detect" that we got a specific tl-type or tl-function. I think we should hide all the details about the implementation. `CertInfoItem` looks like implementation detail.

mkornaukhov03: Your logic is following: magics for functions and request/response types should be public, magic of everything else should be hidden in `.cpp` file, am I right?

runtime-light/tl/tl-types.cpp
mkornaukhov03: Yeah, fixed

runtime-light/tl/tl-types.h
apolyakov: I think that every high-level entity should have a publicly available magic, while all the implementation details must be hidden

mkornaukhov03: Who defines whether it "high-level entity" or "implementation details"?
What is specific criteria?

In my opinion, all TL stuff is implementation details regarding other runtime code

apolyakov: We do, of course :) For example, we can use following rule: everything that is only used inside some TL-type of -function should be hidden.

mkornaukhov03: OK, according to this rule all the magics (for now, at least) should be hidden?
Because all magics are used only inside some TL-combinator entity -- for now, all of them has `fetch` and/or `store` methods which are the single use cases

mkornaukhov03: Decided to move `CertInfoItem` into `.cpp` and later all magics defined in `.cpp`.

apolyakov (approved):
LGTM

PR #1106, opened 2024-09-18
Implement openssl_sign/openssl_verify in light runtime

mkornaukhov03:


builtin-functions/kphp-light/crypto.txt
apolyakov, line 1: I don't understand why do you moved all the functions out of `unsupported.txt`? Why not just do that to the functions you implemented?

apolyakov (changes requested):

builtin-functions/kphp-light/crypto.txt
mkornaukhov03: How to do that?
In case of having `unsupported/crypto.txt` and `crypto.txt` we gonna have compile error.
Have different filenames for them? 
I think current version is the most appropriate.
But you may suggest how to deal with it

apolyakov: Renaming `unsupported/crypto.txt` seems good enough

runtime-light/stdlib/crypto/crypto-functions.cpp
```diff
@@ -62,3 +62,48 @@ task_t<Optional<array<mixed>>> f$openssl_x509_parse(const string &data, bool sho
...
   co_return response.data;
 }
+
+task_t<bool> f$openssl_sign(const string &data, string &signature, const string &private_key, int64_t algo) noexcept {
+  tl::DigestSign request{.data = data, .private_key = private_key, .algorithm = static_cast<tl::DigestAlgorithm>(algo)};
+
+  tl::TLBuffer buffer;
+  request.store(buffer);
+
+  auto query = f$component_client_send_request(string("crypto"), string(buffer.data(), buffer.size()));
```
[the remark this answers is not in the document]
apolyakov: Not a blocker, but we prefer to use anonymous namespaces instead of `static` in runtime-light

builtin-functions/kphp-light/crypto.txt
mkornaukhov03: > Renaming `unsupported/crypto.txt` seems good enough

So, is this thread resolved?

runtime-light/stdlib/crypto/crypto-functions.cpp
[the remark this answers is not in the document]
mkornaukhov03: fixed

apolyakov (approved):

PR #1107, opened 2024-09-18
Add support for JSON serialized confdata values

apolyakov:
This PR adds new `MEMCACHE_JSON_SERIALIZED` flag that means that a value we got from MC is serialized in JSON format, so we need to deserialize it accordingly.

### DIFF
--- a/runtime/memcache.h
+++ b/runtime/memcache.h
@@ -23,7 +23,7 @@ bool mc_is_immediate_query(const string &key);
 
 constexpr inline auto MEMCACHE_SERIALIZED = static_cast<int64_t>(1U << 0U);
 constexpr inline auto MEMCACHE_COMPRESSED = static_cast<int64_t>(1U << 1U);
-constexpr inline auto MEMCACHE_JSON_SERIALZIED = static_cast<int64_t>(1U << 3U);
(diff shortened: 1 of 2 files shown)


runtime/memcache.h
```diff
@@ -25,9 +21,9 @@ mixed mc_get_value(const char *result_str, int32_t result_str_len, int64_t flags
 
 bool mc_is_immediate_query(const string &key);
 
-
-constexpr int64_t MEMCACHE_SERIALIZED = 1;
-constexpr int64_t MEMCACHE_COMPRESSED = 2;
+constexpr inline auto MEMCACHE_SERIALIZED = static_cast<int64_t>(1U << 0U);
```
Danil42Russia, line 24: what's wrong with `inline constexpr int64_t MEMCACHE_SERIALIZED = 1U << 0U`?

apolyakov: Nothing wrong :) Can you add more details to your question?

Danil42Russia: `auto` for `auto` and also `static_cast`.
It's not clear why it's there. I suggest not to use it here, but to make it explicit. In general, like other `inline constexpr` in the code

DrDet (approved):

PR #1109, opened 2024-09-25
Enhance K2 TL

apolyakov:
This PR adds following changes:

1. enhances TL primitives for K2 job workers;
2. adds support for some general TL types, e.g., vector, dictionary, etc;
3. adds TL primitives for K2 confdata.

### DIFF
--- a/runtime-light/tl/tl-types.h
+++ b/runtime-light/tl/tl-types.h
@@ -276,8 +276,6 @@ struct ConfdataValue final {
   bool is_json_serialized{};
 
   bool fetch(TLBuffer &tlb) noexcept;
-
-  void store(TLBuffer &tlb) const noexcept;
 };
 
 } // namespace tl
--- a/runtime-light/tl/tl-types.h
+++ b/runtime-light/tl/tl-types.h
@@ -4,16 +4,237 @@
 
 #pragma once
 
+#include <algorithm>
+#include <cstddef>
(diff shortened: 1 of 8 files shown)


runtime-light/tl/tl-types.h
```diff
@@ -4,16 +4,237 @@
...
+    if (opt_value.has_value()) {
+      tlb.store_trivial<uint32_t>(TL_MAYBE_TRUE);
+      (*opt_value).store(tlb);
+    } else {
+      tlb.store_trivial<uint32_t>(TL_MAYBE_FALSE);
+    }
+  }
+};
+
+template<tl_serializable T>
```
mkornaukhov03, line 63: I think it's better to distinguish `tl_fetchable` and `tl_storable` and move such concepts to fetch/store methods.
For example, I have a structure that are stored by KPHP, but read from k2. And I want to add new TL-function that will work with vector of such a structure. Then I'll have to implement fetch method, that will never be called in KPHP. It leads  dead code or forces to write stubs that are bad things

mkornaukhov03 (changes requested):

runtime-light/tl/tl-types.h
apolyakov: Nice point. Fixed

mkornaukhov03 (approved):
lgtm

PR #1111, opened 2024-10-01
add `kphp_request_script_time_max_running_interval` mertric collecting

PetrShumilov:
New metric represents a duration of maximum running interval when worker in script context 

Related issues: KPHP-1939

### DIFF
--- a/server/php-runner.cpp
+++ b/server/php-runner.cpp
@@ -183,6 +183,7 @@ void PhpScript::init(script_t *script, php_query_data_t *data_to_set) noexcept {
 
   script_time = 0;
   net_time = 0;
+  script_max_running_interval = 0;
   script_init_rusage = get_rusage_info();
 
   queries_cnt = 0;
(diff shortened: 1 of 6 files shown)


server/php-runner.cpp
```diff
@@ -281,7 +285,10 @@ void PhpScript::update_net_time() noexcept {
 
 void PhpScript::update_script_time() noexcept {
   double new_cur_timestamp = dl_time();
-  script_time += new_cur_timestamp - cur_timestamp;
+  double delta = new_cur_timestamp - cur_timestamp;
+  script_time += delta;
+  // Timestamp in nanoseconds, 1e-9 precision is enough
+  script_max_running_interval = (script_max_running_interval - delta > TIMESTAMP_PRECISION)  ? script_max_running_interval : delta;
```
apolyakov, line 291: Why not just `std::max` here?

PR #1116, opened 2024-10-11
Support ArrayAccess

mkornaukhov03:
[ArrayAccess](https://www.php.net/manual/en/class.arrayaccess.php) &mdash; interface that provides accessing objects almost the same as arrays.

# Usage

### How to
The basic sample is following:
```php
class MyClass implements \ArrayAccess {
  public offsetExists(mixed $offset): bool {...}
  public offsetGet(mixed $offset): mixed {...}
  public offsetSet(mixed $offset, mixed $value): void {...}
  public offsetUnset(mixed $offset): void {...}
}

function foo() {
  $x = new MyClass(...);
  $x[0] = 42; // $x->offsetSet(0, 42)
  $x[] = [1, 2, 3]; // $x->offsetSet(null, [1, 2, 3])
  $x["s"]; // $x->offsetGet("s")
  unset($x[0]); // $x->offsetUnset(0)
  isset($x[0]); // $x->offsetExists(0)
  empty($x[0]); // !$x->offsetExists(0) || ($x->offsetGet(0) == null)
}
```

> [!NOTE]
> In PHP, `offsetGet()` may return `mixed` reference, in KPHP it doesn't as return-by-reference is not supported.
> So in KPHP, with previous example expression `$x[1][0] = 42` will call `$x->offsetGet(1)` without modification.

### Finding out a structure of untyped global
Imagine you have a global that is typed as `mixed` and mainly stores array data.
And you don't have clear understanding of what exactly it stores.
You may use the following class to find out the structure:
```php
class LoggingArray implements ArrayAccess {
  /** @var mixed[] */
  protected $data = [];

  /** @param $data_ mixed[] */
  public function __construct($data_ = []) {
    $this->data = $data_;
  }

  public function offsetSet(mixed $offset, mixed $value) {
    if (flipCoin(0.01)) {
      log("set [" . $offset . "] = " . $value);
    }
    if (is_null($offset)) {
      $this->data[] = $value;
    } else {
      $this->data[$offset] = $value;
    }
  }

  public function offsetExists(mixed $offset): bool {
    return isset($this->data[$offset]);
  }

  public function offsetUnset(mixed $offset) {
    if ($this->offsetExists($offset)) {
      unset($this->data[$offset]);
    }
  }

  public function offsetGet(mixed $offset): mixed {
    if (flipCoin(0.01)) {
      log("get [" . $offset . "]");
    }
    return $this->offsetExists($offset) ? $this->data[$offset] : null;
  }
}
```
You may store derived classes in mixed using [`to_mixed()`](https://github.com/VKCOM/kphp/pull/1010) function.

### Mixing typed and untyped data
Imagine you have untyped data for user and store it like array:
```php
$user = ['id' => 1, "name" => "Paul", friends => [2, 3, 4]];
if (!check($user)) {
  $user = null;
}
```
and you want to use class instances instead of arrays (in the sake of type checking and other benefits of static type system),
but there are a lot functions that receive `mixed $user` and it's hard to change all of them, so you have to
support these legacy function and write new function with better type safety. In this case, you may use
`ArrayAccess` to support both object and array.

Snippet:

```php
class User implements \ArrayAccess {
  public int $id;
  public string $name;
  /** @var int[] */
  public $friends;

  public offsetExists(mixed $offset): bool {...}
  public offsetGet(mixed $offset): mixed {...}
  public offsetSet(mixed $offset, mixed $value): void {...}
  public offsetUnset(mixed $offset): void {...}
}

/** @param $user mixed */
function process_user_old($user) {
  var_dump($user['id']);
  var_dump($user['name']);
  var_dump($user['friends']);
}

/** @param $user User */
function process_user_new($user) {
  var_dump($user->id);
  var_dump($user->name);
  var_dump($user->$friends);
}

function test() {
  $user_obj = new Classes\User(1, "Paul", [2, 3, 4]);
  $user_arr = ["id" => 1, "name" => "Paul", "friends" => [2, 3, 4]];
    
  process_user_old($user_arr);
  process_user_old(to_mixed($user_obj));

  process_user_new($user_obj);
}
```

# Implementation details
This section is intended for KPHP developes.

Basically, it's just replacement of AST `operator[]` node to function calls of corresponding functions.

First of all, this PR introduces extends `phpdoc` annotation `@kphp-required`.
When used for interfaces, it forces generation of polymorphic interface functions based on user PHP-code.

The changes in the compiler affect `OptimizationPass`. Previously, it done some things with `op_index`, for example
this pass replaces `op_index` nodes in some cases with special array or mixed methods, depending on previous or children nodes.
But this passes doesn't allow to correctly replace `isset($obj[0])` or `empty($obj[0])` because we need context that node is 
inside `isset, empty, unset` call. That's why I added `ArrayAcessTransformPass` pass that allow to do such transformations.
But there are some nuances. Look at the following snippet with some class `A` implementing `ArrayAccess`:
```php
$obj_a = new A(...);
$obj_a[0] = $obj[1] = 123;
```
How subexpression `$obj[1] = 123;` can be executed? You may think that it's `offsetSet()` + `offsetGet()`, but according to PHP 
implementation `offsetGet()` shouldn't be called.
Or look at another example:
```php
class B {
  /** @var A */
  public $a;
}

$obj_a = new A(...);
$obj_a[0] = [1, 2, 3];
$b = new B();
$b->a = $obj_a;

isset($b->obj_a[0][0]);
```
According to PHP implementation, `$b->obj_a[0]` should call `$b->obj_a->offsetExist(0)` and 
only in case of returning `true`, `$b->obj_a->offsetGet(0)` should be called.
For two previous samples I added custom vertices and runtime macros for them.
There is also a change in the type system: in case of `expr[0]`-expression, when type of `expr` is class that implements
`ArrayAccess`, then type of `expr[0]` is mixed.

In one of the previous [PR](https://github.com/VKCOM/kphp/pull/1010) we've supported class instances inside `mixed`. It's
obvious that we should support `ArrayAccess` objects inside mixed. There're some issues with current `mixed` interface, so I've changed previous and added some new functions (`empty_at()`, `set_value_return()` and so on).
I've also added `Materialized` class for mixed that is used for temporary mixed objects that are generated by `operator[]`.

Besides, it turns out that we cannot do optimization `$mix === null --> isset($mix)`, because for object implementing `ArrayAccess` logics of `isset()` may be absolutely custom, this is not necessarily checking for null.

> [!NOTE]
> In K2 mode `interruptible` functions are prohibited. The main reason is that a lot of `mixed` methods become `interruptible`, for now I don't want to rewrite whole `mixed` class.


apolyakov (changes requested):
Let's start with some high level topics:

- Can you add tests for cases like: `array[index1](arg)[index2]`? I.e. accesing `index2` in an object of `ArrayAccess` type returned by function we got from accessing `array` at `index1`
- Do we really need new `kphp-internal-interface` tag? Can we modify or even use as it the `kphp-required` tag?
- Slots in ArrayAccessContext look oddly. It seems like a hack that we need due to errors in high-level design
- Generally, I don't your approach of simplifying the compiler by adding lots of helpers into runtime. I think it's better to implement `ArrayAccess` with AST manipulation, e.g. `empty($obj["one"]); // => empty($obj.offsetExists("one") ? $obj(offsetGet("one")) : mixed())
`

compiler/pipes/calc-bad-vars.cpp
```diff
@@ -547,6 +549,11 @@ class CalcBadVars {
     for (const auto &func : call_graph.functions) {
       if (into_interruptible[func]) {
         func->is_interruptible = true;
+        if (unlikely(func->class_id && func->class_id->is_required)) {
+          std::string func_name = func->name;
+          std::replace(func_name.begin(), func_name.end(), '$', ':');
+          kphp_error(false, fmt_format("{} cannot be interruptible", func_name).c_str());
```
apolyakov, line 555: What is it for?

runtime-common/core/array_access.h
```diff
@@ -0,0 +1,24 @@
+// Compiler for PHP (aka KPHP)
+// Copyright (c) 2024 LLC «V Kontakte»
+// Distributed under the GPL v3 License, see LICENSE.notice.txt
+
+#pragma once
```
apolyakov, line 5: missing newline

runtime-common/core/array_access.h
```diff
@@ -0,0 +1,24 @@
...
+#include "runtime-common/core/class-instance/refcountable-php-classes.h"
+
+#ifndef INCLUDED_FROM_KPHP_CORE
+  #error "this file must be included only from runtime-core.h"
+#endif
+
+struct C$ArrayAccess : public may_be_mixed_base {
+  virtual int get_hash() const noexcept = 0;
+
+  C$ArrayAccess() __attribute__((always_inline)) = default;
```
apolyakov, line 16: Let's not use these attributes. These methods are implicitly inline

runtime-common/core/array_access.h
```diff
@@ -0,0 +1,24 @@
...
+  ~C$ArrayAccess() override __attribute__((always_inline)) = default;
+};
+
+
+bool f$ArrayAccess$$offsetExists(class_instance<C$ArrayAccess> const & /*v$this*/, mixed const & /*v$offset*/) noexcept;
```
apolyakov, line 21: `const T &`

runtime-common/core/core-context.h
```diff
@@ -25,5 +25,6 @@ struct RuntimeContext final : vk::not_copyable {
   void init() noexcept;
   void free() noexcept;
 
+
```
apolyakov, line 28: Redundant change

runtime-common/core/core-types/decl/mixed_decl.inl
```diff
@@ -24,6 +26,52 @@ template<typename T>
 struct is_type_acceptable_for_mixed<class_instance<T>> : std::is_base_of<may_be_mixed_base, T> {
 };
 
+template<class Mix>
+class Materialized {
```
apolyakov, line 30: Let's mark it `final`

runtime-common/core/core-types/decl/mixed_decl.inl
```diff
@@ -24,6 +26,52 @@ template<typename T>
...
+  using MixRef = Mix &;
+  std::variant<Mix, Mix *> slot_;
+
+  Materialized(Mix *ptr)
```
apolyakov, line 34: Please, add `noexcept` modifiers

runtime-common/core/core-types/decl/mixed_decl.inl
```diff
@@ -222,6 +285,15 @@ public:
   const char *get_type_c_str() const;
   const char *get_type_or_class_name() const;
 
+
+  template<typename T>
+  inline bool empty_on(T key) const;
```
apolyakov, line 290: Redundant inline modifier

apolyakov, line 290: Can we give a better name? Now it's a little confusing

runtime-common/core/core-types/definition/mixed.cpp
```diff
@@ -1019,6 +1024,55 @@ const string mixed::get_type_str() const {
   return string(get_type_c_str());
 }
 
+std::pair<class_instance<C$ArrayAccess>, bool> try_as_array_access(const mixed &m) noexcept {
```
apolyakov, line 1027: Don't you want to add a new tag `ARRAY_ACCESS_OBJECT` instead?

apolyakov, line 1027: It's worth considering `dynamic_cast`-like interface for that function

tests/cpp/runtime/_runtime-tests-env.cpp
```diff
@@ -82,3 +83,22 @@ char **get_runtime_options(int *) noexcept {
   assert(0 && "this code shouldn't be executed and only for linkage test");
   return nullptr;
 }
+
+
+bool f$ArrayAccess$$offsetExists(class_instance<C$ArrayAccess> const & /*v$this*/, mixed const & /*v$offset*/) noexcept {
```
apolyakov, line 88: `const T &`

apolyakov (changes requested):

compiler/code-gen/vertex-compiler.cpp
```diff
@@ -850,6 +857,20 @@ void compile_func_call(VertexAdaptor<op_func_call> root, CodeGenerator &W, func_
     }
   }
 
+  if (root->func_id->is_extern() && root->str_val == "empty") {
```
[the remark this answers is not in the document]
mkornaukhov03: Yes, fixed

compiler/code-gen/vertex-compiler.cpp
```diff
@@ -1786,13 +1807,23 @@ bool try_compile_append_inplace(VertexAdaptor<op_set_dot> root, CodeGenerator &W
...
 }
 
-void compile_index_of_array(VertexAdaptor<op_index> root, CodeGenerator &W) {
+void compile_index_of_array_or_mixed(VertexAdaptor<op_index> root, CodeGenerator &W) {
   bool used_as_rval = root->rl_type != val_l;
   if (!used_as_rval) {
     kphp_assert(root->has_key());
-    W << root->array() << "[" << root->key() << "]";
+    if (tinf::get_type(root->array())->ptype() == tp_mixed) {
+      W << "static_cast<mixed&>(" << root->array() << "[" << root->key() << "])";
```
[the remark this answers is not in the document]
mkornaukhov03: fixed

compiler/pipes/calc-bad-vars.cpp
mkornaukhov03: It `offsetGet()` function calls interruptible function, this error will be printed. Without this, it would fail in `C++ to bin` stage with unclear error.

runtime-common/core/array_access.h
mkornaukhov03: fixed

runtime-common/core/array_access.h
mkornaukhov03: Firstly, this attribute to about linkage things, it's about forcing compiler to inline this function in terms of getting rid of extra function call.
But this is exactly how it would be codegenerated and I want to leave consistency with codegen.

runtime-common/core/array_access.h
mkornaukhov03: Same, I want it to be as codegenerated

runtime-common/core/core-context.h
mkornaukhov03: fixed

runtime-common/core/core-context.h
```diff
@@ -25,5 +25,6 @@ struct RuntimeContext final : vk::not_copyable {
...
   static RuntimeContext &get() noexcept;
-};
+};
```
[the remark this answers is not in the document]
mkornaukhov03: fixed

runtime-common/core/core-types/decl/mixed_decl.inl
```diff
@@ -6,6 +6,8 @@
 
 #include "common/smart_ptrs/intrusive_ptr.h"
 #include "runtime-common/core/class-instance/refcountable-php-classes.h"
+#include <utility>
```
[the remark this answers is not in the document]
mkornaukhov03: fixed

runtime-common/core/core-types/decl/mixed_decl.inl
```diff
@@ -24,6 +26,52 @@ template<typename T>
 struct is_type_acceptable_for_mixed<class_instance<T>> : std::is_base_of<may_be_mixed_base, T> {
 };
 
+template<class Mix>
```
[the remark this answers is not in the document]
mkornaukhov03: Technically yes. I would get rid of this template argument, but I cannot do it. `mixed` needs this wrapper class definition, this wrapper needs `mixed` class definition. This problem is like in `CRTP` pattern. 
But if you know how to do it in another way – you're welcome.

runtime-common/core/core-types/decl/mixed_decl.inl
mkornaukhov03: fixed

runtime-common/core/core-types/decl/mixed_decl.inl
mkornaukhov03: done

runtime-common/core/core-types/decl/mixed_decl.inl
mkornaukhov03: fixed

mkornaukhov03: Could you give a tip?
This methods "checks that item that's stored by `key` index is empty". `empty_on/empty_at` are good for me, but you may advise something better

tests/cpp/runtime/_runtime-tests-env.cpp
mkornaukhov03: Same thing, it's like codegenerated

runtime-common/core/core-types/definition/mixed.cpp
mkornaukhov03: No, I don't.
In this case `ARRAY_ACCESS_OBJECT` would imply `OBJECT` etc, so I think it's overcomplication

compiler/code-gen/vertex-compiler.cpp
```diff
@@ -1786,19 +1805,31 @@ bool try_compile_append_inplace(VertexAdaptor<op_set_dot> root, CodeGenerator &W
...
+      past = ")";
+    }
+    W << pre << root->array() << "[" << root->key() << "]" << past;
   } else {
-    W << root->array() << ".get_value (" << root->key();
-    // if it's a const string key access like $a['somekey'],
-    // compute the 'somekey' string hash during the compile time and call array<T>::get_value(string, precomputed_hash)
-    if (auto precomputed_hash = can_use_precomputed_hash_indexing_array(root->key())) {
-      W << ", " << precomputed_hash << "_i64";
+    if (tinf::get_type(root->array())->ptype() == tp_mixed && root->inside_isset) {
```
apolyakov, line 1820: Can we use precomputed hash for these cases?

compiler/pipes/fix-array-access.h
```diff
@@ -0,0 +1,18 @@
+// Compiler for PHP (aka KPHP)
+// Copyright (c) 2024 LLC «V Kontakte»
+// Distributed under the GPL v3 License, see LICENSE.notice.txt
+
+#pragma once
+
+#include "compiler/function-pass.h"
+
+class FixArrayAccessPass final : public FunctionPassBase {
```
apolyakov, line 9: I'd suggest to change the name of the pass. For example, `ArrayAccessTransformPass`

runtime-common/core/array_access.h
apolyakov: I mean that any definition inside class declaration is implicitly `inline`, so no need to add explicit attributes here

runtime-common/core/array_access.h
apolyakov: Is there a reason for that?

runtime-common/core/array_access.h
apolyakov: Also, you can add `noexcept` here.

runtime-common/core/core-types/definition/mixed.cpp
apolyakov: It makes things more complicated, yep. But it eliminates dynamic cast on each call to a frequently used operation.

apolyakov (approved):
:+1:

PR #1117, opened 2024-10-11
K2: add new CI steps

apolyakov:


tests/kphp_ci_pipeline_runner.py
```diff
@@ -170,10 +170,18 @@ def parse_args():
         type=str,
         dest="cxx_name",
         default="g++",
-        choices=["g++", "clang++", "clang++-16"],
+        choices=["g++", "clang++", "clang++-16", "clang++-18", "clang++-19"],
```
mkornaukhov03, line 173: I think we should throw away clang16.
The current solution raises a question: why clang16 is here, but clang17 is not?
The last release of clang16 was more than a year ago, it's obsolete now, v18 and v19 are good I think

tests/kphp_ci_pipeline_runner.py
```diff
@@ -281,6 +309,20 @@ def parse_args():
...
+        description="run k2-kphp tests with cxx={}".format(args.cxx_name),
+        cmd="KPHP_TESTS_POLYFILLS_REPO={kphp_polyfills_repo} "
+            "{kphp_runner} -j{jobs} --cxx-name {cxx_name} --k2-bin {k2_bin}".format(
+            jobs=n_cpu,
+            kphp_polyfills_repo=kphp_polyfills_repo,
+            kphp_runner=kphp_test_runner,
+            cxx_name=args.cxx_name,
+            k2_bin=args.k2_bin,
+        ),
+        skip=not args.k2_bin or args.steps and "k2-kphp-tests" not in args.steps,
```
mkornaukhov03, line 323: Too hard to parse, I would add some parentheses

mkornaukhov03 (changes requested):

tests/kphp_ci_pipeline_runner.py
apolyakov: `clang16` is used in current `KPHP` pipelines. Removing it also requires modifying the pipelines. It's removal definitely should be a distinct step.

tests/kphp_ci_pipeline_runner.py
apolyakov: Fixed

astrophysik (approved):

tests/kphp_ci_pipeline_runner.py
```diff
@@ -8,7 +8,7 @@
 import sys
 from enum import Enum
 
-from python.lib.colors import red, green, yellow, cyan
+from python.lib.colors import cyan, green, red, yellow
```
mkornaukhov03, line 11: Do we really need to break blame in such places?

mkornaukhov03 (approved):
Looks good
I wouldn't make formatting changes, but up to you

PR #1121, opened 2024-10-17
K2: add TL types and functions for HTTP queries

apolyakov:
This PR also does some small refactoring of K2 TL.

### DIFF
--- a/runtime-light/stdlib/job-worker/job-worker-api.cpp
+++ b/runtime-light/stdlib/job-worker/job-worker-api.cpp
@@ -68,7 +68,7 @@ task_t<int64_t> kphp_job_worker_start_impl(string request, double timeout, bool 
     const string response{(co_await wait_with_timeout_t{task_t<string>::awaiter_t{std::addressof(fetch_task)}, timeout}).value_or(string{})};
 
     tl::TLBuffer tlb{};
-    tlb.store_bytes(response.c_str(), static_cast<size_t>(response.size()));
+    tlb.store_bytes({response.c_str(), static_cast<size_t>(response.size())});
     tl::K2JobWorkerResponse jw_response{};
     if (!jw_response.fetch(tlb)) {
       co_return string{};
--- a/runtime-light/stdlib/rpc/rpc-api.cpp
+++ b/runtime-light/stdlib/rpc/rpc-api.cpp
@@ -249,7 +249,7 @@ task_t<array<mixed>> rpc_tl_query_result_one_impl(int64_t query_id) noexcept {
     co_return make_fetch_error(string{"rpc response timeout"}, TL_ERROR_QUERY_TIMEOUT);
   }
   rpc_ctx.rpc_buffer.clean();
-  rpc_ctx.rpc_buffer.store_bytes(data.c_str(), data.size());
+  rpc_ctx.rpc_buffer.store_bytes({data.c_str(), static_cast<size_t>(data.size())});
   co_return fetch_function_untyped(rpc_query);
 }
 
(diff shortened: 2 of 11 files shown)


runtime-light/tl/tl-types.h
```diff
@@ -304,4 +312,148 @@ struct confdataValue final {
...
+    V3 = 0xe725'a7a2,
+  };
+
+  Version version{Version::Invalid};
+
+  bool fetch(TLBuffer &tlb) noexcept {
+    using version_utype = std::underlying_type_t<Version>;
+
+    switch (tlb.fetch_trivial<uint32_t>().value_or(TL_ZERO)) {
+      case static_cast<version_utype>(Version::V09): {
```
mkornaukhov03, line 333: I would prefer to get rid of `enum class`, just `enum`. And here wouldn't be awful `static_casts`-s.
But up to you

runtime-light/tl/tl-types.h
```diff
@@ -275,7 +280,10 @@ struct K2JobWorkerResponse final {
 
 // Actually it's "Maybe (Dictionary CertInfoItem)"
 // But I now want to have this logic separately
-struct GetPemCertInfoResponse final {
+class GetPemCertInfoResponse final {
```
mkornaukhov03, line 283: Why both structs and classes here?

runtime-light/stdlib/rpc/rpc-api.cpp
```diff
@@ -404,7 +404,7 @@ bool is_int32_overflow(int64_t v) noexcept {
 }
 
 void store_raw_vector_double(const array<double> &vector) noexcept { // TODO: didn't we forget vector's length?
-  RpcComponentContext::get().rpc_buffer.store_bytes(reinterpret_cast<const char *>(vector.get_const_vector_pointer()), sizeof(double) * vector.count());
+  RpcComponentContext::get().rpc_buffer.store_bytes({reinterpret_cast<const char *>(vector.get_const_vector_pointer()), sizeof(double) * vector.count()});
```
mkornaukhov03, line 407: I would introduce new variable here for view, too hard to read

runtime-light/stdlib/job-worker/job-worker-api.cpp
```diff
@@ -68,7 +68,7 @@ task_t<int64_t> kphp_job_worker_start_impl(string request, double timeout, bool
     const string response{(co_await wait_with_timeout_t{task_t<string>::awaiter_t{std::addressof(fetch_task)}, timeout}).value_or(string{})};
 
     tl::TLBuffer tlb{};
-    tlb.store_bytes(response.c_str(), static_cast<size_t>(response.size()));
+    tlb.store_bytes({response.c_str(), static_cast<size_t>(response.size())});
```
mkornaukhov03, line 71: Maybe should add new method for string that returns size that is casted to `size_t`?

mkornaukhov03 (changes requested):
I've seen only tl-stuff, haven't checked semantics of http things

runtime-light/tl/tl-types.h
apolyakov: I always prefer enum classes due to their safety

runtime-light/stdlib/rpc/rpc-api.cpp
apolyakov: Done

runtime-light/tl/tl-types.h
apolyakov: Just to not introduce a lot of `public, private` annotations

runtime-light/stdlib/job-worker/job-worker-api.cpp
apolyakov: We can't overload a function by its return type

mkornaukhov03: We can call it in another way
`size_cast<T>`, `std_size` or smth like that

runtime-light/tl/tl-types.h
mkornaukhov03: Seems strange a little, but OK

runtime-light/stdlib/job-worker/job-worker-api.cpp
apolyakov: I'd rather change the signature of `string::size()` method so that it returns `size_t`. But it requires a lot of refactoring in the current runtime

mkornaukhov03: Yeah, that's why new method has only benefits
Just to think about, not really related to PR

mkornaukhov03 (approved):
ok

PR #1122, opened 2024-10-21
Get rid of CMAKE_INSTALL_PREFIX in some places

mkornaukhov03:


cmake/init-global-vars.cmake
Danil42Russia, line 1: Are we sure we still need the `CPACK_PACKAGING_INSTALL_PREFIX` variable in this file?

mkornaukhov03: What are your suggestions?
I see 2 approaches:
* Move it another file. Which one and what for?
* Get rid of. But it's special variable that cmake uses as prefix dir for installing packages

Or maybe I got you in a wrong way?

Danil42Russia: yes, you right

Danil42Russia (approved):

PR #1124, opened 2024-10-22
Initial support for K2 http queries

apolyakov:
Changes it the PR: 

- initialize some common superglobals for all kinds of components
- initialize most of PHP superglobals for K2 server


runtime-light/component/component.cpp
```diff
@@ -48,8 +51,30 @@ void ComponentState::init_script_execution() noexcept {
 template<ComponentKind kind>
 task_t<void> ComponentState::run_component_prologue() noexcept {
   static_assert(kind != ComponentKind::Invalid);
-
   component_kind_ = kind;
+
+  // common initialization
+  auto &superglobals{php_script_mutable_globals_singleton.get_superglobals()};
```
astrophysik, line 57: Maybe move it to separate function like `common_superglobals_initialization`. Because in the future, body of `run_component_prologue` may become too complex.

runtime-light/core/globals/php-script-globals.h
```diff
@@ -8,7 +8,40 @@
...
+inline constexpr auto *REMOTE_ADDR = "REMOTE_ADDR";
+inline constexpr auto *REMOTE_PORT = "REMOTE_PORT";
+
+inline constexpr auto *QUERY_STRING = "QUERY_STRING";
+inline constexpr auto *GATEWAY_INTERFACE = "GATEWAY_INTERFACE";
+
+}; // namespace PhpServerSuperGlobalIndices
+
 struct PhpScriptBuiltInSuperGlobals {
+
```
astrophysik, line 44: Unnecessary diff

runtime-light/core/globals/php-script-globals.h
```diff
@@ -8,7 +8,40 @@
 #include "runtime-core/memory-resource/unsynchronized_pool_resource.h"
 #include "runtime-core/runtime-core.h"
 
+namespace PhpServerSuperGlobalIndices {
+
+inline constexpr auto *ARGC = "argc";
```
astrophysik, line 13: What is the logic behind the constants location? For example `php-script-globals.h` contains `JOB_ID` that doesn't look generic

runtime-light/server/http/init-functions.cpp
```diff
@@ -0,0 +1,188 @@
...
+      }
+      if (std::strstr(header.value.c_str(), ENCODING_DEFLATE) != nullptr) {
+        http_server_ctx.encoding |= HttpServerComponentContext::ENCODING_DEFLATE;
+      }
+    } else if (std::strcmp(header_name.c_str(), HEADER_COOKIE) == 0) {
+      process_cookie_header(header.value, superglobals);
+    } else if (std::strcmp(header_name.c_str(), HEADER_HOST) == 0) {
+      server.set_value(string{SERVER_NAME, std::char_traits<char>::length(SERVER_NAME)}, header.value);
+    } else if (std::strcmp(header_name.c_str(), HEADER_AUTHORIZATION) == 0) {
+      // TODO parse authorization
```
astrophysik, line 101: Let's add warning in such cases

runtime-light/server/http/init-functions.cpp
```diff
@@ -0,0 +1,188 @@
...
+    }
+    // add header entries
```
astrophysik, line 103: In [`runtime/interface.cpp`](https://github.com/VKCOM/kphp/blob/master/runtime/interface.cpp#L1648) this block stored in else branch

runtime-light/server/http/init-functions.cpp
```diff
@@ -0,0 +1,188 @@
...
+    server.set_value(string{REQUEST_URI, std::char_traits<char>::length(REQUEST_URI)}, invoke_http.uri.path);
+  }
+
+  process_headers(std::move(invoke_http.headers), superglobals);
+  // TODO connection_kind, CONTENT_TYPE, is_head_query, _REQUEST, v_POST
+
+  string real_scheme_header{HTTP_X_REAL_SCHEME, std::char_traits<char>::length(HTTP_X_REAL_SCHEME)};
+  string real_host_header{HTTP_X_REAL_HOST, std::char_traits<char>::length(HTTP_X_REAL_HOST)};
+  string real_request_header{HTTP_X_REAL_REQUEST, std::char_traits<char>::length(HTTP_X_REAL_REQUEST)};
+  if (server.isset(real_scheme_header) || server.isset(real_host_header) || server.isset(real_request_header)) {
```
astrophysik, line 175: In [`runtime/interface.cpp`](https://github.com/VKCOM/kphp/blob/master/runtime/interface.cpp#L1680) condition is different

runtime-light/core/globals/php-script-globals.h
apolyakov: I don't actually see it there

runtime-light/server/http/init-functions.cpp
apolyakov: The semantics is equal in fact

runtime-light/server/http/init-functions.cpp
apolyakov: Can you please provide a link or an example?

apolyakov: You're right, fixed

astrophysik (approved):

PR #1126, opened 2024-10-23
Move runtime-core into runtime-common

apolyakov:


astrophysik:
Maybe rename `runtime-core` to just `core`? Because `/runtime-common/runtime-core` looks strange as for me

apolyakov:
> Maybe rename `runtime-core` to just `core`? Because `/runtime-common/runtime-core` looks strange as for me

I was thinking about such name. But after consideration I decided to leave it as `runtime-core` since `core` doesn't clarify enough what core is it? Is it runtime-core, server-core, stdlib-core etc

astrophysik:
I think it's clear that if the core directory is in runtime-common, it refers to runtime

astrophysik (approved):

PR #1127, opened 2024-10-23
add efficient user context manipulation routines

PetrShumilov:
Goals:
* Remove useless `rt_sigprocmask` syscall invocation during user context manipulation 
* Add platform-independent API for user context-related things

Related issues: KPHP-1944


server/php-runner.cpp
```diff
@@ -80,13 +81,24 @@ void send_slow_net_event_stats(const net_event_t &event, double time_sec) noexce
...
   assert (in_script_context == true);
   in_script_context = false;
   current_script->state = run_state_t::error;
   current_script->error_message = error_message;
   current_script->error_type = error_type;
-  stack_end = reinterpret_cast<char *>(exit_context.uc_stack.ss_sp) + exit_context.uc_stack.ss_size;
+  stack_end = reinterpret_cast<char *>(get_context_stack_ptr_portable(exit_context)) + get_context_stack_size_portable(exit_context);
+
+#if defined(__linux__) && defined(__x86_64__)
+  // The PhpScript::error may be produced in process of signal handling. The default behavior on Linux-based platforms
```
astrophysik, line 93: We also have own implementation for `darwin aarch64`. Do we need same logic for this?

server/php-runner.cpp
```diff
@@ -477,7 +489,7 @@ void PhpScript::run() noexcept {
       php_assert(CurException.is_null());
       CurException = saved_exception;
     }
-    error(php_uncaught_exception_error(CurException), script_error_t::exception);
+    error(php_uncaught_exception_error(CurException), script_error_t::exception, /* no triggered by signal */ std::nullopt);
```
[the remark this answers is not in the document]
PetrShumilov: I don't think so. It's a good hint for reader, furthermore for developers without IDE. I suppose it's best practice. 
Reader first!!!

server/php-runner.cpp
PetrShumilov: I will check it. Previously, we also had own implementation for `darwin aarch64` but no needed a special handling for this case. I may be a result of different signal subsystems.

PetrShumilov: Checked, thanks for notice, fixed for `darwin aarch64`

DrDet (approved):

DrDet (approved):

PR #1130, opened 2024-10-26
Move most of string functions into runtime-common/stdlib

apolyakov:


astrophysik:
Okay, but this pr breaks git history. That isn't necessary as for me

PR #1132, opened 2024-10-31
Move some KPHP tracing functionality into runtime-common/stdlib

apolyakov:


runtime/kphp_tracing.cpp
```diff
@@ -633,16 +640,17 @@ class_instance<C$KphpDiv> f$kphp_tracing_init(const string &root_span_title) {
...
 
 void f$kphp_tracing_set_level(int64_t trace_level) {
-  if (kphp_tracing::cur_trace_level == -1 || kphp_tracing::cur_trace_level == trace_level || trace_level < 0 || trace_level > 2) {
+  auto cur_trace_level = kphp_tracing::TracingContext::get().cur_trace_level;
+  if (cur_trace_level == -1 || cur_trace_level == trace_level || trace_level < 0 || trace_level > 2) {
     return;
   }
 
-  kphp_tracing::cur_trace_level = trace_level; // 0 may be used to disable tracing in the middle
+  cur_trace_level = trace_level; // 0 may be used to disable tracing in the middle
```
DrDet, line 648: It must be global var assignment here, but now it just modifies local variable and statement is redundant. Maybe we need to use some stat analyzers to catch bugs like this automatically
Also let's add small test for it like:
```php
<?php
kphp_tracing_set_level(2);
assert (kphp_tracing_get_level() == 2);
```

DrDet (changes requested):

runtime/kphp_tracing.cpp
apolyakov: Fixed

apolyakov: We can't add such test since stubs in `kphp-polyfills` are empty

DrDet (approved):

PR #1136, opened 2024-11-03
Move regex into runtime-common

apolyakov:


apolyakov:
Functionality already done in #1165

PR #1137, opened 2024-11-05
K2: update header to version 11

apolyakov:
Changes:

1. new `k2` namespace with wrappers of raw K2 primitives;
2. rename `ComponentState` -> `InstanceState`;
3. dummy support of new `ComponentState`;
4. new initiallization scheme for `ImageState`, `ComponentState`, `InstanceState`.


runtime-light/utils/logs.h
```diff
@@ -4,10 +4,12 @@
...
 
-#include "runtime-light/header.h"
+#include <cstddef>
 
-#define Error 1
-#define Warn 2
-#define Info 3
-#define Debug 4
-#define Trace 5
+enum LogLevel : size_t {
```
astrophysik, line 9: Maybe do it enum class to hide values?

runtime-light/utils/php_assert.cpp
```diff
@@ -61,5 +60,5 @@ void php_error(char const *message, ...) {
 void php_assert__(const char *msg, const char *file, int line) {
   php_error("Assertion \"%s\" failed in file %s on line %d", msg, file, line);
   critical_error_handler();
-  _exit(1);
+  k2::exit(1);
```
astrophysik, line 63: I don't think `k2::exit` is necessary here. It already done in `critical_error_handler`

astrophysik (approved):

runtime-light/utils/php_assert.cpp
apolyakov: I think it makes it clear that we pass the control to the platform.

astrophysik (approved):

astrophysik (approved):

PR #1145, opened 2024-11-12
Move some url functions into runtime-common

apolyakov:


runtime-common/stdlib/server/url-functions.cpp
```diff
@@ -0,0 +1,471 @@
...
+#include <string_view>
+
+#include "common/macos-ports.h"
+#include "runtime-common/core/runtime-core.h"
+#include "runtime-common/stdlib/string/string-context.h"
+#include "runtime-common/stdlib/string/string-functions.h"
+
+namespace {
+
+constexpr short base64_reverse_table[256] = {-2, -2, -2, -2, -2, -2, -2, -2, -2, -1, -1, -2, -2, -1, -2, -2, -2, -2, -2, -2, -2, -2, -2, -2, -2, -2, -2, -2, -2,
```
astrophysik, line 19: Maybe make it `std::array`?

astrophysik (changes requested):

astrophysik (approved):

PR #1148, opened 2024-11-13
Added new confdata statshouse metrics

DrDet:


server/statshouse/statshouse-manager.cpp
```diff
@@ -334,6 +334,16 @@ void StatsHouseManager::add_confdata_master_stats(const ConfdataStats &confdata_
   client.metric("kphp_confdata_memory").tag("used").write_value(memory_stats.memory_used);
   client.metric("kphp_confdata_memory").tag("real_used").write_value(memory_stats.real_memory_used);
 
+  client.metric("kphp_confdata_tree_nodes").tag("no_dots").write_value(confdata_stats.simple_key_elements);
+  client.metric("kphp_confdata_tree_nodes").tag("one_dot").write_value(confdata_stats.one_dot_wildcards);
+  client.metric("kphp_confdata_tree_nodes").tag("two_dots").write_value(confdata_stats.two_dots_wildcards);
+  client.metric("kphp_confdata_tree_nodes").tag("predefined_wildcard").write_value(confdata_stats.predefined_wildcards);
+  client.metric("kphp_confdata_tree_nodes").tag("total").write_value(confdata_stats.total_rb_tree_size);
```
Danil42Russia, line 341: maybe not `total`, but `size`?
for me, `total` is the number of elements, not the `size`

DrDet: I think metric name `kphp_confdata_tree_nodes` clarifies it exhaustively, so it means `total` of `tree_nodes`

server/confdata-stats.h
```diff
@@ -17,6 +17,7 @@ struct ConfdataStats : private vk::not_copyable {
 
   std::chrono::nanoseconds initial_loading_time{std::chrono::nanoseconds::zero()};
   std::chrono::nanoseconds total_updating_time{std::chrono::nanoseconds::zero()};
+  std::chrono::nanoseconds time_since_last_update{std::chrono::nanoseconds::zero()};
```
astrophysik, line 20: Is it really time since last update? Looks like it time between two previous updates

server/statshouse/statshouse-manager.cpp
```diff
@@ -334,6 +334,16 @@ void StatsHouseManager::add_confdata_master_stats(const ConfdataStats &confdata_
...
+
+  if (confdata_stats.time_since_last_update > std::chrono::nanoseconds::zero()) {
+    client.metric("kphp_confdata_time_since_last_update").write_value(confdata_stats.time_since_last_update.count());
```
astrophysik, line 344: Why last update time is not looks like 'now - last update time'?

DrDet: Because `last update time` here is already updated and `now - las_update_time` is epsilon

PR #1152, opened 2024-11-15
Add Zlib as a git subtree

apolyakov:


third-party/zlib/.github/workflows/cmake.yml
Danil42Russia, line 1: You need to add the `third-party` folder to the `linguist-vendor`

apolyakov: Nice point, thanks. Added `.gitattributes` file and marked `third-party` directory for Linguist

PR #1157, opened 2024-11-18
Added debug info disable probability options

DrDet:
Added two compiler options:
- `debug-info-force-disable-probability` - probability to partially disable debug info for some random part of files. Needed as short-term fix of `.debug_info` ELF section overflow
- `debug-info-force-disable-probability-seed` - the seed for the former, to be able to change set of randomly chosen files

### DIFF
--- a/compiler/kphp2cpp.cpp
+++ b/compiler/kphp2cpp.cpp
@@ -270,7 +270,7 @@ int main(int argc, char *argv[]) {
              "extra-linker-flags", "KPHP_EXTRA_LDFLAGS", get_default_extra_ldflags());
   parser.add("C++ compiler debug level for building the output binary", settings->extra_cxx_debug_level,
              "debug-level", "KPHP_DEBUG_LEVEL");
-  parser.add("Probability to partially disable debug info for some random part of files",
+  parser.add("Probability of partially disabling debug info for some random part of files",
              settings->debug_info_force_disable_prob, "debug-info-force-disable-probability", "KPHP_DEBUG_INFO_FORCE_DISABLE_PROBABILITY");
   parser.add("Seed for random check from --debug-info-force-disable-probability option",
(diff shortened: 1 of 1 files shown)


compiler/code-gen/common.h
```diff
@@ -24,6 +27,23 @@ struct OpenFile {
...
+        this->compile_with_debug_info_flag = false;
+      }
+    }
+  }
+
+  bool flip_coin(double prob) const {
+    static constexpr size_t PROB_RANGE = 100000;
+    size_t hash = std::hash<std::string>{}(file_name);
+    size_t seed = 1;
+    if (auto seed_from_options = G->settings().debug_info_force_disable_prob_seed.get()) {
```
mkornaukhov03: Hard to read, I'd prefer `size_t` instead of auto.
Also I'd check `!=0` explicitly, but up for you

compiler/kphp2cpp.cpp
```diff
@@ -270,6 +270,11 @@ int main(int argc, char *argv[]) {
              "extra-linker-flags", "KPHP_EXTRA_LDFLAGS", get_default_extra_ldflags());
   parser.add("C++ compiler debug level for building the output binary", settings->extra_cxx_debug_level,
              "debug-level", "KPHP_DEBUG_LEVEL");
+  parser.add("Probability to partially disable debug info for some random part of files",
```
mkornaukhov03, line 273: probability of, typo

compiler/code-gen/common.h
DrDet: added `size_t`

compiler/kphp2cpp.cpp
DrDet: fixed

mkornaukhov03 (approved):
lgtm

PR #1160, opened 2024-11-20
[k2] crypto cbc ciphers

irlgirl:


runtime-light/stdlib/crypto/crypto-functions.cpp
```diff
@@ -115,3 +110,175 @@ task_t<int64_t> f$openssl_verify(const string &data, const string &signature, co
 
   co_return 1;
 }
+
+Optional<tl::CipherAlgorithm> parse_algorithm(const string &method) {
+  if (strncasecmp(method.c_str(), "aes-128-cbc", strlen("aes-128-cbc")) == 0) {
```
apolyakov, line 115: Let's convert it to something more high-level:
```
if (std::ranges::equal(lhs, rhs, ichar_equals)) {...}
```
where 'lhs' and 'rhs' are `std::string_view`s

runtime-light/stdlib/crypto/crypto-functions.cpp
```diff
@@ -115,3 +110,175 @@ task_t<int64_t> f$openssl_verify(const string &data, const string &signature, co
 
   co_return 1;
 }
+
+Optional<tl::CipherAlgorithm> parse_algorithm(const string &method) {
```
apolyakov, line 114: Please, add `noexcept` to all the functions

apolyakov, line 114: Also, it's better to use `std::optional` unless you need to pass the object back to the user code

runtime-light/stdlib/crypto/crypto-functions.cpp
```diff
@@ -115,3 +110,175 @@ task_t<int64_t> f$openssl_verify(const string &data, const string &signature, co
...
+    return tl::CipherAlgorithm::AES128;
+  } else if (strncasecmp(method.c_str(), "aes-256-cbc", strlen("aes-256-cbc")) == 0) {
+    return tl::CipherAlgorithm::AES256;
+  }
+  return {};
+}
+
+int64_t algorithm_iv_len(tl::CipherAlgorithm algorithm __attribute__((unused))) {
+  /* since only aes-128/256-cbc supported for now */
+  return 16;
```
apolyakov, line 125: Let's get rid of magic constants

runtime-light/stdlib/crypto/crypto-functions.cpp
```diff
@@ -115,3 +110,175 @@ task_t<int64_t> f$openssl_verify(const string &data, const string &signature, co
...
+Optional<tl::CipherAlgorithm> parse_algorithm(const string &method) {
+  if (strncasecmp(method.c_str(), "aes-128-cbc", strlen("aes-128-cbc")) == 0) {
+    return tl::CipherAlgorithm::AES128;
+  } else if (strncasecmp(method.c_str(), "aes-256-cbc", strlen("aes-256-cbc")) == 0) {
+    return tl::CipherAlgorithm::AES256;
+  }
+  return {};
+}
+
+int64_t algorithm_iv_len(tl::CipherAlgorithm algorithm __attribute__((unused))) {
```
apolyakov, line 123: We are in C++20, so `[[maybe_unused]]` is a perfect choice here

runtime-light/stdlib/crypto/crypto-functions.cpp
```diff
@@ -115,3 +110,175 @@ task_t<int64_t> f$openssl_verify(const string &data, const string &signature, co
...
+}
+
+int64_t algorithm_key_len(tl::CipherAlgorithm algorithm) {
+  switch (algorithm) {
+    case tl::CipherAlgorithm::AES128: {
+      return 16;
+    }
+    case tl::CipherAlgorithm::AES256: {
+      return 32;
+    }
```
apolyakov, line 135: missing `default`

runtime-light/stdlib/crypto/crypto-functions.cpp
```diff
@@ -115,3 +110,175 @@ task_t<int64_t> f$openssl_verify(const string &data, const string &signature, co
...
+Optional<int64_t> f$openssl_cipher_iv_length(const string &method) noexcept {
+  auto algorithm = parse_algorithm(method);
+  if (algorithm.is_null()) {
+    php_warning("Unknown cipher algorithm");
+    return false;
+  }
+  return algorithm_iv_len(algorithm.val());
+}
+
+enum cipher_opts : int { OPENSSL_RAW_DATA = 1, OPENSSL_ZERO_PADDING = 2, OPENSSL_DONT_ZERO_PAD_KEY = 4 };
```
apolyakov, line 153: 1. Let's move non-public symbols into an anonymous namespace
2. Let's also either use `enum class` or distinct constants since I don't see any value from using `enum` here.
3. We also try to avoid implicit types like `int`

runtime-light/stdlib/crypto/crypto-functions.cpp
```diff
@@ -115,3 +110,175 @@ task_t<int64_t> f$openssl_verify(const string &data, const string &signature, co
...
+
+  auto query = f$component_client_send_request(string{CRYPTO_COMPONENT_NAME}, string{buffer.data(), static_cast<string::size_type>(buffer.size())});
+  string resp = co_await f$component_client_fetch_response(co_await query);
+
+  buffer.clean();
+  buffer.store_bytes({resp.c_str(), static_cast<size_t>(resp.size())});
+
+  std::optional<uint32_t> magic = buffer.fetch_trivial<uint32_t>();
+  // TODO: parse error?
+  if (!magic.has_value() || *magic != TL_STRING) {
```
apolyakov, line 226: `if (const auto magic = buffer.fetch_trivial<uint32_t>; magic.value_or(TL_ZERO) != TL_STRING)`

runtime-light/stdlib/crypto/crypto-functions.cpp
irlgirl: fux

irlgirl: fix

runtime-light/stdlib/crypto/crypto-functions.cpp
irlgirl: fix

runtime-light/stdlib/crypto/crypto-functions.cpp
irlgirl: fix

runtime-light/stdlib/crypto/crypto-functions.cpp
irlgirl: fix: use warning + `return 0`

runtime-light/stdlib/crypto/crypto-functions.cpp
irlgirl: fix

runtime-light/stdlib/crypto/crypto-functions.cpp
```diff
@@ -115,3 +110,192 @@ task_t<int64_t> f$openssl_verify(const string &data, const string &signature, co
 
   co_return 1;
 }
+
+static bool ichar_equals(char a, char b) noexcept {
```
apolyakov, line 114: Let's put it into an anonymous namespace. Also, it's also acceptable to define it as lambda inside `parse_algorithm`.

runtime-light/stdlib/crypto/crypto-functions.cpp
```diff
@@ -115,3 +110,192 @@ task_t<int64_t> f$openssl_verify(const string &data, const string &signature, co
...
+  return std::tolower(static_cast<unsigned char>(a)) == std::tolower(static_cast<unsigned char>(b));
+}
+
+static std::optional<tl::CipherAlgorithm> parse_algorithm(const string &method) noexcept {
```
apolyakov, line 118: We prefer defining something in an anonymous namespace

runtime-light/stdlib/crypto/crypto-functions.cpp
```diff
@@ -115,3 +110,192 @@ task_t<int64_t> f$openssl_verify(const string &data, const string &signature, co
...
+  using namespace std::string_view_literals;
+  std::string_view method_sv{method.c_str(), method.size()};
+
+  if (std::ranges::equal(method_sv, "aes-128-cbc"sv, ichar_equals)) {
```
apolyakov, line 122: It looks like you use constants like `aes-128-cbc` multiple times. Let's define something like:
`constexpr std::string_view AES_128_CBC = "aes-128-cbc";`

runtime-light/tl/tl-types.h
```diff
@@ -262,6 +262,27 @@ struct Dictionary final {
   }
 };
 
+struct String {
+  string value;
+
+  bool fetch(TLBuffer &tlb) noexcept {
+    if (tlb.lookup_trivial<uint32_t>().value_or(TL_ZERO) != TL_STRING) {
```
apolyakov, line 269: You can use `fetch_trivial` here. `fetch` is assumed to fetch something from TL buffer. It may fail, leaving the buffer in invalid state. There is distinct functionality to save/restore TL buffer's state.

apolyakov (changes requested):

runtime-light/stdlib/crypto/crypto-functions.cpp
irlgirl: +

runtime-light/stdlib/crypto/crypto-functions.cpp
irlgirl: +

runtime-light/tl/tl-types.h
irlgirl: +

apolyakov (approved):

Ahmed-su-do:
$pwd_short = "my password";
  $pwd_16 = "This_is_16_bytes";
  $pwd_32 =
?

PR #1167, opened 2024-11-27
K2: implement md5 function

apolyakov:


runtime-light/stdlib/crypto/crypto-functions.h
```diff
@@ -4,11 +4,29 @@
...
 #include "runtime-common/core/runtime-core.h"
 #include "runtime-light/coroutine/task.h"
 #include "runtime-light/stdlib/crypto/crypto-state.h"
+#include "runtime-light/stdlib/string/string-state.h"
 #include "runtime-light/tl/tl-types.h"
 
+inline string f$md5(const string &str, bool binary = false) noexcept {
+  constexpr auto MD5_HASH_LEN = 16;
+  string output{static_cast<string::size_type>(MD5_HASH_LEN * (binary ? 1 : 2)), false};
+  md5(reinterpret_cast<unsigned char *>(const_cast<char *>(str.c_str())), static_cast<int32_t>(str.size()), reinterpret_cast<unsigned char *>(output.buffer()));
```
mkornaukhov03, line 19: It's strange that `md5()` first parameter is not pointer to const, as I see it doesn't mutate. Maybe better change signature of it?

mkornaukhov03 (changes requested):

runtime-light/stdlib/crypto/crypto-functions.h
apolyakov: I agree. Added `const` qualifier

mkornaukhov03 (dismissed):
lgtm

PR #1169, opened 2024-11-28
[k2] move json-encoder builtins into runtime-common

irlgirl:


builtin-functions/kphp-light/unsupported/unsupported-serialize.txt
apolyakov, line 1: Not a stopper, but I'd appreciate if you put all these functions into `serialize.txt`. It's actually enough to just split supported/unsupported functions with comment

runtime-common/stdlib/string/from-json-processor.h
```diff
@@ -91,13 +90,13 @@ class FromJsonVisitor {
   }
 
   void do_set(JsonRawString &value, const mixed &json) noexcept {
-    kphp_runtime_context.static_SB.clean();
-    if (!impl_::JsonEncoder{0, false, get_json_obj_magic_key()}.encode(json)) {
+    RuntimeContext::get().static_SB.clean();
```
apolyakov, line 93: Please, minimise the number of static `get` method calls. It's never inlined. It's ok to make just one call and save a reference to a specific context in a variable

runtime-common/stdlib/string/string-context.h
```diff
@@ -45,6 +45,9 @@ class StringLibContext final : vk::not_copyable {
   // mb_string context
   bool detect_incorrect_encoding_names{};
 
+  // from-json-processor context
+  string last_json_processor_error;
```
apolyakov, line 49: I think we need separate context for (de)serialization routines. I'd like to prevent our contexts from being a trash for former global variables. Let's have something like `SerializationContext`. Also, let's create a new directory `stdlib/serialize/`

apolyakov: After adding a new context to `runtime-common`, please refer to how string contexts are included in k2 runtime

apolyakov (changes requested):
Minor notes only. Also, I think it's better to ask @astrophysik for a review as he's the last developer who worked on that code.

builtin-functions/kphp-light/unsupported/unsupported-serialize.txt
irlgirl: +

runtime-common/stdlib/string/from-json-processor.h
irlgirl: +

runtime-common/stdlib/string/string-context.h
irlgirl: +, introduce serparate SerializationContext in runtime-common/serialize/serialization-context.h

compiler/code-gen/declarations.cpp
```diff
@@ -906,10 +906,6 @@ void ClassDeclaration::compile_accept_json_visitor(CodeGenerator &W, ClassPtr kl
 }
 
 void ClassDeclaration::compile_accept_visitor_methods(CodeGenerator &W, ClassPtr klass) {
-  if (G->is_output_mode_k2()) {
-    // The current version of runtime-light does not support visitors
-    return;
-  }
   bool need_generic_accept =
```
astrophysik, line 909: Let's add condition to skip generation of visitors for `instance_cache` and `memory_estimate` cases. For now it generate c++ compile errors if I try code like `instance_cache_store("key", $value);`

compiler/code-gen/declarations.cpp
```diff
@@ -949,10 +945,6 @@ void ClassDeclaration::compile_accept_visitor_methods(CodeGenerator &W, ClassPtr
 }
 
 void ClassDeclaration::compile_msgpack_declarations(CodeGenerator &W, ClassPtr klass) {
-  if (G->is_output_mode_k2()) {
```
astrophysik, line 952: Why you remove this condition without adding definition of `vk::msgpack::packer` for runtime light? Compilation of classes with `/** @kphp-serializable */` tag fails

compiler/code-gen/declarations.cpp
```diff
@@ -1068,10 +1060,6 @@ void ClassDeclaration::compile_job_worker_shared_memory_piece_methods(CodeGenera
 }
 
 void ClassMembersDefinition::compile(CodeGenerator &W) const {
-  if (G->is_output_mode_k2()) {
```
astrophysik, line 1071: same as for `ClassDeclaration::compile_accept_visitor_methods`

astrophysik (changes requested):

compiler/code-gen/declarations.cpp
irlgirl: +

compiler/code-gen/declarations.cpp
irlgirl: +

compiler/code-gen/declarations.cpp
```diff
@@ -936,7 +932,7 @@ void ClassDeclaration::compile_accept_visitor_methods(CodeGenerator &W, ClassPtr
     compile_accept_visitor(W, klass, "CommonMemoryEstimateVisitor");
   }
 
-  if (klass->need_instance_cache_visitors) {
+  if (klass->need_instance_cache_visitors && !G->is_output_mode_k2()) {
```
astrophysik, line 935: Same for previous if with `need_instance_memory_estimate_visitor`

astrophysik (changes requested):

compiler/code-gen/declarations.cpp
irlgirl: +, thx!

astrophysik (approved):

PR #1179, opened 2024-12-09
Fix bug with traversal of directories during reading KML models

mkornaukhov03:
Previously we had segfault in case of `--kml-dir` parameter points to the directory that is owned by `root`, but the executor of compiled program is not `root`.

Related to [KPHP-1982](https://jira.mvk.com/browse/KPHP-1982)

### DIFF
--- a/runtime/kphp_ml/kphp_ml_init.cpp
+++ b/runtime/kphp_ml/kphp_ml_init.cpp
@@ -4,6 +4,7 @@
 
 #include "runtime/kphp_ml/kphp_ml_init.h"
 
+#include <cstring>
 #include <dirent.h>
 #include <exception>
 #include <string>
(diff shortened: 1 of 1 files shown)


runtime/kphp_ml/kphp_ml_init.cpp
```diff
@@ -47,24 +48,25 @@ static void load_kml_file(const std::string &path) {
 }
 
 static void traverse_kml_dir(const std::string &path) {
-  if (ends_with(path.c_str(), ".kml")) {
+  struct stat st{};
+  bool is_directory = stat(path.c_str(), &st) == 0 && S_ISDIR(st.st_mode);
```
Danil42Russia, line 52: maybe use `std::filesystem::is_directory()`?

PetrShumilov (approved):
LGTM

PR #1193, opened 2024-12-18
move bcmath functions to runtime common

astrophysik:


runtime-common/stdlib/math/bcmath.cpp
```diff
@@ -0,0 +1,923 @@
...
+  if (scale < 0) {
+    MathLibContext::get().bc_scale = 0;
+  } else {
+    MathLibContext::get().bc_scale = static_cast<int>(scale);
+  }
+}
+
+string f$bcdiv(const string &lhs_str, const string &rhs_str, int64_t scale) noexcept {
+  if (scale == std::numeric_limits<int64_t>::min()) {
+    scale = MathLibContext::get().bc_scale;
```
apolyakov, line 494: You can store a reference to math lib context in a local variable to prevent calling it multiple times in a single function.

apolyakov (changes requested):

runtime-common/stdlib/math/math-context.h
```diff
@@ -0,0 +1,47 @@
...
+  static MathLibContext &get() noexcept;
+};
+
+struct MathLibConstants final : vk::not_copyable {
+  string ONE{math_context_impl::ONE.data(), static_cast<string::size_type>(math_context_impl::ONE.size())};
+  string ZERO{math_context_impl::ZERO.data(), static_cast<string::size_type>(math_context_impl::ZERO.size())};
+  string TEN{math_context_impl::TEN.data(), static_cast<string::size_type>(math_context_impl::TEN.size())};
+  string ZERO_DOT_FIVE{math_context_impl::ZERO_DOT_FIVE.data(), static_cast<string::size_type>(math_context_impl::ZERO_DOT_FIVE.size())};
+
+  bcmath_impl_::BcNum BC_NUM_ONE{bcmath_impl_::bc_parse_number(ONE).first};
```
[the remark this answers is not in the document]
astrophysik: Actually it equals to for_global_const since string constants, but I make it clearer

runtime-common/stdlib/math/math-context.h
```diff
@@ -27,21 +36,16 @@ struct MathLibContext final : vk::not_copyable {
...
-  bcmath_impl_::BcNum BC_NUM_ZERO{bcmath_impl_::bc_parse_number(ZERO).first};
-  bcmath_impl_::BcNum BC_ZERO_DOT_FIVE{bcmath_impl_::bc_parse_number(ZERO_DOT_FIVE).first};
-
-  MathLibConstants() noexcept {
-    ONE.set_reference_counter_to(ExtraRefCnt::for_global_const);
-    ZERO.set_reference_counter_to(ExtraRefCnt::for_global_const);
-    TEN.set_reference_counter_to(ExtraRefCnt::for_global_const);
-    ZERO_DOT_FIVE.set_reference_counter_to(ExtraRefCnt::for_global_const);
-  }
+  string ONE{math_context_impl_::ONE.data(), static_cast<string::size_type>(math_context_impl_::ONE.size())};
```
apolyakov, line 39: Let's move this initialization to a `cpp` file as well. There is no need to keep it here. Also, it means that we may get rid of `math_context_impl_` namespace. You may move those constants to an anonymous namespace in `cpp` file.

apolyakov (approved):

PR #1195, opened 2024-12-19
Fix problem with similiar name in different runtime dirs

mkornaukhov03:
Previously, if there were two files with similar subpath in different runtime dirs, in case of K2, we had a failure during transpilation. This PR fixes it.


compiler/make/make.cpp
```diff
@@ -463,9 +466,19 @@ static std::vector<File *> build_runtime_and_common_from_sources(const std::stri
...
 
-  for (const auto *dir : std::vector<const Index*>{&runtime_core_dir, &runtime_dir, &common_dir, &unicode_dir}) {
+  std::unordered_set<size_t> files_hashes;
+  files_hashes.reserve(runtime_core_dir.get_files_count() + runtime_dir.get_files_count() + common_dir.get_files_count() + unicode_dir.get_files_count());
+
+  for (const auto *dir : std::vector<const Index *>{&runtime_core_dir, &runtime_dir, &common_dir, &unicode_dir}) {
     for (File *cpp_file : dir->get_files()) {
-      File *obj_file = obj_dir.insert_file(static_cast<std::string>(cpp_file->name_without_ext) + ".o");
+      auto obj_name = static_cast<std::string>(cpp_file->name_without_ext) + ".o";
+      if (unlikely(files_hashes.count(vk::std_hash(obj_name)) != 0)) {
```
apolyakov, line 475: Not a blocker, but isn't it better to use `contains` instead of `count != 0`?

apolyakov (dismissed):

compiler/make/make.cpp
mkornaukhov03: fixed

PR #1198, opened 2024-12-23
Fix usage of resumable function with ArrayAccess methods

mkornaukhov03:
Prohibit 2 things:
* Calling transitively resumable function
* Forking ArrayAccess methods

Previously, we had runtime problem in case of heavily using forks inside ArrayAccess methods.

### DIFF
--- a/compiler/pipes/calc-bad-vars.cpp
+++ b/compiler/pipes/calc-bad-vars.cpp
@@ -585,6 +585,13 @@ class CalcBadVars {
     }
     for (const auto &func : call_graph.functions) {
       func->can_be_implicitly_interrupted_by_other_resumable = into_resumable[func];
+      if (unlikely(func->class_id && func->class_id->is_required_interface && into_resumable[func])) {
+        std::string func_name = func->name;
+        std::replace(func_name.begin(), func_name.end(), '$', ':');
+        std::vector<std::string> chain;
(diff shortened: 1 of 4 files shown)


compiler/pipes/calc-bad-vars.cpp
```diff
@@ -591,7 +591,25 @@ class CalcBadVars {
         func->wait_prev = to_parents[func];
       }
     }
+
+    auto get_call_resumable_path = [&](FunctionPtr foo) {
```
Danil42Russia, line 595: why `foo`? o_O
IMHO, but `func` is better, but in general `function`

mkornaukhov03: Fixed

compiler/pipes/calc-bad-vars.cpp
```diff
@@ -591,7 +591,25 @@ class CalcBadVars {
         func->wait_prev = to_parents[func];
       }
     }
+
+    auto get_call_resumable_path = [&](FunctionPtr func) {
```
astrophysik, line 595: Boring comment - What if func will be null here?

astrophysik (commented):
LGTM

compiler/pipes/calc-bad-vars.cpp
mkornaukhov03: It cannot be null there because it's only called with valid function, line 612. I've added assert.

astrophysik (approved):

astrophysik (approved):

PR #1203, opened 2025-01-09
[k2] refactor TL library

apolyakov:
Main changes:

1. TL library now doesn't depend on `runtime-core`
2. TL library now is a "view" on the data being encoded/decoded (except TL types that need extra memory to hold their data, e.g. `tl::vector`)


runtime-light/core/std/containers.h
```diff
@@ -0,0 +1,51 @@
+//  Compiler for PHP (aka KPHP)
+//  Copyright (c) 2024 LLC «V Kontakte»
```
Danil42Russia, line 2: 2025? 🤣

Danil42Russia (dismissed):

runtime-light/stdlib/confdata/confdata-functions.cpp
```diff
@@ -25,17 +25,17 @@ namespace {
 
 constexpr std::string_view CONFDATA_COMPONENT_NAME = "confdata"; // TODO: it may actually have an alias specified in linking config
 
-mixed extract_confdata_value(tl::confdataValue &&confdata_value) noexcept {
-  if (confdata_value.is_php_serialized && confdata_value.is_json_serialized) { // check that we don't have both flags set
+mixed extract_confdata_value(tl::confdataValue confdata_value) noexcept {
```
astrophysik, line 28: Why not `const tl::confdataValue &`?

runtime-light/tl/tl-types.h
```diff
@@ -61,104 +60,132 @@ struct Maybe final {
...
+
+public:
+  std::string_view value;
+
+  bool fetch(TLBuffer &tlb) noexcept;
+
+  void store(TLBuffer &tlb) const noexcept;
+};
+
+struct String final {
```
astrophysik, line 83: Maybe better to name `string_view` and `string`? Instead of `string` and `String`

runtime-light/tl/tl-types.cpp
```diff
@@ -4,136 +4,154 @@
...
+    str_len = 0;
+    tlb.store_trivial<uint8_t>(str_len);
+  }
+  tlb.store_bytes({str_buf, str_len});
+
+  const auto total_len{size_len + str_len};
+  const auto total_len_with_padding{(total_len + 3) & ~3};
+  const auto padding{total_len_with_padding - total_len};
+
+  std::array padding_array{'\0', '\0', '\0', '\0'};
```
astrophysik, line 102: `constexpr std::array`?

apolyakov: Thanks, added `constexpr`

runtime-light/tl/tl-types.h
apolyakov: I think the current naming is ok since it fully reflect type naming in the TL scheme

astrophysik (approved):

PR #1205, opened 2025-01-10
prevent false-positive stack-buffer-overflow ASan detections on signal stack, attemp 2

PetrShumilov:
`__sanitizer_start_switch_fiber` and ` __sanitizer_finish_switch_fiber` are not sufficient for informing ASan about the alternate signal stack. In our case, when a signal occurs, the activation of the alternate signal stack mislead ASan about current actual stack and leads to incorrect behavior of `__asan_handle_no_return`:
```
==15381==WARNING: ASan is ignoring requested __asan_handle_no_return: stack top: 0x7f98b59d0000; bottom 0x00000bb21000; size: 0x7f98a9eaf000 (140293662502912)
False positive error reports may follow
For details see https://github.com/google/sanitizers/issues/189
```
Need to use explicit signal stack unpoisoning instead. 

Fix of #1201

### DIFF
--- a/server/php-runner.cpp
+++ b/server/php-runner.cpp
@@ -4,6 +4,7 @@
 
 #include "server/php-runner.h"
 
+#include <array>
 #include <cassert>
 #include <cerrno>
 #include <cstdlib>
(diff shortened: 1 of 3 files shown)


apolyakov (commented):
lgtm

apolyakov (approved):

PR #1208, opened 2025-01-15
[k2] make sort functions implementation runtime specific

astrophysik:


runtime-light/stdlib/array/array-functions.h
```diff
@@ -221,19 +424,108 @@ array<T> f$array_combine(const array<T1> &keys, const array<T> &values) {
...
+
+template<class T>
+task_t<void> f$asort(array<T> &a, int64_t flag = SORT_REGULAR) {
+  switch (flag) {
+    case SORT_REGULAR:
+      co_return co_await array_functions_impl_::sort<task_t<void>>(a, array_functions_impl_::sort_compare<T>(), false);
+    case SORT_NUMERIC:
+      co_return co_await array_functions_impl_::sort<task_t<void>>(a, array_functions_impl_::sort_compare_numeric<T>(), false);
+    case SORT_STRING:
+      co_return co_await array_functions_impl_::sort<task_t<void>>(a, array_functions_impl_::sort_compare_string<T>(), false);
```
mkornaukhov03, line 505: Why is it interruptible?
I thought that all sort function that are not receiving callable as argument should **not** be interruptible.
Am I wrong?

mkornaukhov03 (changes requested):

runtime-light/stdlib/array/array-functions.h
astrophysik: You are right
I did it interruptible for easy implementation. But maybe it would be better to make it a right way in this pr

mkornaukhov03: If it's a big deal then it's OK to leave as is.

runtime-light/stdlib/array/array-functions.h
```diff
@@ -5,16 +5,218 @@
...
+#include "runtime-common/stdlib/array/array-functions.h"
 #include "runtime-light/coroutine/task.h"
 #include "runtime-light/stdlib/math/random-functions.h"
 #include "runtime-light/utils/concepts.h"
 
+namespace dl {
+
+template<typename T, typename Comparator>
+requires(std::invocable<Comparator, T, T>) task_t<void> sort(T *begin_init, T *end_init, Comparator compare) noexcept {
+  auto compare_call = [compare]<typename U>(U lhs, U rhs) -> task_t<int64_t> {
```
apolyakov, line 22: Lambda coroutine should not have captured values as they're not saved in a coroutine frame

runtime-light/stdlib/array/array-functions.h
```diff
@@ -5,16 +5,218 @@
...
+
+    while (begin < end) {
+      const auto offset = (end - begin) >> 1;
+      swap(*begin, begin[offset]);
+
+      T *i = begin + 1;
+      T *j = end;
+
+      while (true) {
+        while (i < j && (co_await compare_call(*begin, *i)) > 0) {
```
apolyakov, line 47: We can get rid of `co_await` completely unless `Comparator` is async function.

runtime-light/stdlib/array/array-functions.h
```diff
@@ -5,16 +5,218 @@
...
+  }
+  co_return;
+}
+
+} // namespace dl
+
 namespace array_functions_impl_ {
 
+template<typename Result, typename U, typename Comparator>
+Result sort(array<U> &arr, Comparator comparator, bool renumber) noexcept {
```
apolyakov, line 87: It's unsafe to pass an array to a coroutine by reference. In this function you access internal pointers of the array. Having such an array passed by reference, we can face with a case of the array being changed while we were blocked after `co_await` call.

apolyakov (changes requested):

runtime-light/stdlib/array/array-functions.h
astrophysik: Yes, we can, but it require to create copy paste function without co_await. Is it good decision - I don't think so. In any case, we should make base sort functions not interruptible, it's part of a larger task

runtime-light/stdlib/array/array-functions.h
astrophysik: Sort functions works inplace

runtime-light/stdlib/array/array-functions.h
astrophysik: fix

runtime-light/stdlib/array/array-functions.h
astrophysik: fix

runtime-light/stdlib/array/array-functions.h
astrophysik: fix

runtime-light/stdlib/array/array-functions.h
astrophysik: make temporary copy

runtime/array_functions.h
```diff
@@ -720,3 +720,15 @@ T f$vk_dot_product(const array<T> &a, const array<T> &b) {
   }
   return vk_dot_product_sparse<T>(a, b);
 }
+
+template<typename Result, typename U, typename Comparator>
+Result array_functions_impl_::async_sort(__attribute__((unused)) array<U> & arr, __attribute__((unused)) Comparator comparator, __attribute__((unused)) bool renumber) noexcept {
```
mkornaukhov03, line 725: NIT: `[[maybe_unused]]` is better

mkornaukhov03 (changes requested):
Why  `cl/156_recursive_assumptions.php` is not enabled for k2?

astrophysik:
> Why `cl/156_recursive_assumptions.php` is not enabled for k2?

Now is turned on

mkornaukhov03 (changes requested):
LGTM

runtime-light/stdlib/array/array-functions.h
```diff
@@ -5,16 +5,201 @@
...
 #include <functional>
 #include <utility>
 
 #include "runtime-common/core/runtime-core.h"
+#include "runtime-common/stdlib/array/array-functions.h"
 #include "runtime-light/coroutine/task.h"
 #include "runtime-light/stdlib/math/random-functions.h"
 #include "runtime-light/utils/concepts.h"
 
+namespace {
```
apolyakov, line 18: Anonymous namespaces in header files can lean to name clashing. Let's too its content into `array_functions_impl_` namespace

astrophysik: fix

apolyakov (approved):

mkornaukhov03 (approved):
ok

PR #1209, opened 2025-01-15
[k2] move ip functions to common

astrophysik:


runtime-common/stdlib/server/net-functions.cpp
```diff
@@ -0,0 +1,40 @@
...
+
+  const size_t buf_size = 25;
+  char buf[buf_size];
+  int len = snprintf(buf, buf_size, "%u", ntohl(result.s_addr));
+  return string(buf, len);
+}
+
+string f$long2ip(int64_t num) noexcept {
+  auto &runtime_context{RuntimeContext::get()};
+  runtime_context.static_SB.clean().reserve(100);
```
Danil42Russia, line 32: why 100? If the maximum for IPv4 is 15.

runtime-common/stdlib/server/net-functions.cpp
```diff
@@ -0,0 +1,40 @@
...
+  for (int i = 3; i >= 0; i--) {
```
Danil42Russia, line 33: are we sure we don't want to use the off-the-shelf `inet_ntop` function for the conversion?

runtime-common/stdlib/server/net-functions.cpp
```diff
@@ -0,0 +1,40 @@
+// Compiler for PHP (aka KPHP)
+// Copyright (c) 2024 LLC «V Kontakte»
```
mkornaukhov03, line 2: 2025

astrophysik: fix

runtime-common/stdlib/server/net-functions.cpp
```diff
@@ -0,0 +1,40 @@
...
+
+  const size_t buf_size = 25;
+  char buf[buf_size];
+  int len = snprintf(buf, buf_size, "%u", ntohl(result.s_addr));
+  return string(buf, len);
+}
+
+string f$long2ip(int64_t num) noexcept {
+  auto &runtime_context{RuntimeContext::get()};
+  runtime_context.static_SB.clean().reserve(100);
```
mkornaukhov03, line 32: I thinks it's better to define this as const/constexpr with comments, why it's more than 15. Possibly, it's used in other places?

mkornaukhov03 (approved):
OK, minor suggestion

mkornaukhov03 (approved):
OK

runtime-common/stdlib/server/net-functions.cpp
astrophysik: fix

PR #1211, opened 2025-01-16
[k2] add missing RPC bultins

apolyakov:
This PR adds implementation of following builtins:

- rpc_tl_pending_queries_count
- rpc_tl_query_result_synchronously
- set_fail_rpc_on_int32_overflow

### DIFF
--- a/runtime-light/stdlib/rpc/rpc-api.h
+++ b/runtime-light/stdlib/rpc/rpc-api.h
@@ -134,7 +134,9 @@ inline task_t<array<int64_t>> f$rpc_tl_query(const class_instance<C$RpcConnectio
 
 // === Rpc Misc ===================================================================================
 
-void f$rpc_clean() noexcept;
+inline void f$rpc_clean() noexcept {
+  RpcInstanceState::get().rpc_buffer.clean();
+}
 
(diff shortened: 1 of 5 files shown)


runtime-light/stdlib/rpc/rpc-api.h
```diff
@@ -129,7 +138,17 @@ inline task_t<array<int64_t>> f$rpc_tl_query(const class_instance<C$RpcConnectio
...
+  RpcInstanceState::get().rpc_buffer.clean();
+}
+
+inline int64_t f$rpc_tl_pending_queries_count() noexcept {
+  return RpcInstanceState::get().response_waiter_forks.size();
+}
+
+inline constexpr bool f$set_fail_rpc_on_int32_overflow(bool /*unused*/) noexcept {
+  return true;
+}
```
mkornaukhov03, line 151: Why is it noop?
For now it leads to additional check in tl layer with tl error or warning

mkornaukhov03 (changes requested):

runtime-light/stdlib/rpc/rpc-api.h
apolyakov: You're right, fixed

mkornaukhov03 (approved):
OK

PR #1212, opened 2025-01-16
Implement hash builtins for K2

mkornaukhov03:
This PR add support for the following builtins in K2 mode:
* hash()
* hash_hmac()
* hash_algos()
* hash_hmac_algos()
* hash_equals()
* sha1()

Also one hash test was divided into several.

### DIFF
--- a/runtime-common/stdlib/hash/hash-functions.h
+++ b/runtime-common/stdlib/hash/hash-functions.h
@@ -1,0 +1,7 @@
+// Compiler for PHP (aka KPHP)
+// Copyright (c) 2025 LLC «V Kontakte»
+// Distributed under the GPL v3 License, see LICENSE.notice.txt
+
+#include "runtime-common/core/runtime-core.h"
+
+bool f$hash_equals(const string &known_string, const string &user_string) noexcept;
--- a/runtime-common/stdlib/string/string-functions.h
+++ b/runtime-common/stdlib/string/string-functions.h
@@ -7,6 +7,7 @@
 #include <cstdint>
 #include <cstring>
 #include <limits>
+#include <string_view>
 
 #include "runtime-common/core/runtime-core.h"
 #include "runtime-common/core/utils/kphp-assert-core.h"
@@ -18,7 +19,7 @@ string f$addslashes(const string &str) noexcept;
 
 string f$hex2bin(const string &str) noexcept;
 
-inline string f$bin2hex(const string &str) noexcept {
+inline string bin2hex_impl(std::string_view str) noexcept {
   int len = str.size();
   string result(2 * len, false);
 
@@ -28,6 +29,10 @@ inline string f$bin2hex(const string &str) noexcept {
   }
 
   return result;
+}
+
+inline string f$bin2hex(const string &str) noexcept {
+  return bin2hex_impl(std::string_view{str.c_str(), str.size()});
 }
 
 string f$convert_cyr_string(const string &str, const string &from_s, const string &to_s) noexcept;
--- a/runtime-light/stdlib/crypto/crypto-functions.cpp
+++ b/runtime-light/stdlib/crypto/crypto-functions.cpp
@@ -100,7 +100,7 @@ task_t<Optional<array<mixed>>> f$openssl_x509_parse(const string &data, bool sho
 task_t<bool> f$openssl_sign(const string &data, string &signature, const string &private_key, int64_t algo) noexcept {
   tl::DigestSign request{.data = {.value = {data.c_str(), data.size()}},
                          .private_key = {.value = {private_key.c_str(), private_key.size()}},
-                         .algorithm = static_cast<tl::DigestAlgorithm>(algo)};
+                         .algorithm = static_cast<tl::HashAlgorithm>(algo)};
 
   tl::TLBuffer buffer;
   request.store(buffer);
@@ -127,7 +127,7 @@ task_t<bool> f$openssl_sign(const string &data, string &signature, const string 
 task_t<int64_t> f$openssl_verify(const string &data, const string &signature, const string &pub_key, int64_t algo) noexcept {
   tl::DigestVerify request{.data = {.value = {data.c_str(), data.size()}},
                            .public_key = {.value = {pub_key.c_str(), pub_key.size()}},
-                           .algorithm = static_cast<tl::DigestAlgorithm>(algo),
+                           .algorithm = static_cast<tl::HashAlgorithm>(algo),
                            .signature = {.value = {signature.c_str(), signature.size()}}};
 
   tl::TLBuffer buffer;
--- a/runtime-light/stdlib/crypto/crypto-functions.cpp
+++ b/runtime-light/stdlib/crypto/crypto-functions.cpp
@@ -343,3 +343,17 @@ task_t<Optional<string>> f$openssl_decrypt(string data, const string &method, co
   }
   co_return string{response.inner.value.data(), static_cast<string::size_type>(response.inner.value.size())};
 }
+
+array<string> f$hash_algos() noexcept {
+  static constexpr const char *algos[] = {"md5", "sha1", "sha224", "sha256", "sha384", "sha512"};
+  array<string> response;
+  response.reserve(std::size(algos), true);
+  for (const char *algo : algos) {
+    response.push_back(string(algo));
+  }
+  return response;
+}
+
+array<string> f$hash_hmac_algos() noexcept {
+  return f$hash_algos();
+}
--- a/runtime-light/stdlib/crypto/crypto-functions.cpp
+++ b/runtime-light/stdlib/crypto/crypto-functions.cpp
@@ -4,11 +4,14 @@
 
 #include "runtime-light/stdlib/crypto/crypto-functions.h"
 
+#include <algorithm>
 #include <cstddef>
 #include <string_view>
 
 #include "common/tl/constants/common.h"
+#include "runtime-common/core/utils/kphp-assert-core.h"
 #include "runtime-common/stdlib/server/url-functions.h"
+#include "runtime-common/stdlib/string/string-functions.h"
 #include "runtime-light/stdlib/component/component-api.h"
 #include "runtime-light/tl/tl-core.h"
 #include "runtime-light/tl/tl-functions.h"
@@ -155,7 +158,7 @@ namespace {
 constexpr std::string_view AES_128_CBC = "aes-128-cbc";
 constexpr std::string_view AES_256_CBC = "aes-256-cbc";
 
-std::optional<tl::CipherAlgorithm> parse_algorithm(const string &method) noexcept {
+std::optional<tl::CipherAlgorithm> parse_cipher_algorithm(const string &method) noexcept {
   using namespace std::string_view_literals;
   std::string_view method_sv{method.c_str(), method.size()};
 
@@ -237,7 +240,7 @@ array<string> f$openssl_get_cipher_methods([[maybe_unused]] bool aliases) noexce
 }
 
 Optional<int64_t> f$openssl_cipher_iv_length(const string &method) noexcept {
-  auto algorithm = parse_algorithm(method);
+  auto algorithm = parse_cipher_algorithm(method);
   if (!algorithm.has_value()) {
     php_warning("Unknown cipher algorithm");
     return false;
@@ -247,7 +250,7 @@ Optional<int64_t> f$openssl_cipher_iv_length(const string &method) noexcept {
 
 task_t<Optional<string>> f$openssl_encrypt(const string &data, const string &method, const string &source_key, int64_t options, const string &source_iv,
                                            string &tag, const string &aad, int64_t tag_length __attribute__((unused))) noexcept {
-  auto algorithm = parse_algorithm(method);
+  auto algorithm = parse_cipher_algorithm(method);
   if (!algorithm.has_value()) {
     php_warning("Unknown cipher algorithm");
     co_return false;
@@ -303,7 +306,7 @@ task_t<Optional<string>> f$openssl_decrypt(string data, const string &method, co
     data = std::move(decoding_data.val());
   }
 
-  auto algorithm = parse_algorithm(method);
+  auto algorithm = parse_cipher_algorithm(method);
   if (!algorithm.has_value()) {
     php_warning("Unknown cipher algorithm");
     co_return false;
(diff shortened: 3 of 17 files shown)


runtime-common/stdlib/string/string-functions.h
```diff
@@ -18,7 +19,7 @@ string f$addslashes(const string &str) noexcept;
 
 string f$hex2bin(const string &str) noexcept;
 
-inline string f$bin2hex(const string &str) noexcept {
+inline string bin2hex_impl(std::string_view str) noexcept {
```
astrophysik, line 22: Let's add it namespace like `string_functions_impl_`

astrophysik (changes requested):

runtime-common/stdlib/string/string-functions.h
mkornaukhov03: Fixed

runtime-light/stdlib/crypto/crypto-functions.cpp
```diff
@@ -343,3 +347,97 @@ task_t<Optional<string>> f$openssl_decrypt(string data, const string &method, co
   }
   co_return string{response.inner.value.data(), static_cast<string::size_type>(response.inner.value.size())};
 }
+
+namespace {
+constexpr std::pair<const char *, tl::HashAlgorithm> HASH_ALGOS[] = {{"md5", tl::HashAlgorithm::MD5},       {"sha1", tl::HashAlgorithm::SHA1},
```
apolyakov, line 352: Why not use `std::array` or `std::span`?

runtime-light/stdlib/crypto/crypto-functions.cpp
```diff
@@ -343,3 +347,97 @@ task_t<Optional<string>> f$openssl_decrypt(string data, const string &method, co
...
+                                                                     {"sha224", tl::HashAlgorithm::SHA224}, {"sha256", tl::HashAlgorithm::SHA256},
+                                                                     {"sha384", tl::HashAlgorithm::SHA384}, {"sha512", tl::HashAlgorithm::SHA512}};
+
+std::optional<tl::HashAlgorithm> parse_hash_algorithm(const string &algo_str) noexcept {
```
apolyakov, line 356: It looks redundant... Do you really need such a complexity to just check that an array contains a string?

runtime-light/stdlib/crypto/crypto_schema.tl
```diff
@@ -9,7 +9,7 @@ resultFalse#27930a7b {t:Type} = Maybe t;
 resultTrue#3f9c8ef8 {t:Type} result:t = Maybe t;
 
 // Vector
-vector#1cb5c415 {t:Type} # [t] = Vector t;
```
apolyakov, line 12: Is it intended change?

runtime-light/tl/tl-functions.h
```diff
@@ -39,6 +39,8 @@ inline constexpr uint32_t DIGEST_SIGN_MAGIC = 0xd345'f658;
 inline constexpr uint32_t DIGEST_VERIFY_MAGIC = 0x5760'bd0e;
 inline constexpr uint32_t CBC_DECRYPT_MAGIC = 0x7f2e'e1e4;
 inline constexpr uint32_t CBC_ENCRYPT_MAGIC = 0x6d4e'e36a;
```
apolyakov, line 41: It looks like these magics are not needed outside of `store/fetch` implementations. Can we hide them?

apolyakov (changes requested):

runtime-light/stdlib/crypto/crypto-functions.h
```diff
@@ -44,3 +30,24 @@ task_t<Optional<string>> f$openssl_encrypt(const string &data, const string &met
                                            int64_t tag_length = 16) noexcept;
 task_t<Optional<string>> f$openssl_decrypt(string data, const string &method, const string &key, int64_t options = 0, const string &iv = string{},
                                            string tag = string{}, const string &aad = string{}) noexcept;
+
+array<string> f$hash_algos() noexcept;
```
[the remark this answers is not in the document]
mkornaukhov03: It's auto formatted. What is wrong with formatting?

runtime-light/stdlib/crypto/crypto-functions.cpp
```diff
@@ -343,3 +347,97 @@ task_t<Optional<string>> f$openssl_decrypt(string data, const string &method, co
...
+  buffer.store_bytes({resp.c_str(), static_cast<size_t>(resp.size())});
+
+  tl::String response{};
+  // TODO: parse error?
+  if (!response.fetch(buffer)) {
+    co_return string("");
+  }
+
+  if (!raw_output) {
+    co_return string_functions_impl_::bin2hex(response.inner.value);
```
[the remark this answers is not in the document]
mkornaukhov03: Why can't I?
I did it according to the following logic: in string-functions.h there are builtins plus some helpers. This helper is better because using f$bin2hex leads to extra allocation

runtime-light/stdlib/crypto/crypto-functions.cpp
mkornaukhov03: What exactly do you consider redundant?
`find_if` instead of for loop?

runtime-light/tl/tl-functions.h
mkornaukhov03: Yes, but I don't think it's a point, for now at least. Confdata magics should be hidden in this case, too, but there is out of scope of current PR

runtime-light/stdlib/crypto/crypto-functions.h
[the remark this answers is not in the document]
apolyakov: I'd like to see a bit more of newlines between distincts declarations

runtime-light/stdlib/crypto/crypto-functions.cpp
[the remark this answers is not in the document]
apolyakov: imo you should not access private APIs. If you really need it, rework it so it can be made public. For example, you may look at `kphp::zlib::encode`

runtime-light/stdlib/crypto/crypto-functions.cpp
apolyakov: I'm just curious whether we can simplify it a bit

runtime-light/stdlib/crypto/crypto-functions.cpp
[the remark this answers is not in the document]
apolyakov: Another example is `kphp::http::header`

runtime-light/stdlib/crypto/crypto-functions.h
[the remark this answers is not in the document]
mkornaukhov03: fix

runtime-light/stdlib/crypto/crypto-functions.h
```diff
@@ -44,3 +30,24 @@ task_t<Optional<string>> f$openssl_encrypt(const string &data, const string &met
...
+array<string> f$hash_hmac_algos() noexcept;
+task_t<string> f$hash(const string &algo_str, const string &s, bool raw_output = false) noexcept;
+task_t<string> f$hash_hmac(const string &algo_str, const string &s, const string &key, bool raw_output = false) noexcept;
+task_t<string> f$sha1(const string &s, bool raw_output = false) noexcept;
```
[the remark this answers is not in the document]
mkornaukhov03: fix

runtime-light/stdlib/crypto/crypto-functions.cpp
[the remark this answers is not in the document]
mkornaukhov03: fix

runtime-light/stdlib/crypto/crypto-functions.cpp
mkornaukhov03: fix

runtime-light/stdlib/crypto/crypto-functions.cpp
mkornaukhov03: Idk how, as for me this code is simple enough. It can be rewritten with additional 2 for loops, but it wouldn't become simpler.

runtime-light/stdlib/crypto/crypto_schema.tl
mkornaukhov03: fixed

runtime-common/stdlib/hash/hash-functions.h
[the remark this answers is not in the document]
mkornaukhov03: fixed

runtime-common/stdlib/crypto/crypto-functions.cpp
```diff
@@ -0,0 +1,19 @@
+// Compiler for PHP (aka KPHP)
+// Copyright (c) 2025 LLC «V Kontakte»
+// Distributed under the GPL v3 License, see LICENSE.notice.txt
+
+#include "runtime-common/stdlib/crypto/crypto-functions.h"
```
apolyakov, line 5: Let's add include of `runtime-core.h`. Indirect includes can lead to compilation fails in case of some specific modifications in the header file

runtime-common/stdlib/string/string-functions.h
```diff
@@ -7,18 +7,16 @@
...
 #include "runtime-common/core/utils/kphp-assert-core.h"
 #include "runtime-common/stdlib/string/string-context.h"
 
-string f$addcslashes(const string &str, const string &what) noexcept;
-
-string f$addslashes(const string &str) noexcept;
-
-string f$hex2bin(const string &str) noexcept;
+namespace kphp {
+namespace string_functions {
```
apolyakov, line 17: I'd suggest to name it `strings`. `kphp::strings::bin2hex` looks quite good

runtime-light/stdlib/crypto/crypto-functions.cpp
```diff
@@ -343,3 +348,100 @@ task_t<Optional<string>> f$openssl_decrypt(string data, const string &method, co
...
+
+namespace {
+constexpr std::array<std::pair<const char *, tl::HashAlgorithm>, 6> HASH_ALGOS = {{{"md5", tl::HashAlgorithm::MD5},
+                                                                                   {"sha1", tl::HashAlgorithm::SHA1},
+                                                                                   {"sha224", tl::HashAlgorithm::SHA224},
+                                                                                   {"sha256", tl::HashAlgorithm::SHA256},
+                                                                                   {"sha384", tl::HashAlgorithm::SHA384},
+                                                                                   {"sha512", tl::HashAlgorithm::SHA512}}};
+
+std::optional<tl::HashAlgorithm> parse_hash_algorithm(string algo_str) noexcept {
```
apolyakov, line 360: For example, 
```
constexpr std::array<std::pair<std::string_view, tl::HashAlgorithm>, 6> HASH_ALGOS = {{{"md5", tl::HashAlgorithm::MD5},
                                                                                       {"sha1", tl::HashAlgorithm::SHA1},
                                                                                       {"sha224", tl::HashAlgorithm::SHA224},
                                                                                       {"sha256", tl::HashAlgorithm::SHA256},
                                                                                       {"sha384", tl::HashAlgorithm::SHA384},
                                                                                       {"sha512", tl::HashAlgorithm::SHA512}}};

std::optional<tl::HashAlgorithm> parse_hash_algorithm(std::string_view user_algo) noexcept {
  const auto *it{std::ranges::find_if(
    HASH_ALGOS,
    [user_algo = std::ranges::transform_view(user_algo, [](auto c) { return std::tolower(c); })](auto hash_algo) noexcept {
      return std::ranges::equal(user_algo, hash_algo);
    },
    [](const auto &hash_algo) noexcept { return hash_algo.first; })};

  return it != nullptr && it != HASH_ALGOS.end() ? std::optional{it->second} : std::nullopt;
}

```

runtime-light/stdlib/crypto/crypto-functions.cpp
```diff
@@ -343,3 +348,100 @@ task_t<Optional<string>> f$openssl_decrypt(string data, const string &method, co
...
+  const auto *it = std::find_if(std::begin(HASH_ALGOS), std::end(HASH_ALGOS),
+                                [&](auto pair) { return std::ranges::equal(algo_sv, std::string_view{pair.first}, ichar_equals); });
+
+  if (it == std::end(HASH_ALGOS)) {
+    return std::nullopt;
+  }
+  return it->second;
+}
+
+task_t<string> send_and_get_string(tl::TLBuffer &&buffer, bool raw_output) {
```
apolyakov, line 374: Is there a reason for rvalue reference?

runtime-light/stdlib/crypto/crypto-functions.cpp
```diff
@@ -343,3 +348,100 @@ task_t<Optional<string>> f$openssl_decrypt(string data, const string &method, co
...
+  request.store(buffer);
+
+  co_return co_await send_and_get_string(std::move(buffer), raw_output);
+}
+
+} // namespace
+
+array<string> f$hash_algos() noexcept {
+  array<string> response;
+  response.reserve(std::size(HASH_ALGOS), true);
```
apolyakov, line 409: why not just `HASH_ALGOS.size()`?

runtime-light/stdlib/crypto/crypto-functions.cpp
```diff
@@ -343,3 +348,100 @@ task_t<Optional<string>> f$openssl_decrypt(string data, const string &method, co
...
+  request.store(buffer);
+
+  co_return co_await send_and_get_string(std::move(buffer), raw_output);
+}
+
+task_t<string> f$sha1(string s, bool raw_output) noexcept {
+  co_return co_await hash_impl(tl::HashAlgorithm::SHA1, s, raw_output);
+}
+
+int64_t f$crc32(string s) {
```
apolyakov, line 445: `const string &`?

runtime-light/stdlib/crypto/crypto-functions.h
```diff
@@ -44,3 +30,29 @@ task_t<Optional<string>> f$openssl_encrypt(const string &data, const string &met
...
+
+array<string> f$hash_hmac_algos() noexcept;
+
+task_t<string> f$hash(string algo_str, string s, bool raw_output = false) noexcept;
+
+task_t<string> f$hash_hmac(string algo_str, string s, string key, bool raw_output = false) noexcept;
+
+task_t<string> f$sha1(string s, bool raw_output = false) noexcept;
+
+inline string f$md5(string str, bool binary = false) noexcept {
```
apolyakov, line 44: `const string &`?

runtime-light/stdlib/crypto/crypto-functions.h
```diff
@@ -44,3 +30,29 @@ task_t<Optional<string>> f$openssl_encrypt(const string &data, const string &met
...
+    const auto &lhex_digits{StringImageState::get().lhex_digits};
+    for (int64_t i = MD5_HASH_LEN - 1; i >= 0; --i) {
+      output[2 * i + 1] = lhex_digits[output[i] & 0x0F];
+      output[2 * i] = lhex_digits[(output[i] >> 4) & 0x0F];
+    }
+  }
+  return output;
+}
+
+int64_t f$crc32(string s);
```
apolyakov, line 58: `const string &`?

apolyakov (changes requested):

runtime-common/stdlib/crypto/crypto-functions.cpp
mkornaukhov03: fixed

runtime-light/stdlib/crypto/crypto-functions.cpp
mkornaukhov03: fixed

runtime-light/stdlib/crypto/crypto-functions.cpp
mkornaukhov03: Yes, it's semantically borrows buffer.

runtime-light/stdlib/crypto/crypto-functions.cpp
mkornaukhov03: fixed

runtime-light/stdlib/crypto/crypto-functions.h
mkornaukhov03: fixed

runtime-light/stdlib/crypto/crypto-functions.h
mkornaukhov03: fixed

runtime-common/stdlib/string/string-functions.h
apolyakov: I meant `strings`, not `string`

runtime-light/stdlib/crypto/crypto-functions.cpp
apolyakov: The only outcome of your decision is that you force everyone to write `std::move(buffer)`. Nothing else happens

runtime-light/stdlib/crypto/crypto-functions.cpp
```diff
@@ -343,3 +351,98 @@ task_t<Optional<string>> f$openssl_decrypt(string data, const string &method, co
...
+  request.store(buffer);
+
+  co_return co_await send_and_get_string(std::move(buffer), raw_output);
+}
+
+task_t<string> f$sha1(string s, bool raw_output) noexcept {
+  co_return co_await hash_impl(tl::HashAlgorithm::SHA1, s, raw_output);
+}
+
+int64_t f$crc32(const string &s) {
```
apolyakov, line 446: missing `noexcept`. Please take a look at other functions too

runtime-light/stdlib/crypto/crypto-functions.cpp
mkornaukhov03: Fixed

runtime-common/stdlib/string/string-functions.h
```diff
@@ -14,7 +14,7 @@
 #include "runtime-common/stdlib/string/string-context.h"
 
 namespace kphp {
-namespace string {
+namespace strings {
 
 inline class string bin2hex(std::string_view str) noexcept {
```
apolyakov, line 19: Now `class` should be redundant here

apolyakov (changes requested):

runtime-light/tl/tl-core.h
```diff
@@ -9,16 +9,16 @@
...
-#include "common/mixin/not_copyable.h"
 #include "runtime-common/core/utils/kphp-assert-core.h"
 #include "runtime-light/allocator/allocator.h"
 #include "runtime-light/core/std/containers.h"
 #include "runtime-light/utils/concepts.h"
 
 namespace tl {
 
-class TLBuffer final : private vk::not_copyable {
+class TLBuffer final {
```
apolyakov, line 21: I'm almost sure it's not necessary to extend `TLBuffer`'s API for now. You may want to rethink the way you use it in your code in case you're facing troubles with current API

apolyakov (changes requested):

runtime-light/tl/tl-core.h
mkornaukhov03: What is problem of allowing moving TLBuffer?

apolyakov: We thus make it easier for `TLBuffer` to outlive a data it views.

apolyakov: Sorry, I was thinking about that with another `TLBuffer`'s design in my mind. I want to get rid of `TLBuffer::store_bytes` or at least limit its usage. I think that it's not necessary to store bytes in order to fetch some TL type from it. It should be enough to just init `TLBuffer` with a view on byte sequence

mkornaukhov03: This class is called TL_Buffer_, so I assume that it is owner of all its memory. If I got you right, you wanna builder-like behaviour. I think it should be separate class

mkornaukhov03: BTW, such a TLBuilder in general case cannot outlive async operation because views inside it can become changed

apolyakov: Ok. There are 2 cases for now:

1. add move-semantics to `TLBuffer`
2. just pass it as reference (which is safe in your case)

mkornaukhov03: I think mone semantics is useful in anycase.

In my case I prefer move semantics because really that function semantically  does not guarantee anything about tl buffer. That is why initially there were double ref.

mkornaukhov03: Why?
What's wrong with awaitables' move assignment?
And with TLBuffer's move assignment operator?

apolyakov: I just don't want to add additional functionality unless it's necessary

mkornaukhov03: Fixed

apolyakov (approved):
lgtm

common/crc32_generic.h
```diff
@@ -0,0 +1,98 @@
+// Compiler for PHP (aka KPHP)
+// Copyright (c) 2025 LLC «V Kontakte»
+// Distributed under the GPL v3 License, see LICENSE.notice.txt
+
+#pragma once
+
+constexpr unsigned int crc32_table[256] =
```
astrophysik, line 7: Maybe use `std::array` in such cases?

mkornaukhov03: I've decided not to do modification apart from changing `const` on `constexpr`

PR #1213, opened 2025-01-16
Support CompileTimeLocation in K2

mkornaukhov03:


astrophysik (commented):
LGTM

mkornaukhov03:
> LGTM

Github doesn't consider it as an approve =(

runtime-light/stdlib/compile-time-location/compile-time-location.h
apolyakov, line 1: Is there a reason for that class not being moved to `runtime-common`?

apolyakov (changes requested):

runtime-light/stdlib/compile-time-location/compile-time-location.h
mkornaukhov03: Yes, I thought about this. The cause is different `accept` methods, see `CommonDefaultVisitorMethods` and `DummyVisitorMethods`.

apolyakov: So why don't we use `DummyVisitorMethods`?

mkornaukhov03: Because it depends on visitors that are defined only in runtime dir.

apolyakov: Can we define them in `runtime-common`?

mkornaukhov03: Technically yes. But we need to define some other things, too. For example, some structures from `runtime/ffi.h" and `runtime/critical_section.h`. I think copypasting 30 lines of code for implementing required feature is better.

apolyakov: Let's then just put it under `stdlib/exception/` folder.

mkornaukhov03: For what?
It is not connected with exception at all

apolyakov: It looks like it's used for error reporting mainly. I'd put it under `exception` folder. Another nice option IMO is to create the `error/` folder and put both CT location and exception there

mkornaukhov03: It's wrong. It is separate feature that can be used in any way. In `197_compile_time_location.php` it is used for logging.

apolyakov: I don't like that we have separate folders for exceptions and CT location. Let's join them. `diagnostic/`?

mkornaukhov03: Fixed

apolyakov (approved):
lgtm

PR #1218, opened 2025-01-18
Fix problem with fmt lib in bookworm

mkornaukhov03:
Default fmtlib version in bookworm gives warning when passing enums to fmt_* functions. Behavior is not changed.

### DIFF
--- a/compiler/pipes/calc-const-types.cpp
+++ b/compiler/pipes/calc-const-types.cpp
@@ -4,6 +4,7 @@
 
 #include "compiler/pipes/calc-const-types.h"
 
+#include "common/wrappers/to_underlying.h"
 #include "compiler/data/class-data.h"
 #include "compiler/data/src-file.h"
 #include "compiler/data/var-data.h"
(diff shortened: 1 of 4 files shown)


compiler/pipes/calc-const-types.cpp
```diff
@@ -63,7 +64,7 @@ VertexPtr CalcConstTypePass::on_exit_vertex(VertexPtr v) {
       break;
     }
     default:
-      kphp_error (0, fmt_format("Unknown cnst-type for [op = {}]", v->type()));
+      kphp_error (0, fmt_format("Unknown cnst-type for [op = {}]", vk::to_underlying(v->type())));
```
apolyakov, line 67: Instead of adding a function with unknown usage perspective, I'd just use `static_cast<std::to_underlying_t<decltype(v->type())>>(v->type())` here

mkornaukhov03: Disagree. 

> unknown usage
It'a a value judgment. Such a function is accepted in C++23, so I think it's useful wrapper and some programmers khow how to use it.

Using pure static cast with decltypes is too verbose and reduces readability, as for me.  I think it's a matter of taste and not really the deal for discussion.

apolyakov (approved):

PR #1222, opened 2025-01-21
[k2] fix runtime compilation from sources 

astrophysik:


compiler/make/make.cpp
```diff
@@ -451,6 +451,7 @@ static std::string get_light_runtime_compiler_options() {
     }
   }
   s << "-std=c++20 ";
+  s << "-Wno-invalid-source-encoding "; // required by vkext/string-processing.cpp
```
apolyakov, line 454: Let's add that flag for that single file only

apolyakov (changes requested):

compiler/make/make.cpp
astrophysik: Done

apolyakov (approved):

PR #1224, opened 2025-01-22
Fix ubuntu 24.04 noble build

mkornaukhov03:
This PR allows to build all targets using `clang++-18` and `g++-11`.
`g++-12` and newer are going crazy with too strict warnings and they are _won't_ be supported in current PR.
There were 2 fixes needed:
1) Get rid of thread pool (it's unused anyway)
2) Fix optimizations rule generation. There were possible dangling references so I've changed `const auto&` on `const auto` because there are only pointer wrappers on the right side of assignment.


astrophysik (approved):
LGTM

PR #1225, opened 2025-01-23
add few builtins in runtime-light

astrophysik:


runtime-light/stdlib/server/http-functions.cpp
```diff
@@ -181,3 +181,16 @@ void f$setrawcookie(const string &name, const string &value, int64_t expire_or_o
   }
   kphp::http::header({static_SB_spare.c_str(), static_SB_spare.size()}, false, kphp::http::status::NO_STATUS);
 }
+
+array<string> f$headers_list() noexcept {
```
apolyakov, line 185: Maybe move it's implementation into the header?

runtime-light/stdlib/server/http-functions.cpp
```diff
@@ -181,3 +181,16 @@ void f$setrawcookie(const string &name, const string &value, int64_t expire_or_o
...
+  auto &http_server_instance_st{HttpServerInstanceState::get()};
```
apolyakov, line 186: Just a thought: isn't it better to reduce number of variables in functions? Here, for example, we can get rid of `http_server_instance_st` as it's used only once

runtime-light/stdlib/server/http-functions.cpp
```diff
@@ -181,3 +181,16 @@ void f$setrawcookie(const string &name, const string &value, int64_t expire_or_o
...
+  const auto &headers{http_server_instance_st.headers()};
+
+  array<string> list;
```
apolyakov, line 189: Let's add call to `reserve`

runtime-light/stdlib/server/http-functions.cpp
```diff
@@ -181,3 +181,16 @@ void f$setrawcookie(const string &name, const string &value, int64_t expire_or_o
...
+  for (const auto & header : headers) {
+    string value{(header.first + ": " + header.second).c_str()};
```
apolyakov, line 191: No need to temporary `std::string` here. It's better to use a combination of `reserve_at_least` and `append`

apolyakov (changes requested):

runtime-light/stdlib/server/http-functions.cpp
astrophysik: fix

runtime-light/stdlib/memory/memory-usage.h
```diff
@@ -0,0 +1,19 @@
+// Compiler for PHP (aka KPHP)
+// Copyright (c) 2025 LLC «V Kontakte»
+// Distributed under the GPL v3 License, see LICENSE.notice.txt
+
+#pragma once
+
+#include "runtime-common/core/allocator/runtime-allocator.h"
```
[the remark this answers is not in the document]
astrophysik: fix

runtime-light/stdlib/serialization/json-functions.h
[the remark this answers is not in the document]
astrophysik: ok

runtime-light/stdlib/server/http-functions.cpp
astrophysik: move

runtime-light/stdlib/server/http-functions.cpp
astrophysik: fix

apolyakov (approved):

apolyakov (approved):

apolyakov (approved):

PR #1229, opened 2025-01-31
do not initialize StringLibContext's static buffers

apolyakov:
This pull request enhances the performance of the HelloWorld benchmark for K2 by a factor of 32. Previously, we used zero-initialized `std::array` as a buffer in `StringLibContext`. This zeroing of the 8MB array at the start of each request consumed significant CPU time. By switching to uninitialized `std::array`, we avoid unnecessary initialization, resulting in substantial performance improvements.

Following benchmark is enough to see the difference:
```
?php

echo "Hello, World!\n";
```

### DIFF
--- a/runtime-common/stdlib/string/string-context.h
+++ b/runtime-common/stdlib/string/string-context.h
@@ -43,7 +43,7 @@ struct StringLibContext final : private vk::not_copyable {
   // The buffer is intended to be used as raw storage, and its contents will be
   // explicitly managed elsewhere in the code.
   std::array<char, STATIC_BUFFER_LENGTH + 1> static_buf;
-  std::array<char, MASK_BUFFER_LENGTH> mask_buffer;
+  std::array<char, MASK_BUFFER_LENGTH> mask_buf;
 
   StringLibContext() noexcept = default;
 
--- a/runtime-common/stdlib/string/string-context.h
+++ b/runtime-common/stdlib/string/string-context.h
@@ -30,17 +30,22 @@ inline constexpr std::string_view PERCENT_ = "%";
 
 }; // namespace string_context_impl_
 
-class StringLibContext final : vk::not_copyable {
+struct StringLibContext final : private vk::not_copyable {
   static constexpr int32_t MASK_BUFFER_LENGTH = 256;
-
-public:
   static constexpr int32_t STATIC_BUFFER_LENGTH = 1U << 23U;
-
-  std::array<char, STATIC_BUFFER_LENGTH + 1> static_buf{};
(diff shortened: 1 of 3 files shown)


runtime-common/stdlib/string/string-context.h
```diff
@@ -30,13 +32,16 @@ inline constexpr std::string_view PERCENT_ = "%";
...
-class StringLibContext final : vk::not_copyable {
+struct StringLibContext final : vk::not_copyable {
+  static constexpr int32_t STATIC_BUFFER_LENGTH = 1U << 23U;
+
+private:
   static constexpr int32_t MASK_BUFFER_LENGTH = 256;
 
-public:
-  static constexpr int32_t STATIC_BUFFER_LENGTH = 1U << 23U;
+  std::aligned_storage<STATIC_BUFFER_LENGTH + 1> storage;
```
mkornaukhov03, line 41: Are you sure that non-zeroing this memory won't lead to unexpected errors?
Did you test it with vkcom? This change seems hazardous
We have this kind of errors previously, see https://github.com/VKCOM/kphp/pull/1011/files#diff-a72cbbf809305fc21fdf70974cef5d9c5263b5ccfcb3d471fad561e03d323a0eR29

runtime-common/stdlib/string/string-context.h
```diff
@@ -30,13 +32,16 @@ inline constexpr std::string_view PERCENT_ = "%";
...
 
-  std::array<char, STATIC_BUFFER_LENGTH + 1> static_buf{};
+public:
+  char *static_buf{reinterpret_cast<char *>(std::addressof(storage))};
```
mkornaukhov03, line 44: I think it's better to have a method `static_buf()`.
We would get rid of extra invariant for self-referential data structure so there will be a narrower space for bugs.

runtime-common/stdlib/string/string-context.h
mkornaukhov03: By the way, we can align storage by size of cache line and it may lead benefits for small (up to 64 bytes) strings. I think it's worth to do, but up to you

mkornaukhov03 (changes requested):

runtime-common/stdlib/string/string-context.h
apolyakov: Let's do it in a separate PR

mkornaukhov03 (changes requested):
OK, but let's run better internal tests. I'll give an approve after passing them.

mkornaukhov03 (approved):
OK

PR #1237, opened 2025-02-17
Write `CompilerCore` hash tables stats and tune sizes

mkornaukhov03:


compiler/stats.cpp
```diff
@@ -90,6 +90,27 @@ void Stats::write_to(std::ostream &out, bool with_indent) const {
   out << indent << "compilation.total_time: " << total_time << std::endl;
   out << indent << "compilation.object_out_size: " << object_out_size << std::endl;
   out << block_sep;
+  out << indent << "hash_tables.max_files: " << max_files << std::endl;
+  out << indent << "hash_tables.total_files: " << total_files << std::endl;
+  out << indent << "hash_tables.max_dirs: " << max_dirs << std::endl;
+  out << indent << "hash_tables.total_dirs: " << total_dirs << std::endl;
+  out << indent << "hash_tables.max_functions: " << max_functions << std::endl;
+  out << indent << "hash_tables.total_functions: " << total_functions_ << std::endl;
```
Danil42Russia, line 98: why duplicate if there is already `functions.total`?

mkornaukhov03: I think it would be convenient to draw a aboard with all metrics that starts with `hash_tables`. It would be ugly if all except one will start with `hash_tables` and one will not.

Danil42Russia: ok I agree.

Danil42Russia (approved):

DrDet (approved):

PR #1238, opened 2025-02-19
fix msgpack allocations

astrophysik:


apolyakov (changes requested):

apolyakov (approved):

runtime/php_assert.cpp
```diff
@@ -93,7 +93,7 @@ static void print_demangled_adresses(void **buffer, int nptrs, int num_shift, bo
 }
 
 static void php_warning_impl(bool out_of_memory, int error_type, char const *message, va_list args) {
-  if (!out_of_memory && (php_warning_level == 0 || RuntimeContext::get().php_disable_warnings)) {
```
Danil42Russia: and are we sure we don't want to leave that logic out?
Otherwise, there will be OOM in the code with error suppression again, and we will waste time looking for it again.

Danil42Russia: but if you think it is not necessary, you can close conversation

astrophysik: I was thinking about this part. Maybe it's better to add here condition like `if error_type != E_ERROR`, but it could give large count of hided logs, that we're not ready for. I Suggest to make some experiments with that code and make final decision in the future

PR #1240, opened 2025-02-20
use murmur hash for class name hashing

apolyakov:


common/algorithms/hashes.cpp
```diff
@@ -0,0 +1,70 @@
...
+// Distributed under the GPL v3 License, see LICENSE.notice.txt
+
+#include "common/algorithms/hashes.h"
+
+#include <cstddef>
+#include <cstdint>
+
+namespace {
+
+size_t unaligned_load(const char *p) {
```
astrophysik, line 12: noexcept?

common/algorithms/hashes.cpp
```diff
@@ -0,0 +1,70 @@
...
+// MurMur hash function was taken from libstdc++
+template<>
+uint64_t murmur_hash<uint64_t>(const void *ptr, size_t len, size_t seed) noexcept {
+  static constexpr size_t mul = (static_cast<size_t>(0xc6a4a793UL) << 32UL) + static_cast<size_t>(0x5bd1e995UL);
+  const char *const buf = static_cast<const char *>(ptr);
+  // Remove the bytes not divisible by the sizeof(size_t).  This
+  // allows the main loop to process the data as 64-bit integers.
+  const int len_aligned = len & ~0x7;
+  const char *const end = buf + len_aligned;
+  size_t hash = seed ^ (len * mul);
```
astrophysik, line 45: Maybe make type of hash `uint64_t` as it declared in function signature?

common/algorithms/hashes.h
```diff
@@ -6,17 +6,20 @@
...
-#include <numeric>
 #include <utility>
 
 #include "common/type_traits/range_value_type.h"
 #include "common/wrappers/span.h"
 
 namespace vk {
 
+template<typename T>
+T murmur_hash(const void *ptr, size_t len, size_t seed = static_cast<size_t>(0xc70f6907UL)) noexcept;
```
astrophysik, line 21: As I saw in code there's a pattern
 ```C++
 std::string_view name{...};
 vk::murmur_hash(name.data(), name.size());
 ```
Maybe do signature murmur_hash like  `T murmur_hash(std::string_view`,...)?

runtime/ffi.h
```diff
@@ -15,7 +17,7 @@ struct C$FFI$CData: public refcountable_php_classes<C$FFI$CData<T>>, private Dum
   T c_value;
 
   const char *get_class() const noexcept { return "FFI\\CData"; }
-  int get_hash() const noexcept { return 1945543994; }
+  int get_hash() const noexcept { return static_cast<int32_t>(-1934403570); }
```
astrophysik, line 20: Constant is changed. Is it safety?

common/algorithms/hashes.h
apolyakov: I don't want to require `std::string_view` in order to use the hashing algorithm

runtime/ffi.h
apolyakov: Yes, it has changed since the hashing algorithm was updated

common/algorithms/hashes.h
```diff
@@ -47,9 +50,8 @@ size_t hash_range(const Rng &range, Hasher hasher = Hasher()) {
 template<class Rng, class Hasher = std::hash<range_value_type<Rng>>>
 class range_hasher : Hasher {
 public:
-  explicit range_hasher(Hasher hasher = Hasher()) :
-    Hasher(std::move(hasher)) {
-  }
+  explicit range_hasher(Hasher hasher = Hasher())
+    : Hasher(std::move(hasher)) {}
```
mkornaukhov03, line 54: Extra diff here and in other places in this file

runtime-light/stdlib/job-worker/job-worker.h
```diff
@@ -83,10 +85,10 @@ inline class_instance<C$KphpJobWorkerResponseError> f$KphpJobWorkerResponseError
   return v$this;
 }
 
-inline string f$KphpJobWorkerResponseError$$getError(class_instance<C$KphpJobWorkerResponseError> v$this) noexcept {
+inline string f$KphpJobWorkerResponseError$$getError(const class_instance<C$KphpJobWorkerResponseError> &v$this) noexcept {
```
mkornaukhov03, line 88: Why do we need it?

common/algorithms/murmur-hash.cpp
```diff
@@ -0,0 +1,83 @@
...
+  return v ^ (v >> 47);
+}
+
+} // namespace
+
+namespace vk {
+
+// MurMur hash function was taken from libstdc++
+template<>
+uint64_t murmur_hash<uint64_t>(const void *ptr, size_t len, size_t seed) noexcept {
```
mkornaukhov03, line 51: Can't it be defined as `constexpr`?

runtime/datetime/datetime.h
```diff
@@ -21,7 +25,8 @@ struct C$DateTime : public refcountable_polymorphic_php_classes<C$DateTimeInterf
   }
 
   int get_hash() const noexcept final {
-    return 2141635158;
+    std::string_view name_view{get_class()};
+    return static_cast<int32_t>(vk::murmur_hash<uint32_t>(name_view.data(), name_view.size()));
```
mkornaukhov03, line 29: Can't `murmurhash` be defined as constexpr?

mkornaukhov03 (changes requested):

builtin-functions/kphp-full/_functions.txt
```diff
@@ -1584,13 +1584,13 @@ class JsonEncoder {
   static function from_json_impl(string $encoder_tag, string $json, string $class_name) ::: instance<^3>;
 }
 
-class DateTimeZone {
+final class DateTimeZone {
```
Danil42Russia, line 1587: why? in php these are not `final` classes. We support inheritance from classes in the standard library

common/common.cmake
```diff
@@ -9,6 +9,7 @@ prepend(COMMON_MAIN_SOURCES ${COMMON_DIR}/
 
         sha1.cpp
         allocators/freelist.cpp
+	algorithms/murmur-hash.cpp
```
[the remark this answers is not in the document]
apolyakov: It looks nicely indented in both github editor and my IDE

runtime-light/stdlib/job-worker/job-worker.h
apolyakov: Just to get rid of a warning

common/algorithms/murmur-hash.cpp
apolyakov: It'd force us to define it in header. Anyway, there is not much point in doing this as most of `get_hash` methods are `virtual` which means we can't make them `constexpr`

builtin-functions/kphp-full/_functions.txt
apolyakov: Fixed. Kept them non-`final`. Also, fixed the declarations of such classes so they can really be extended in user code

common/algorithms/hashes.h
apolyakov: Fixed

common/common.cmake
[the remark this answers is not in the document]
mkornaukhov03: It's tab, we use spaces instead

compiler/compiler.cmake
```diff
@@ -15,6 +15,7 @@ endif()
 
 prepend(KPHP_COMPILER_COMMON ${COMMON_DIR}/
         dl-utils-lite.cpp
+	algorithms/murmur-hash.cpp
```
mkornaukhov03, line 18: Same here

mkornaukhov03 (approved):
OK

PR #1242, opened 2025-02-24
[k2] do not set matches in regex functions unless it's necessary

apolyakov:
This pull request enhances the handling of the `matches` parameter by making it a `Optional`. This change improves the flexibility and safety of the code by explicitly representing the presence or absence of a `matches` parameter to fill.

Key changes:

- The matches variable is now a `Optional`, allowing it to explicitly indicate whether a match is present.
- The use of RegexInstanceState has been eliminated. Previously, it was used to default-initialize certain parameters with lvalue references. This refactoring enhances the overall performance of regex functions since the call to `RegexInstanceState::get()` leads to accessing a thread-local variable inside k2 platform.

### DIFF
--- a/runtime-light/stdlib/string/regex-functions.cpp
+++ b/runtime-light/stdlib/string/regex-functions.cpp
@@ -511,8 +511,8 @@ Optional<int64_t> f$preg_match(const string &pattern, const string &subject, Opt
   std::optional<std::reference_wrapper<mixed>> matches{};
   if (opt_matches.has_value()) {
     php_assert(std::holds_alternative<std::reference_wrapper<mixed>>(opt_matches.val()));
-    auto inner_ref{std::get<std::reference_wrapper<mixed>>(opt_matches.val())};
-    inner_ref.get() = array<mixed>{};
+    auto &inner_ref{std::get<std::reference_wrapper<mixed>>(opt_matches.val()).get()};
+    inner_ref = array<mixed>{};
     matches.emplace(inner_ref);
   }
   set_matches(regex_info, flags, matches, trailing_unmatch::skip);
--- a/runtime-light/stdlib/string/regex-functions.cpp
+++ b/runtime-light/stdlib/string/regex-functions.cpp
@@ -16,6 +16,7 @@
 #include <optional>
 #include <string_view>
 #include <type_traits>
+#include <variant>
 
 #include "common/containers/final_action.h"
 #include "runtime-common/core/allocator/script-allocator.h"
@@ -493,8 +494,8 @@ bool replace_regex(RegexInfo &regex_info, uint64_t limit) noexcept {
 
 } // namespace
 
-Optional<int64_t> f$preg_match(const string &pattern, const string &subject, mixed &matches, int64_t flags, int64_t offset) noexcept {
-  matches = array<mixed>{};
+Optional<int64_t> f$preg_match(const string &pattern, const string &subject, Optional<std::variant<std::monostate, std::reference_wrapper<mixed>>> opt_matches,
+                               int64_t flags, int64_t offset) noexcept {
   RegexInfo regex_info{{pattern.c_str(), pattern.size()}, {subject.c_str(), subject.size()}, {}};
 
   bool success{valid_regex_flags(flags, kphp::regex::PREG_NO_FLAGS, kphp::regex::PREG_OFFSET_CAPTURE, kphp::regex::PREG_UNMATCHED_AS_NULL)};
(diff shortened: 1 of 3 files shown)


runtime-light/stdlib/string/regex-functions.cpp
```diff
@@ -320,17 +322,24 @@ bool match_regex(RegexInfo &regex_info, size_t offset) noexcept {
...
   // get the ouput vector from the match data
   const auto *ovector{pcre2_get_ovector_pointer_8(regex_state.regex_pcre2_match_data.get())};
+  const auto end_offset{ovector[1]};
+  // early return in case we don't need to actually set matches
+  if (!opt_matches.has_value()) {
+    return end_offset;
+  }
+
+  const auto offset_capture{static_cast<bool>(flags & kphp::regex::PREG_OFFSET_CAPTURE)};
+  const auto unmatched_as_null{static_cast<bool>(flags & kphp::regex::PREG_UNMATCHED_AS_NULL)};
```
mkornaukhov03, line 342: Why cast to `bool` and not explicit comparison to 0?

runtime-light/stdlib/string/regex-functions.cpp
```diff
@@ -374,22 +383,28 @@ PCRE2_SIZE set_matches(const RegexInfo &regex_info, int64_t flags, mixed &matche
...
+PCRE2_SIZE set_all_matches(const RegexInfo &regex_info, int64_t flags, std::optional<std::reference_wrapper<mixed>> opt_all_matches) noexcept {
   const auto pattern_order{!static_cast<bool>(flags & kphp::regex::PREG_SET_ORDER)};
 
+  // early return in case we don't actually need to set matches
+  if (!opt_all_matches.has_value()) {
+    return set_matches(regex_info, flags, std::nullopt, pattern_order ? trailing_unmatch::include : trailing_unmatch::skip);
+  }
+
   mixed matches;
   PCRE2_SIZE offset{set_matches(regex_info, flags, matches, pattern_order ? trailing_unmatch::include : trailing_unmatch::skip)};
```
mkornaukhov03, line 402:
```C++
PCRE2_SIZE offset{set_matches(regex_info, flags, matches, pattern_order ? trailing_unmatch::include : trailing_unmatch::skip)};
if (!opt_all_matches.has_value()) { return offset; }
mixed matches;
...
```

runtime-light/stdlib/string/regex-functions.cpp
```diff
@@ -543,8 +574,17 @@ Optional<int64_t> f$preg_match_all(const string &pattern, const string &subject,
   return entire_match_count;
 }
 
-Optional<string> f$preg_replace(const string &pattern, const string &replacement, const string &subject, int64_t limit, int64_t &count) noexcept {
-  count = 0;
+Optional<string> f$preg_replace(const string &pattern, const string &replacement, const string &subject, int64_t limit,
+                                Optional<std::variant<std::monostate, std::reference_wrapper<int64_t>>> opt_count) noexcept {
```
mkornaukhov03, line 578: NIT: alias for `Optional<std::variant<std::monostate, std::reference_wrapper<T>>>`. It looks awfully large

runtime-light/stdlib/string/regex-functions.cpp
```diff
@@ -604,8 +653,17 @@ Optional<string> f$preg_replace(const mixed &pattern, const string &replacement,
...
 
-Optional<string> f$preg_replace(const mixed &pattern, const mixed &replacement, const string &subject, int64_t limit, int64_t &count) noexcept {
-  count = 0;
+Optional<string> f$preg_replace(const mixed &pattern, const mixed &replacement, const string &subject, int64_t limit,
+                                Optional<std::variant<std::monostate, std::reference_wrapper<int64_t>>> opt_count) noexcept {
+  int64_t count{};
+  vk::final_action count_finalizer{[&count, &opt_count]() noexcept {
+    if (opt_count.has_value()) {
+      php_assert(std::holds_alternative<std::reference_wrapper<int64_t>>(opt_count.val()));
+      auto &inner_ref{std::get<std::reference_wrapper<int64_t>>(opt_count.val()).get()};
```
mkornaukhov03, line 662: NIT: And better syntax for `holds_alternative` and `get`?

mkornaukhov03 (changes requested):

runtime-light/stdlib/string/regex-functions.cpp
apolyakov: It'd break the semantics of the function

runtime-light/stdlib/string/regex-functions.cpp
apolyakov: Yeah, it looks awful. But I think we should not have an alias for that. Having such alias would hide the complexity of this type, e.g. the variant semantics under the hood. Ideally, I'd create a wrapper type with its own API since such a pattern is gonna be used many time across runtime. But let's have it as a separate task

runtime-light/stdlib/string/regex-functions.cpp
apolyakov: I didn't understand your comment. Can you explain it a bit?

runtime-light/stdlib/string/regex-functions.cpp
mkornaukhov03: I don't see, give an example

runtime-light/stdlib/string/regex-functions.cpp
mkornaukhov03: Referring to the previous conversation about type alias or wrapper

mkornaukhov03 (approved):
OK

PR #1243, opened 2025-02-27
[k2] implement array_reduce

apolyakov:
This PR:

- add `array_reduce` implementation for K2 runtime
- refined the way return types are declared for implicit lambdas used in asynchronous built-in callbacks


mkornaukhov03 (approved):
OK

PR #1247, opened 2025-03-03
[k2] fix async builtin lambda's return type

apolyakov:


mkornaukhov03 (changes requested):
Please, add a test or an example

mkornaukhov03 (approved):
LGTM

PR #1249, opened 2025-03-04
added `hostname` output to json logs

Danil42Russia:


DrDet (changes requested):
Actually `host_name` is written as one word, i.e. `hostname`, let's change it everywhere

server/json-logger.cpp
```diff
@@ -325,6 +326,7 @@ void JsonLogger::reset_buffers() noexcept {
 
 void JsonLogger::write_general_info(JsonBuffer *json_out_it, int type, int64_t created_at, bool uncaught) {
   json_out_it->append_key("version").append_integer(release_version_);
+  json_out_it->append_key("host_name").append_string(host_name_);
```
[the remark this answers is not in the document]
Danil42Russia: Why? This is precisely the most important field for searching and grouping

server/json-logger.h
```diff
@@ -65,6 +65,7 @@ class JsonLogger : vk::not_copyable {
   volatile sig_atomic_t json_traces_count{0};
 
   int64_t release_version_{0};
+  std::string_view hostname_{""};
```
DrDet, line 68: do you mean this?
```suggestion
  vk::string_view hostname_{""};
```

server/json-logger.cpp
```diff
@@ -185,8 +185,9 @@ JsonLogger::JsonBuffer &JsonLogger::JsonBuffer::append_raw_string(vk::string_vie
   return *this;
 }
 
-void JsonLogger::init(int64_t release_version) noexcept {
+void JsonLogger::init(const int64_t release_version, const vk::string_view hostname) noexcept {
```
apolyakov, line 188: No need for `const` here, but up to you

server/json-logger.h
apolyakov, line 68: Please, use one of `std::string_view` or `vk::string_view`, but not both

apolyakov: Also, no need for explicit initialisation in your case `""`. `std::string_view hostname;` is enough

Danil42Russia: yes, you're right. I must have written `std::string_view` out of habit when I wrote the code here

server/json-logger.cpp
Danil42Russia: i believe it is good practice to use `const'

server/json-logger.h
Danil42Russia: fixed :)

DrDet (approved):

server/json-logger.h
```diff
@@ -17,7 +17,7 @@ class JsonLogger : vk::not_copyable {
 public:
   friend class vk::singleton<JsonLogger>;
 
-  void init(int64_t release_version) noexcept;
+  void init(int64_t release_version, vk::string_view hostname) noexcept;
```
astrophysik, line 20: Is it okay, that args don't have `const` qualifier, but in implementation they have?

astrophysik (changes requested):

PR #1250, opened 2025-03-07
add forbid stubs option in compiler

astrophysik:
General
-----------
This pr add option `--forbid-stubs` in kphp2cpp which doesn't allow to build code, if it use stubs

Since we can't generate body for every stubs and we have some 'manual implementation' tag `generate-body` was divided into two `stub` and `generation-required`

Also for easier understanding declarations in .txt files were refactored


compiler/kphp2cpp.cpp
```diff
@@ -221,6 +221,8 @@ int main(int argc, char *argv[]) {
              'l', "link-with", "KPHP_LINK_FILE");
   parser.add("Build runtime from sources", settings->force_link_runtime,
              "force-link-runtime", "KPHP_FORCE_LINK_RUNTIME");
+  parser.add("Enable safe compilation mode to forbid builtins stubs using", settings->forbid_stubs_using,
```
apolyakov, line 224: `Enable safe compilation mode that treats calls to stub functions as compilation error`

builtin-functions/kphp-light/stdlib/component-functions.txt
```diff
@@ -0,0 +1,22 @@
+<?php
+
+// ===== SUPPORTED =====
```
mkornaukhov03, line 3: NIT: remove

builtin-functions/kphp-light/stdlib/confdata-functions.txt
```diff
@@ -1,4 +1,6 @@
-<?
+<?php
+
+// ===== SUPPORTED =====
```
mkornaukhov03, line 3: What for?
Isn't it supported by default, unless the opposite is written?

mkornaukhov03 (changes requested):
Why do we need to separate this phpdoc into two?
Are there any case when we need `stub` only and do not need `generation-required`?

mkornaukhov03 (approved):
LGTM

builtin-functions/kphp-light/stdlib/component-functions.txt
astrophysik: fix

builtin-functions/kphp-light/stdlib/confdata-functions.txt
astrophysik: fix

mkornaukhov03 (approved):
OK

PR #1255, opened 2025-03-11
dependencies vendoring, part 3

PetrShumilov:
Plan for libs:

- [x] Uber H3
- [x] KPHP Timelib
- [x] YAML cpp
- [x] Numa


astrophysik:
It is great to see just one static lib `libvk-flex-data.a` in linker flags. Good job!

.gitmodules
```diff
@@ -27,3 +27,17 @@
 	path = third-party/pcre
 	url = https://github.com/VKCOM/PCRE.git
 	branch = debian/buster
+[submodule "third-party/uber-h3"]
+	path = third-party/uber-h3
+	url = https://github.com/VKCOM/uber-h3.git
```
Danil42Russia, line 32: are we sure we need our own fork? We don't have our own patches.

PetrShumilov: @Danil42Russia  You mean we should use upstream `https://github.com/uber/h3` as source instead of `https://github.com/VKCOM/uber-h3`, am I right? If yes, it doesn't conduct our philosophy -- as much as possible third-parties libraries must be located at controlled by us repositories.

Our future goal -- increase the control for each dependency

apolyakov (approved):
LGTM

promo4ka (approved):

PR #1261, opened 2025-03-19
added datacenter name to statistics

Danil42Russia:


server/statshouse/statshouse-manager.cpp
```diff
@@ -196,7 +196,8 @@ void StatsHouseManager::add_common_master_stats(const workers_stats_t &workers_s
     client.metric("kphp_by_host_version", true).tag(std::to_string(engine_tag_number)).write_count(1);
   }
 
-  client.metric("kphp_uptime").write_value(get_uptime());
+  static const auto dc_name = vk::singleton<ServerConfig>::get().get_dc_name();
+  client.metric("kphp_uptime").tag(dc_name).write_value(get_uptime());
```
DrDet, line 200: I think we want to have this tag for all our metric. Check [this](https://github.com/VKCOM/kphp/blob/bf6cd9a7d16e15c31a0b69b534c887305f935d7b/server/statshouse/statshouse-client.cpp#L11) out

Danil42Russia: changed

server/statshouse/statshouse-client.cpp
```diff
@@ -7,6 +7,7 @@
 statshouse::TransportUDPBase::MetricBuilder StatsHouseClient::metric(std::string_view name, bool force_tag_host) {
   auto builder = transport.metric(name);
   builder.tag(tag_cluster);
+  builder.tag(tag_dc_name);
```
DrDet, line 10: Unnamed tag like this will break all current dashboards, because all other unnamed tags will shift. We need to use named tag here

DrDet (changes requested):

server/statshouse/statshouse-client.cpp
Danil42Russia: omg, you're right as always. I thought of that and promptly forgot about it. Thanks for noticing that.

DrDet (approved):

PR #1262, opened 2025-03-19
[k2] add php_sapi_name 

astrophysik:


runtime-light/state/instance-state.cpp
```diff
@@ -83,8 +84,9 @@ task_t<void> InstanceState::run_instance_prologue() noexcept {
     using namespace PhpServerSuperGlobalIndices;
     superglobals.v$_SERVER.set_value(string{REQUEST_TIME.data(), REQUEST_TIME.size()}, static_cast<int64_t>(time_mcs));
     superglobals.v$_SERVER.set_value(string{REQUEST_TIME_FLOAT.data(), REQUEST_TIME_FLOAT.size()}, static_cast<double>(time_mcs));
+    superglobals.v$d$PHP_SAPI = system_functions_impl_::php_sapi_name(image_kind_);
```
apolyakov, line 87: let's better implement this as a template function (`template <ImageKind kind>`) here. Thus we remove the number of includes and therefore the number of inner dependencies

runtime-light/state/instance-state.cpp
```diff
@@ -83,8 +84,9 @@ task_t<void> InstanceState::run_instance_prologue() noexcept {
...
   }
-  // TODO sapi, env
```
apolyakov, line 87: please, remove `env` as well. it's already done

runtime-light/stdlib/system/system-functions.h
```diff
@@ -60,6 +79,8 @@ inline string f$php_uname(const string &mode = string{1, 'a'}) noexcept {
   }
 }
 
+string f$php_sapi_name() noexcept;
```
apolyakov, line 82: if you address previous comments, there will be no need in moving this implementation into the `cpp` file.

apolyakov (changes requested):

apolyakov (approved):
LGTM

PR #1272, opened 2025-03-28
propagate custom constructor of base class through abstract class in inheritance chain

PetrShumilov:
Following code:
```
class A {
    public function __construct() {
        echo "A!";
    }
}

abstract class B extends A {}

class C extends B {}

new C();
```
should lead to output like this:
```
A1
```

Currently, we get empty output due incorrect code gen. Constructor of class `C` looks like: 
```
#pragma once
#include "runtime-headers.h"
#include "cl/C@C.h"

inline class_instance<C$C> f$C$$__construct(class_instance<C$C> const &v$this) noexcept ;

//source = [test.php]
//14: class C extends B {
class_instance<C$C> f$C$$__construct(class_instance<C$C> const &v$this) noexcept  {
  return v$this;
} 
```
Constructor for `A` has not been generated


mkornaukhov03 (approved):
LGTM

ZemlyakovDmitry (approved):

ZemlyakovDmitry (approved):

mkornaukhov03 (approved):
OK

PR #1273, opened 2025-04-01
[k2] coroutines refactoring

apolyakov:
Main changes in the PR:

- `task_t`
    - rename to `task`
    - move to `kphp::coro` namespace
    - get rid of `std::exception`
    - refactor to make its implementation similar to `shared_task_t`'s one
- `shared_task_t`
    - rename to `shared_task`
    - move to `kphp::coro` namespace
    - replace linked list of awaiters with doubly-linked list
    - reimplement `cancel_awaiter` to cancel an awaiter in `O(1)` instead of `O(N)`


runtime-light/coroutine/shared-task.h
```diff
@@ -16,32 +16,32 @@
...
-
-      constexpr bool await_ready() const noexcept {
+    struct awaiter {
+      constexpr auto await_ready() const noexcept -> bool {
         return false;
       }
 
-      constexpr std::coroutine_handle<> await_suspend(std::coroutine_handle<promise_type> coro) const noexcept {
-        promise_base_t &promise{coro.promise()};
+      auto await_suspend(std::coroutine_handle<promise_type> coro) const noexcept -> std::coroutine_handle<> {
```
mkornaukhov03, line 41: Why constexpr has been moved?

runtime-light/coroutine/shared-task.h
```diff
@@ -57,20 +57,20 @@ struct promise_base_t {
...
     php_critical_error("internal unhandled exception");
   }
 
-  constexpr bool ready() const noexcept {
+  auto done() const noexcept -> bool {
     return m_waiters == this;
   }
 
-  constexpr void add_ref() noexcept {
+  auto add_ref() noexcept -> void {
```
mkornaukhov03, line 73: Why do you use such a style `auto foo() -> Bar {...}`?
As I see, it isn't commonly used in repo

runtime-light/coroutine/shared-task.h
```diff
@@ -129,28 +129,22 @@ struct promise_base_t {
     const auto *waiter_ptr{std::addressof(waiter)};
     if (m_waiters == waiter_ptr) { // waiter is the head of the list
       m_waiters = waiter_ptr->m_next;
-      return;
+    } else if (waiter_ptr->m_next == nullptr) { // waiter is the last in the list
+      waiter_ptr->m_prev->m_next = nullptr;
+    } else { // waiter is somewhere in the middle of the list
+      waiter_ptr->m_next->m_prev = waiter_ptr->m_prev;
+      waiter_ptr->m_prev->m_next = waiter_ptr->m_next;
```
mkornaukhov03, line 136: Can't `waiter_ptr->m_prev` be nullptr here?

mkornaukhov03 (changes requested):

runtime-light/coroutine/shared-task.h
apolyakov: I decided that it's better ti give less guarantees for several functions in `shared-task.h` and `task.h`. They are actually never `constexpr`, but `constexpr` adds some limitations to the ways these functions can be used.

runtime-light/coroutine/shared-task.h
apolyakov: I decided to use it here to increase readability. It's not so important to first see the return type of coroutine traits functions. In opposite it's crucial to clearly see the structure behind these functions, e.g. clearly separate awaiters and promises, initial and final suspenders etc

runtime-light/coroutine/shared-task.h
apolyakov: No, it can't since the waiter here is somewhere in the middle of the waiters list, so it has both prev and next

mkornaukhov03 (approved):
LGTM

PR #1276, opened 2025-04-09
[k2] optimize run_instance_prologue by changing ComponentState::env's type

apolyakov:
Changing `ComponentState::env`'s type from `array<string>` to `array<mixed>` eliminates call to `array<T>::copy_from` on assignment


mkornaukhov03 (approved):
LGTM

PR #1285, opened 2025-04-13
Danil42russia/logs refactoring

Danil42Russia:


DrDet (commented):
It seems, that we already have the criteria to distinguish critical 'runtime' and 'server' errors from warnigns -- and it's `uncaught == true`
https://github.com/VKCOM/kphp/blob/6c0468825aa0afedbafec230fe2ed6ebd6356709/server/json-logger.cpp#L352

DrDet (dismissed):

astrophysik (approved):

PR #1289, opened 2025-04-17
fix incorrect reference to OpenSSL headers

PetrShumilov:
Fix issues like this:
```
make[2]: *** No rule to make target '../objs/include/openssl/aes.h', needed by 'CMakeFiles/binlog-src-no-pic.dir/common/binlog/kdb-binlog-common.cpp.o'.  Stop.
  make[1]: *** [CMakeFiles/Makefile2:591: CMakeFiles/binlog-src-no-pic.dir/all] Error 2
  make[1]: *** Waiting for unfinished jobs..
```


mkornaukhov03 (approved):
OK

PR #1292, opened 2025-04-18
add support for k2-server in python test

astrophysik:
Description
---------------
This pr add support for K2 in python server tests. 

Interface
------------
> How to run tests in new mode?

To run tests in K2 mode, should be setted `K2_BIN=/path/to/k2-node` env variable. By default, all tests are considered to be running unless they are manually switched off.

> How to disable test in K2 mode?

To disable test there are two options:
1) To skip one or more tests in a test suite, but run the remaining tests, there is the attribute `@pytest.mark.k2_skip`

Example:
```python
class TestSuite:
   def test_one():
      pass
      
   @pytest.mark.k2_skip
   def test_two():
      pass
``` 
In this example `test_two()` will not be executed in new mode, but it will work fine with old kphp server

2) Second option to skip tests is the attribute for test suite class `@pytest.mark.k2_skip_suite`. If a class has this tag no one test will not be launched. Moreover, functions of test's prologue and epilogue will not be performed. It's important when php code contains a lot of unimplemented logic

Example:
```python
@pytest.mark.k2_skip_suite
class TestSuite:
   def test_one():
      pass
``` 

> How to configure tests for K2?

K2Server tests support same configuration style as for KPHPServer with function `extra_class_setup()` and `extra_class_teardown()`. To determine type of web server inside test there is a method `should_use_k2()`. Also note that linking file of component should be stored in `/data/component-config.yaml` (look at the existing tests)

 Implementation
 --------------------
 
Now the `WebServerAutoTestCase` type stores web_server which is defined when setting up the class. `K2Server` and `KphpServer` are implementation of `WebServer`


tests/kphp_ci_pipeline_runner.py
```diff
@@ -413,6 +413,23 @@ def parse_args():
         skip=args.steps and "functional-tests" not in args.steps,
     )
 
+    runner.add_test_group(
+        name="k2-functional-tests",
+        description="run k2-kphp functional tests with cxx={}".format("gcc"),
```
apolyakov, line 418: It should not be just `gcc`. It actually may vary

tests/python/lib/k2_server.py
```diff
@@ -0,0 +1,47 @@
...
+import time
+
+from .web_server import WebServer
+
+
+class K2Server(WebServer):
+    def __init__(self, k2_server_bin, working_dir, kphp_build_dir, options=None, auto_start=False):
+        """
+        :param k2_server_bin: Path to k2 binary
+        :param working_dir: Working dir where k2 will be launch
```
apolyakov, line 11: `lauched`

apolyakov (dismissed):

tests/python/lib/testcase.py
```diff
@@ -255,12 +275,22 @@ def extra_class_teardown(cls):
...
         return nocc_env("NOCC_SERVERS_FILENAME", None) is not None
 
+    @classmethod
+    def should_use_k2(cls):
+        return search_k2_bin() is not None
+
+    @classmethod
+    def k2_server_env(cls):
+        env = {"KPHP_MODE": "k2-server", "KPHP_ENABLE_FULL_PERFORMANCE_ANALYZE": "0",
+               "KPHP_PROFILER": "0", "KPHP_USER_BINARY_PATH": "component.so", "KPHP_FORCE_LINK_RUNTIME": "1"}
```
mkornaukhov03, line 285: It isn't k2 server env, incorrect naming. It's KPHP specific build flags, I would call it smth like `kphp_env_for_k2_build()`.  Or I'd move this stuff to `kphp_builder.py` file.

tests/python/lib/testcase.py
```diff
@@ -313,18 +343,32 @@ def custom_teardown(cls):
...
 
+    @classmethod
+    def should_use_k2(cls):
+        return search_k2_bin() is not None
+
+    @classmethod
+    def k2_cli_env(cls):
+        env = {"KPHP_MODE": "k2-cli", "KPHP_ENABLE_FULL_PERFORMANCE_ANALYZE": "0",
+               "KPHP_PROFILER": "0", "KPHP_USER_BINARY_PATH": "component.so", "KPHP_FORCE_LINK_RUNTIME": "1"}
+        return env
```
mkornaukhov03, line 354: Same for this

tests/python/lib/testcase.py
mkornaukhov03: By the way, I haven't found real k2_server_env where I expected to see thinks like `RUST_LOG` or rpc connection pools

mkornaukhov03 (changes requested):

mkornaukhov03 (approved):
LGTM

PR #1293, opened 2025-04-18
Get rid of C++ exceptions in msgpack

mkornaukhov03:


runtime-common/core/core-context.h
```diff
@@ -4,10 +4,13 @@
...
 #pragma once
 
-#include <cstddef>
+#include <array>
 #include <cstdint>
+#include <cstring>
+#include <optional>
 
 #include "common/mixin/not_copyable.h"
+#include "common/wrappers/string_view.h"
```
astrophysik, line 13: Why this include is added here?

compiler/code-gen/declarations.cpp
```diff
@@ -1212,7 +1212,7 @@ void ClassMembersDefinition::compile_msgpack_deserialize(CodeGenerator &W, Class
   cases.emplace_back("default: break;");
 
   W << "void " << klass->src_name << "::msgpack_unpack(const vk::msgpack::object &msgpack_o) " << BEGIN
-    << "if (msgpack_o.type != vk::msgpack::stored_type::ARRAY) { throw vk::msgpack::type_error{}; }" << NL
+    << "if (msgpack_o.type != vk::msgpack::stored_type::ARRAY) { RuntimeContext::get().set_msgpack_error(\"Unknown type found during deserialization\"); return; }" << NL
```
astrophysik, line 1215: Can we here use value of constant `TYPE_ERROR`?

runtime-common/stdlib/msgpack/adaptors.h
```diff
@@ -28,16 +29,19 @@ struct convert_integer_sign<T, true> {
   static T convert(const msgpack::object& o) {
     if (o.type == stored_type::POSITIVE_INTEGER) {
       if (o.via.u64 > static_cast<uint64_t>(std::numeric_limits<T>::max())) {
-        throw type_error{};
+        RuntimeContext::get().set_msgpack_error(TYPE_ERROR);
+        return T{};
```
astrophysik, line 33: Can type T be non default constructible?

runtime-common/stdlib/msgpack/object_visitor.h
```diff
@@ -65,10 +67,19 @@ class object_visitor : vk::not_copyable {
   void parse_error(size_t /*parsed_offset*/, size_t /*error_offset*/) const;
   void insufficient_bytes(size_t /*parsed_offset*/, size_t /*error_offset*/) const;
 
+  bool has_error() const {
+    return error.has_value();
+  }
+
+  std::optional<vk::string_view> get_error() const {
```
astrophysik, line 74: Maybe `std::string_view`? It is available from c++17 and old runtime require this standart

runtime-common/stdlib/msgpack/packer.h
```diff
@@ -64,6 +64,7 @@ class packer : private vk::not_copyable {
   void append_buffer(const char* buf, size_t len) noexcept;
 
   Stream& stream_;
+  bool has_error = false;
```
astrophysik, line 67: Maybe add here {false}?) :smiley:. By the way `bool a{}` has same value as `bool a{false}`. Integrals types are initialised with null by default

runtime-common/stdlib/msgpack/unpacker.cpp
```diff
@@ -5,31 +5,47 @@
...
   }
 }
 
 bool unpacker::has_error() const noexcept {
-  return bytes_consumed_ != input_.size();
+  return RuntimeContext::get().msgpack_error.has_value() || visitor_error_.has_value() || zone_.has_error() || bytes_consumed_ != input_.size();
 }
 
 string unpacker::get_error_msg() const noexcept {
+  auto& rt_context = RuntimeContext::get();
```
astrophysik, line 37: As I see rt_context doesn't change here. Maybe add here `const`?

runtime-common/stdlib/msgpack/unpacker.h
```diff
@@ -25,6 +25,10 @@ class unpacker : private vk::not_copyable {
   const string& input_;
   std::size_t bytes_consumed_{0};
   msgpack::zone zone_;
+
+  // It's awful because unpacker and visitor are strongly logically connected,
+  // but there is no aggregation or composition. Should be fixed some day.
+  std::optional<vk::string_view> visitor_error_{std::nullopt};
```
astrophysik, line 31: Let's make initialization of type std::optional consistent with `runtime-common/core/core-context.h:27`.

runtime-common/core/core-context.h
mkornaukhov03: Previously there was string view instead of array. I'll remove it

runtime-common/stdlib/msgpack/object_visitor.h
mkornaukhov03: It doesn't have constructor from two pointers, that is why we have a wrapper for that and use it in both compiler and old runtime.

runtime-common/stdlib/msgpack/adaptors.h
mkornaukhov03: All runtime and codegen classes are default constructible

runtime-common/stdlib/msgpack/unpacker.cpp
```diff
@@ -5,31 +5,48 @@
...
   }
 }
 
 bool unpacker::has_error() const noexcept {
-  return bytes_consumed_ != input_.size();
+  return SerializationLibContext::get().msgpack_error.has_value() || visitor_error_.has_value() || zone_.has_error() || bytes_consumed_ != input_.size();
 }
 
 string unpacker::get_error_msg() const noexcept {
+  const auto& serialization_ctx = SerializationLibContext::get();
```
astrophysik, line 38: Just comment - here can be `serialization_ctx{SerializationLibContext::get()}`

runtime-common/stdlib/serialization/serialization-context.h
```diff
@@ -13,6 +16,16 @@ struct SerializationLibContext final : private vk::not_copyable {
   // msgpack serialization
   size_t instance_depth{0};
   uint32_t serialize_as_float32_{0};
+  std::optional<std::array<char, 256>> msgpack_error{std::nullopt};
```
astrophysik, line 19: Just comment - default constructor of std::optional will create an optional without value

astrophysik (approved):

runtime-common/stdlib/msgpack/object_visitor.h
```diff
@@ -65,10 +67,19 @@ class object_visitor : vk::not_copyable {
   void parse_error(size_t /*parsed_offset*/, size_t /*error_offset*/) const;
   void insufficient_bytes(size_t /*parsed_offset*/, size_t /*error_offset*/) const;
 
+  bool has_error() const {
```
apolyakov, line 70: Can we not add this method? It seems enough to just check if an optional has value in place

runtime-common/stdlib/msgpack/zone.h
```diff
@@ -27,12 +30,14 @@ class zone : private vk::not_copyable {
...
     char* m_ptr{nullptr};
     chunk* m_head{nullptr};
+    bool m_error{false};
   };
 
   char* allocate_expand(size_t size);
 
   size_t m_chunk_size{0};
   chunk_list m_chunk_list;
+  bool m_error{false};
```
apolyakov, line 40: It seems that `zone` only has error in case of bad alloc. Isn't it enough to just check for `nullptr`?

runtime-common/stdlib/serialization/serialization-context.h
apolyakov, line 19: Let's use constant instead of hardcoded value. Also, it's better to just use default init here `msgpack_error{}`

runtime-common/stdlib/serialization/serialization-context.h
```diff
@@ -13,6 +16,16 @@ struct SerializationLibContext final : private vk::not_copyable {
...
+
+  void set_msgpack_error(const char* msg) noexcept {
```
apolyakov, line 21: It's better to have `std::string_view` as interface here

runtime-common/stdlib/serialization/serialization-context.h
```diff
@@ -13,6 +16,16 @@ struct SerializationLibContext final : private vk::not_copyable {
...
+    msgpack_error = std::array<char, 256>();
+    strncpy(msgpack_error->begin(), msg, 256);
```
apolyakov, line 23: `std::strncpy`

runtime-common/stdlib/msgpack/object.h
```diff
@@ -68,7 +68,9 @@ struct object {
    */
   template<typename T>
   T as() const {
-    T v;
+    // Without initializationg we have maybe-uninitialized warning.
+    // Assume that default constructor is cheap.
+    T v{};
```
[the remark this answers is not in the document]
mkornaukhov03: Almost any case. For example, msgpack-test.cpp.

runtime-common/stdlib/msgpack/adaptors.h
```diff
@@ -28,16 +27,19 @@ struct convert_integer_sign<T, true> {
   static T convert(const msgpack::object& o) {
     if (o.type == stored_type::POSITIVE_INTEGER) {
       if (o.via.u64 > static_cast<uint64_t>(std::numeric_limits<T>::max())) {
-        throw type_error{};
+        SerializationLibContext::get().set_msgpack_error(TYPE_ERROR);
```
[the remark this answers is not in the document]
mkornaukhov03: As I see, it's called at most once and only on cold path, isn't it?

runtime-common/stdlib/serialization/serialization-context.h
mkornaukhov03: `msgpack_error{}` is forced by formatter to be `msgpack_error;`. I would prefer explicit init here

runtime-common/stdlib/msgpack/adaptors.h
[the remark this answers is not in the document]
mkornaukhov03: As i remember, we have found out that call of such a getter is relatively expensive, so I think we shouldn't call it on fast path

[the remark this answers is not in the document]
mkornaukhov03: I can add comment about it

runtime-common/stdlib/msgpack/object_visitor.h
mkornaukhov03: done

runtime-common/stdlib/serialization/serialization-context.h
mkornaukhov03: Done

runtime-common/stdlib/serialization/serialization-context.h
```diff
@@ -4,15 +4,31 @@
...
+
 #include "common/mixin/not_copyable.h"
+#include "common/wrappers/string_view.h"
 #include "runtime-common/core/runtime-core.h"
 
 struct SerializationLibContext final : private vk::not_copyable {
   string last_json_processor_error;
 
   // msgpack serialization
+  inline static constexpr size_t MSGPACK_ERROR_BUFFER_SIZE = 256;
```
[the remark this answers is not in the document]
mkornaukhov03: fixed

runtime-common/stdlib/msgpack/zone.h
apolyakov: I think it will. Now you have two sources of truth: `m_error` and the actual result of `alloc_align` for example

runtime-common/stdlib/serialization/serialization-context.h
apolyakov: Then it's better to keep it as `msgpack_error;`. It's common approach for our source code

mkornaukhov03: Done

runtime-common/stdlib/msgpack/zone.h
mkornaukhov03: fixed

apolyakov (approved):
LGTM

PR #1295, opened 2025-04-22
[k2] add support argc/argv superglobals in k2-cli 

astrophysik:


runtime-light/state/component-state.cpp
```diff
@@ -54,6 +56,40 @@ void ComponentState::parse_runtime_config_arg(std::string_view value_view) noexc
...
+  bool in_quote{};
+  string current_arg{};
+  for (char letter : value_view) {
+    if (std::isspace(letter) && !in_quote && !current_arg.empty()) {
+      command_line_argv.push_back(std::move(current_arg));
+      current_arg = string();
+    } else if (letter == '"') {
+      in_quote = !in_quote;
+    } else if (letter == '\'') {
+      php_warning("in command line arg supported only \" quote");
```
Danil42Russia, line 82: i think you added an extra space

compiler/code-gen/files/init-scripts.cpp
```diff
@@ -65,6 +65,14 @@ void StaticInit::compile(CodeGenerator &W) const {
     W << END << NL << NL;
   }
 
+  if (G->is_output_mode_k2()) {
+    W << "namespace kphp::compiler_interface {" << NL;
+    FunctionSignatureGenerator(W) << ("std::string_view get_main_file_name() ") << BEGIN;
```
apolyakov, line 70: let's make this functions `constexpr`

runtime-light/core/compiler-info/compiler_interface.h
apolyakov, line 1: please, considering using `-` instead of `_` in file names

runtime-light/state/component-state.h
```diff
@@ -43,6 +44,7 @@ struct ComponentState final : private vk::not_copyable {
 
 private:
   static constexpr std::string_view INI_ARG_PREFIX = "ini ";
+  static constexpr std::string_view COMMAND_LINE_ARG = "command-line";
```
apolyakov, line 47: let's define it as `CLI_ARG_PREFIX = "cli "`

runtime-light/state/component-state.h
```diff
@@ -19,6 +19,7 @@ struct ComponentState final : private vk::not_copyable {
   const uint32_t argc;
   const uint32_t envc;
   mixed runtime_config;
+  array<string> command_line_argv;
```
apolyakov, line 22: `cli_argv`

apolyakov: and you also missed its init

runtime-light/state/component-state.h
```diff
@@ -43,6 +44,7 @@ struct ComponentState final : private vk::not_copyable {
 
 private:
   static constexpr std::string_view INI_ARG_PREFIX = "ini ";
+  static constexpr std::string_view COMMAND_LINE_ARG = "command-line";
   static constexpr std::string_view RUNTIME_CONFIG_ARG = "runtime-config";
```
apolyakov, line 48: please, add whitespace at the end

runtime-light/state/component-state.cpp
```diff
@@ -54,6 +56,40 @@ void ComponentState::parse_runtime_config_arg(std::string_view value_view) noexc
...
+  if (!command_line_argv.empty()) [[unlikely]] {
+    php_warning("command line argument support no more one usage");
+    return;
+  }
+
+  const auto& main_file_view{kphp::compiler_interface::get_main_file_name()};
+  command_line_argv.push_back(string(main_file_view.data(), main_file_view.size()));
+
+  bool in_quote{};
+  string current_arg{};
```
apolyakov, line 74: It seems like an error. You need to set special reference counter for each reference counted object

runtime-light/state/component-state.cpp
```diff
@@ -54,6 +56,40 @@ void ComponentState::parse_runtime_config_arg(std::string_view value_view) noexc
...
+  for (char letter : value_view) {
```
apolyakov, line 75: It's not actually `letter` since there might be other symbols

runtime-light/state/component-state.cpp
```diff
@@ -54,6 +56,40 @@ void ComponentState::parse_runtime_config_arg(std::string_view value_view) noexc
...
+    if (std::isspace(letter) && !in_quote && !current_arg.empty()) {
```
apolyakov, line 76: Looks overcomplicated. Let's use standard string view functions instead

apolyakov (changes requested):

compiler/code-gen/files/init-scripts.cpp
astrophysik: `clang++` gives error with message that `inline function ... is not defined`

PR #1304, opened 2025-05-06
fixed a bug with bool type mapping in vkext

Danil42Russia:
Fixed a bug where `tlgen` and `tl2php` converted the `bool` type as `boolean` and left `vkext` as is

https://github.com/VKCOM/kphp/blob/51c36da95e307dcd030c0466d3e2be6060ffaa5f/common/tl2php/combinator-to-php.cpp#L90


DrDet (approved):
Nice

PR #1305, opened 2025-05-07
[k2] enhance logs

apolyakov:
Changes:

1. introduce new logging system utilizing `std::format`
2. migrate `runtime-light` to new logging system
3. optimize logging system to conditionally create log string only when log level is sufficient
4. add `info` level log messages to provide insights into the initialization of RPC and HTTP servers

### DIFF
--- a/runtime-light/allocator/runtime-light-allocator.cpp
+++ b/runtime-light/allocator/runtime-light-allocator.cpp
@@ -63,6 +63,7 @@ void RuntimeAllocator::free() {
 }
 
 void* RuntimeAllocator::alloc_script_memory(size_t size) noexcept {
+  kphp::log::assertion(size != 0);
   void* mem{memory_resource.allocate(size)};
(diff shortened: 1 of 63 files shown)


mkornaukhov03 (dismissed):

runtime-light/allocator/runtime-light-allocator.cpp
```diff
@@ -61,65 +63,56 @@ void RuntimeAllocator::free() {
 }
 
 void* RuntimeAllocator::alloc_script_memory(size_t size) noexcept {
-  php_assert(size > 0);
```
Danil42Russia, line 64: are you sure that's the right thing to do? Seems like a logical check

apolyakov: Generally it should not happen, but let's add them back

mkornaukhov03 (approved):

PR #1313, opened 2025-05-14
move http_build_query to runtime-common

apolyakov:


runtime-common/stdlib/server/url-functions.h
```diff
@@ -23,3 +34,82 @@ mixed f$parse_url(const string& s, int64_t component = -1) noexcept;
...
+      result.append(key_value_param);
+      first = false;
+    }
+  }
+  return result;
+}
+
+template<class T>
+string build_query_get_param(const string& key, const T& param, const string& arg_separator, int64_t encoding_type) noexcept {
+  if (f$is_null(param)) {
```
Danil42Russia, line 63: maybe then add `std::optional` for the `param` parameter?

runtime/openssl.cpp
```diff
@@ -32,6 +32,7 @@
 #include "runtime-common/stdlib/string/string-functions.h"
 #include "runtime/allocator.h"
 #include "runtime/array_functions.h"
+#include "runtime/context/runtime-context.h"
```
astrophysik, line 35: Why this include is added? Was it missed?

PR #1318, opened 2025-05-16
[k2] implement strtotime

apolyakov:
Implement `strtotime` function using `timelib` library. As `timelib` doesn't support customising a memory manager, we use `libc_alloc_guard` to allow it to call `malloc`, `free` etc. These symbols are overridden to call script allocator instead of the system one. Also, in this PR we override `strdup` function since it's also used by `timelib` internally.

### DIFF
--- a/builtin-functions/kphp-light/stdlib/time-functions.txt
+++ b/builtin-functions/kphp-light/stdlib/time-functions.txt
@@ -7,11 +7,15 @@ function mktime ($h ::: int = PHP_INT_MIN, $m ::: int = PHP_INT_MIN, $s ::: int 
 
 function date ($format ::: string, $timestamp ::: int = PHP_INT_MIN) ::: string;
 
+function date_default_timezone_get() ::: string;
+
 function date_default_timezone_set ($s ::: string) ::: bool;
 
 function gmdate ($format ::: string, $timestamp ::: int = PHP_INT_MIN) ::: string;
 
 function time() ::: int;
+
+function strtotime ($time ::: string, $timestamp ::: int = PHP_INT_MIN) ::: int | false;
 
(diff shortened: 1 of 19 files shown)


builtin-functions/kphp-light/stdlib/time-functions.txt
```diff
@@ -7,20 +7,23 @@ function mktime ($h ::: int = PHP_INT_MIN, $m ::: int = PHP_INT_MIN, $s ::: int
...
 
+function date_default_timezone_get() ::: string;
+
 function date_default_timezone_set ($s ::: string) ::: bool;
 
 function gmdate ($format ::: string, $timestamp ::: int = PHP_INT_MIN) ::: string;
 
 function time() ::: int;
 
+function strtotime ($time ::: string, $timestamp ::: int = PHP_INT_MIN) ::: int | false;
```
astrophysik, line 18: Can we do argument $timestamp make optional, like it is in [php](https://www.php.net/manual/ru/function.strtotime.php)?

apolyakov: I don't know. It requires some investigation whether such change brake existing code

astrophysik: I suggest find out it

apolyakov: I think it's out of the scope of this PR. It's enough we have the same interface as in `KPHP`

apolyakov:
> OK NIT: tests/phpt/dl/459_strtotime.php:9 -- I would add 'Etc/GMT-4'

Nice point, but `KPHP` supports 2 timezones only: `Moscow` and `GMT-3`. So, to enable the test with `GMT-4` we need to update timezones in `KPHP` first

PR #1319, opened 2025-05-19
http method to give metrics in `OpenMetrics` format

Danil42Russia:


server/php-master.cpp
```diff
@@ -1192,14 +1196,21 @@ int php_master_http_execute(struct connection *c, int op) {
 
   assert (D->first_line_size > 0 && D->first_line_size <= D->header_size);
 
-  const char *allowed_query = "/server-status";
-  if (D->uri_size == strlen(allowed_query) && strncmp(ReqHdr + D->uri_offset, allowed_query, static_cast<size_t>(D->uri_size)) == 0) {
+  const char *server_stats_query = "/server-status";
```
DrDet:
```suggestion
  const char *server_status_query = "/server-status";
```

common/server/http-stats-t.h
```diff
@@ -0,0 +1,78 @@
+// Compiler for PHP (aka KPHP)
+// Copyright (c) 2025 LLC «V Kontakte»
+// Distributed under the GPL v3 License, see LICENSE.notice.txt
+
+#pragma once
+
+#include "common/stats/provider.h"
+
+class http_stats_t final : public stats_t {
```
DrDet: Let's make not confusing naming and name it `open_metric` in all places
```suggestion
class open_metrics_stats_t final : public stats_t {
```

server/php-master.cpp
```diff
@@ -1192,14 +1196,21 @@ int php_master_http_execute(struct connection *c, int op) {
...
+  if (D->uri_size == strlen(server_stats_query) && strncmp(ReqHdr + D->uri_offset, server_stats_query, static_cast<size_t>(D->uri_size)) == 0) {
     std::string stat_html = get_master_stats_http();
     write_basic_http_header(c, 200, 0, static_cast<int>(stat_html.length()), nullptr, "text/plain; charset=UTF-8");
     write_out(&c->Out, stat_html.c_str(), static_cast<int>(stat_html.length()));
     return 0;
   }
 
+  const char *server_metrics_query = "/server-metrics";
```
DrDet:
```suggestion
  const char *server_metrics_query = "/get-open-metrics-stats";
```

DrDet (changes requested):

server/php-master.cpp
Danil42Russia: +

common/server/http-stats-t.h
Danil42Russia: +

server/php-master.cpp
Danil42Russia: +

PR #1329, opened 2025-05-25
[k2] custom format for floating points in logs

apolyakov:
Introduce `floating_wrapper` to automatically ensure consistent logging of floating-point numbers with a fixed decimal precision. This fixes an issue where `std::format` could use `malloc` to format floating point types.

### DIFF
--- a/runtime-light/utils/logs.h
+++ b/runtime-light/utils/logs.h
@@ -8,6 +8,7 @@
 #include <cstddef>
 #include <format>
 #include <source_location>
+#include <type_traits>
 #include <utility>
 
 #include "runtime-light/k2-platform/k2-api.h"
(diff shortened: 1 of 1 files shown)


runtime-light/utils/logs.h
```diff
@@ -24,25 +25,44 @@ namespace kphp::log {
...
+                                                                                                     impl::wrap_log_argument(std::forward<Args>(args))...)};
   *out = '\0';
   k2::log(std::to_underlying(level), size, log_buffer.data());
 }
 
 } // namespace impl
 
 template<typename... Args>
-[[noreturn]] void error(const std::format_string<Args...>& fmt, Args&&... args) noexcept {
+[[noreturn]] void error(std::format_string<impl::wrapped_arg_t<Args>...> fmt, Args&&... args) noexcept {
```
mkornaukhov03, line 65: `fmt` was passed by const reference, why for now it isn't?

mkornaukhov03 (changes requested):

runtime-light/utils/logs.h
apolyakov: I decided to pass it the same way it's passed to `std::format` -- by value. It's ok since it's just a wrapper around `std::string_view`

mkornaukhov03 (approved):

PR #1336, opened 2025-06-03
[k2] add backtrace in exception

astrophysik:


runtime-light/stdlib/diagnostics/exception-types.h
```diff
@@ -89,7 +113,21 @@ inline string exception_trace_as_string(const Throwable& e) noexcept {
...
+  const size_t num_frames{kphp::diagnostic::backtrace(backtrace)};
+  std::span<void*> backtrace_view{backtrace.data(), num_frames};
+
+  auto backtrace_symbols{error_handling_impl_::format_backtrace_symbols(backtrace_view)};
+  if (!backtrace_symbols.empty()) {
+    e->trace = std::move(backtrace_symbols);
+  }
+  auto backtrace_code_addresses{error_handling_impl_::format_backtrace_addresses(backtrace_view)};
+  if (!backtrace_code_addresses.empty()) {
+    e->trace = std::move(backtrace_code_addresses);
```
Danil42Russia: Could there be a case where `trace` already has something in it after `format_backtrace_symbols` and we overwrite it here?

astrophysik: Yeah, add return

PR #1343, opened 2025-06-11
Fix global vars refcnt

mkornaukhov03:
Imagine the following snippet:
```php
<?php

const PREDEFINED_WILDCARDS = [
  'lol',
  'kek',
];

const ALL = [
  PREDEFINED_WILDCARDS,
  [1, 2, 3],
];

warning((string)get_reference_counter(ALL[0]));
```

It's supposed to print `0x7ffffff0`, special counter value for global constants. But previously it printed `2`.
The reason is that `PREDEFINED_WILDCARDS` is and object of type `array<string>`, `ALL` is of type `array<array<mixed>>`, and during global constants initialization, `PREDEFINED_WILDCARDS` was conversed into `array<mixed>`, and counter was changed. For now, we create a copy global constant of `PREDEFINED_WILDCARDS` that is of type `array<mixed>` and behavior is expected now.

### DIFF
--- a/compiler/pipes/optimization.cpp
+++ b/compiler/pipes/optimization.cpp
@@ -20,6 +20,19 @@
 
 namespace {
 
+// This function is copypasted from CollectConstVars pass
+// Should be united in a common part someday
+int get_expr_dep_level(VertexPtr vertex) {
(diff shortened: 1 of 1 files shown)


apolyakov (approved):
:+1:

PR #1346, opened 2025-06-17
make confdata reader more reliable

PetrShumilov:
Goals of this PR:
* Prevent exiting if the next binlog is not available
* Wait for the new binlog to be delivered and send alerts after a specified delay

### DIFF
--- a/common/binlog/binlog-buffer-aio.cpp
+++ b/common/binlog/binlog-buffer-aio.cpp
@@ -14,12 +14,11 @@
 #include <unistd.h>
 
 #include "common/binlog/binlog-buffer.h"
+#include "common/binlog/binlog-stats.h"
 #include "common/kfs/kfs-binlog.h"
-#include "common/kprintf.h"
 #include "common/precise-time.h"
 #include "common/server/engine-settings.h"
 #include "common/server/signals.h"
-#include "server/confdata-stats.h"
 
 DECLARE_VERBOSITY(binlog_buffers);
 
@@ -62,7 +61,7 @@ static int bbw_local_replica_rotate(bb_writer_t* W, bb_rotation_point_t* P) {
   assert(Q->log_pos < P->log_pos);
   kfs_file_handle_t Binlog = Q->Binlog;
   P->Binlog = next_binlog(Binlog);
-  auto& confdata_stats = ConfdataStats::get();
+  auto& binlog_reader_stats = binlog_reader_stats::get();
 
   if (P->Binlog == NULL) {
     update_replica(Binlog->info->replica, 0);
@@ -74,16 +73,14 @@ static int bbw_local_replica_rotate(bb_writer_t* W, bb_rotation_point_t* P) {
       W->unsuccessful_rotation_first_ts = time(NULL);
     }
     W->unsuccessful_rotation_attempts++;
-    std::chrono::seconds elapsed_time{time(NULL) - W->unsuccessful_rotation_first_ts};
+    std::chrono::seconds elapsed_time{std::chrono::duration_cast<std::chrono::seconds>(std::chrono::system_clock::now().time_since_epoch()).count() -
+                                      W->unsuccessful_rotation_first_ts};
     kprintf("attempt #%d (%" PRIu64 " secs): cannot find next binlog file after %s, log position %lld\n", W->unsuccessful_rotation_attempts,
             elapsed_time.count(), Binlog->info->filename, P->log_pos);
 
-    const auto timeout = get_engine_settings()->next_binlog_part_not_found_alert_timeout;
-    if (elapsed_time > timeout) {
-      send_not_too_much_messages_to_assertion_chat("Waiting for the next binlog part for too long: %d attempts, %" PRIu64 " seconds, binlog name %s",
-                                                   W->unsuccessful_rotation_attempts, elapsed_time.count(), Binlog->info->filename);
-      confdata_stats.next_binlog_wait_time = elapsed_time;
-      confdata_stats.next_binlog_expectator_name = Binlog->info->filename;
+    if (elapsed_time > get_engine_settings()->next_binlog_part_not_found_alert_timeout) {
+      binlog_reader_stats.next_binlog_wait_time = elapsed_time;
+      binlog_reader_stats.next_binlog_expectator_name = Binlog->info->filename;
     }
 
     return REPLAY_BINLOG_STOP_READING;
@@ -103,8 +100,8 @@ static int bbw_local_replica_rotate(bb_writer_t* W, bb_rotation_point_t* P) {
   assert(bb_buffer_log_cur_pos(B) + 36 == P->RT.next_log_pos);
 
   W->unsuccessful_rotation_attempts = 0;
-  confdata_stats.next_binlog_wait_time = {0s};
-  confdata_stats.next_binlog_expectator_name = "";
+  binlog_reader_stats.next_binlog_wait_time = {0s};
+  binlog_reader_stats.next_binlog_expectator_name = "";
   return 0;
 }
 
--- a/common/binlog/binlog-buffer-aio.cpp
+++ b/common/binlog/binlog-buffer-aio.cpp
@@ -4,20 +4,26 @@
 
 #include <aio.h>
 #include <assert.h>
+#include <chrono>
+#include <cinttypes>
+#include <cstdint>
 #include <cstdlib>
 #include <errno.h>
 #include <fcntl.h>
 #include <float.h>
 #include <unistd.h>
 
+#include "common/binlog/binlog-buffer.h"
 #include "common/kfs/kfs-binlog.h"
 #include "common/kprintf.h"
 #include "common/precise-time.h"
+#include "common/server/engine-settings.h"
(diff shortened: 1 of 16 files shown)


server/confdata-binlog-replay.cpp
```diff
@@ -39,17 +40,20 @@
...
 
 struct {
-  const char *binlog_mask{nullptr};
   size_t memory_limit{2u * 1024u * 1024u * 1024u};
   double soft_oom_threshold_ratio{CONFDATA_DEFAULT_SOFT_OOM_RATIO};
   double hard_oom_threshold_ratio{CONFDATA_DEFAULT_HARD_OOM_RATIO};
-  double confdata_update_timeout_sec = 0.3;
+  double confdata_update_timeout_sec {0.3};
+  struct {
+    std::chrono::seconds how_long_wait_until_alert{120};
```
DrDet: Let's make it clear here what kind of wait
```suggestion
    std::chrono::seconds how_long_wait_for_next_rotated_binlog_until_alert{120};
```

common/binlog/binlog-buffer-aio.cpp
```diff
@@ -67,10 +74,19 @@ static int bbw_local_replica_rotate(bb_writer_t* W, bb_rotation_point_t* P) {
...
-    kprintf("attempt #%d (%d secs): cannot find next binlog file after %s, log position %lld\n", W->unsuccessful_rotation_attempts, t, Binlog->info->filename,
-            P->log_pos);
-    return (t > 120) ? REPLAY_BINLOG_ERROR : REPLAY_BINLOG_STOP_READING;
+    std::chrono::seconds elapsed_time{time(NULL) - W->unsuccessful_rotation_first_ts};
+    kprintf("attempt #%d (%" PRIu64 " secs): cannot find next binlog file after %s, log position %lld\n", W->unsuccessful_rotation_attempts,
+            elapsed_time.count(), Binlog->info->filename, P->log_pos);
+
+    const auto timeout = get_engine_settings()->next_binlog_part_not_found_alert_timeout;
+    if (elapsed_time > timeout) {
+      send_not_too_much_messages_to_assertion_chat("Waiting for the next binlog part for too long: %d attempts, %" PRIu64 " seconds, binlog name %s",
```
DrDet: I think this function doesn't work for us. Actually we need to be able to send such errors in sentry via json logs. Usually we use `log_server_error` for it, but it can't be used in common code. Can we somehow send this error via `log_server_error` on higher abstraction layer somewhere in `server/...`?

common/binlog/binlog-buffer-aio.cpp
```diff
@@ -4,21 +4,27 @@
...
 
+#include "common/binlog/binlog-buffer.h"
 #include "common/kfs/kfs-binlog.h"
 #include "common/kprintf.h"
 #include "common/precise-time.h"
+#include "common/server/engine-settings.h"
 #include "common/server/signals.h"
-
-#include "common/binlog/binlog-buffer.h"
+#include "server/confdata-stats.h"
```
DrDet: IMHO It's not realy good idea to have not-common dependency in common code

common/binlog/binlog-buffer-replay.cpp
```diff
@@ -323,7 +342,6 @@ static int process_replay_work(void* extra, const void* data, int len) {
       if (l > s) {
         l = s;
       }
-      l &= -4;
```
DrDet, line 326: I checked reference implementation and it differs from our implementation not only in this line. Won't such 'partialy' apllied changes broke something?

server/confdata-binlog-replay.cpp
```diff
@@ -1104,6 +1113,14 @@ void confdata_binlog_update_cron() noexcept {
     return;
   }
 
+  // Force statistics update if binlog hasn't been rotated successfully in specific time
+  if (BinlogBufferWriter.unsuccessful_rotation_attempts > 0) {
+    std::chrono::seconds wait_rotation_time {time(NULL) - BinlogBufferWriter.unsuccessful_rotation_first_ts};
```
DrDet: It's a bit strange to combine old C-style `time(NULL)` with `std::chrono`, maybe it's better to use `std::chrono::system_clock::now()`

server/confdata-binlog-replay.cpp
```diff
@@ -1104,6 +1113,14 @@ void confdata_binlog_update_cron() noexcept {
     return;
   }
 
+  // Force statistics update if binlog hasn't been rotated successfully in specific time
+  if (BinlogBufferWriter.unsuccessful_rotation_attempts > 0) {
```
DrDet, line 1117: Also I think we can easily fire json log error here (see another comment about `send_not_too_much_messages..`

DrDet (changes requested):

server/confdata-binlog-replay.cpp
PetrShumilov: Sure, fixed

common/binlog/binlog-buffer-aio.cpp
PetrShumilov: Moved on `log_server_error`, fixed

common/binlog/binlog-buffer-aio.cpp
PetrShumilov: Yes, an infiltration of `server` into `common` has been removed

server/confdata-binlog-replay.cpp
PetrShumilov: Fixed

server/confdata-binlog-replay.cpp
PetrShumilov: Added

common/binlog/binlog-stats.h
```diff
@@ -0,0 +1,22 @@
+//
```
astrophysik: Let's add header `//  Compiler for PHP (aka KPHP)...` and use `#pragma once`

server/statshouse/statshouse-manager.h
```diff
@@ -89,6 +90,8 @@ class StatsHouseManager : vk::not_copyable {
 
   void add_confdata_master_stats(const ConfdataStats &confdata_stats);
 
+  void add_confdata_binlog_reader_stats(const binlog_reader_stats& confdata_stats);
```
astrophysik: NIT: what about add `noexcept` in function signuture?

server/php-engine.cpp
```diff
@@ -2351,7 +2356,8 @@ void parse_main_args(int argc, char *argv[]) {
                                                                    "Initial binlog is readed with x10 times larger timeout");
   parse_option("confdata-soft-oom-ratio", required_argument, 2039, "Memory limit ratio to start ignoring new keys related events (default: 0.85)."
                                                                    "Can't be > hard oom ratio (0.95)");
-  parse_option("kml-dir", required_argument, 2040, "Directory that contains .kml files");
+  parse_option("confdata-how-long-wait-binlog-until-alert", required_argument, 2040, "Time in seconds to wait before starting to alert if the next binlog part is not found");
```
astrophysik, line 2364: Why option `kml-dir` was changed ?

server/confdata-binlog-replay.cpp
```diff
@@ -39,17 +40,20 @@
...
+    std::chrono::seconds how_long_wait_for_next_binlog_until_alert{120};
+    const char *mask{nullptr};
+  } binlog_reader;
   std::unique_ptr<re2::RE2> key_blacklist_pattern;
   std::forward_list<vk::string_view> force_ignore_prefixes;
   std::unordered_set<vk::string_view> predefined_wildcards;
 
   bool is_enabled() const noexcept {
-    return binlog_mask;
+    return binlog_reader.mask;
```
astrophysik: Maybe add here comparison with nullptr to avoid implicit conversation?

astrophysik (changes requested):

server/php-engine.cpp
```diff
@@ -2227,6 +2227,11 @@ int main_args_handler(int i, const char *long_option) {
       return res;
     }
     case 2040: {
+      return parse_numeric_option<uint32_t>(long_option, 0, std::numeric_limits<uint32_t>::max(), [](uint32_t t) {
```
astrophysik, line 2231: Are we really want to use here `std::numeric_limits<uint32_t>::max()` as max value? Maybe make it x * default_timeout?

server/confdata-binlog-replay.cpp
PetrShumilov: Fixed

server/statshouse/statshouse-manager.h
PetrShumilov: Thanks, added

server/php-engine.cpp
PetrShumilov: I actually didn't change the `kml-dir` option. I just moved new option more near to confdata-related options

common/binlog/binlog-stats.h
PetrShumilov: Thanks, fixed

DrDet (approved):

astrophysik (approved):

PR #1347, opened 2025-06-20
add runtime builtin stats

astrophysik:
General
----------

This pull request enables the collection of statistics on built-in function calls at runtime. To generate a special call that can record statistics, you need to provide the compiler with a list of functions using the option `--tracked-builtins-list` or env `KPHP_TRACKED_BUILTINS_LIST`. To enable runtime statistic collection, the server provides the option `--enable-request-builtin-stats`; by default, this statistic collection is disabled.

Some details
-----------------

* In the `runtime_builtin_stats` namespace, there are two flags that might seem identical. They are separated to prevent attempts to record statistics during script initialization.

* The storage of statistics is allocated in script memory, and to prevent issues before script memory initialization, it is wrapped in an optional.

* The option --tracked-builtins-list was added to the compiler to reduce the overhead of checking a global variable for already processed built-ins in runtime-light.


compiler/compiler-core.cpp
```diff
@@ -335,6 +336,16 @@ void CompilerCore::require_function(FunctionPtr function, DataStream<FunctionPtr
...
+void CompilerCore::parse_tracked_builtins(const std::string& file_name) {
+  std::ifstream tracked_builtins_file(file_name);
+  kphp_error(tracked_builtins_file.is_open(), "Cannot open file with list of tracked builtins");
+
+  std::string function_name;
+  while (tracked_builtins_file >> function_name) {
+    tracked_builtins.emplace(std::move(function_name));
+  }
+}
+
```
mkornaukhov: Is file convenient?
How to add it in CI/CD apps? Aren't env vars more convenient for that?

compiler/compiler-core.h
```diff
@@ -141,6 +142,11 @@ class CompilerCore {
   std::vector<ModulitePtr> get_modulites();
   const ComposerAutoloader &get_composer_autoloader() const;
 
+  void parse_tracked_builtins(const std::string& file_name);
+  auto & get_tracked_builtins() const noexcept {
```
mkornaukhov: `const auto &`

compiler/compiler-settings.cpp
```diff
@@ -243,6 +243,12 @@ void CompilerSettings::init() {
   }
   functions_file.value_ = get_full_path(functions_file.get());
 
+  if (tracked_builtins_file.value_.empty()) {
+    if (!is_k2_mode) {
+      tracked_builtins_file.value_ = kphp_src_path.get() + "builtin-functions/tracked/k2-unsupported.txt";
+    }
```
mkornaukhov: Seems strange a little bit. Really, it will always use `k2-unsupported.txt` file from the repo, this is not flexible

runtime-common/core/allocator/script-allocator.h
```diff
@@ -16,6 +16,11 @@ template<typename T>
 struct script_allocator {
   using value_type = T;
 
+  script_allocator() = default;
+
+  template<typename U>
+  script_allocator(const script_allocator<U>&) {}
+
```
mkornaukhov: What for?

mkornaukhov03 (changes requested):

compiler/compiler-core.h
astrophysik: fix

compiler/compiler-settings.cpp
astrophysik: make it as env and option

runtime-common/core/allocator/script-allocator.h
astrophysik: Required for use in unordered_map

compiler/compiler-core.cpp
astrophysik: Make it as option and env

PR #1349, opened 2025-06-20
Implement special pre-fork run once mode

DrDet:
This PR combines already existing behaviour of `--workers-num N` and `--once` options. To enable this mode one needs just specify both options on start.

New special mode works as the following:
1. Starts N workers, as usual in pre-fork mode from option `--workers-num N`
2. Sends dummy RPC request to itself to run script in each worker M times, as usual with `--once [M]` option (by default `M = 1`)
3. When script is finished as many times as specified, worker process terminates
4. When master process realizes that some workers are terminated it recreates them and execution in each new worker continues from step 2

The main goal of such a strange mode - to automatically enable subsystems designed only for master-workers model in **CLI** mode (confdata, instance cache, errors and traces collecting via json logs, runtime metrics collecting from master cron, etc.)

### DIFF
--- a/server/php-engine.cpp
+++ b/server/php-engine.cpp
@@ -1436,7 +1436,7 @@ void reopen_json_log() {
   }
 }
 
-void generic_event_loop(WorkerType worker_type, bool init_and_listen_rpc_port) noexcept {
+void generic_event_loop(WorkerType worker_type, bool init_and_listen_rpc_port, bool invoke_dummy_self_rpc_request) noexcept {
   if (master_flag && logname_pattern != nullptr) {
     reopen_logs();
     reopen_json_log();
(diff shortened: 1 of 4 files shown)


server/php-engine.cpp
```diff
@@ -2448,9 +2448,14 @@ int run_main(int argc, char **argv, php_mode mode) {
...
 
   if (run_once) {
-    master_flag = 0;
-    rpc_port = -1;
-    setvbuf(stdout, nullptr, _IONBF, 0);
+    if (master_flag) {
+      kprintf("running in special run once pre-fork mode\n");
+      run_once_prefork_mode = true;
+    } else {
+      master_flag = 0;
```
astrophysik, line 2455: I think this is no longer necessary.

DrDet: why? run-once mode will break

PetrShumilov (approved):
OK

PR #1354, opened 2025-06-26
Apolyakov/k2 optimize output buffers

apolyakov:


runtime-light/stdlib/output/output-state.h
```diff
@@ -0,0 +1,84 @@
...
+
+#include "common/mixin/not_copyable.h"
+#include "runtime-common/core/runtime-core.h"
+
+namespace kphp::output {
+
+class output_buffers {
+  static constexpr size_t MAX_BUFFERS = 50;
+
+  std::optional<string_buffer> m_system_buffer;
```
mkornaukhov, line 25: Why not just string_buffer?

runtime-light/stdlib/output/output-state.h
```diff
@@ -0,0 +1,84 @@
...
+  std::array<std::optional<string_buffer>, MAX_BUFFERS> m_user_buffers{};
+  decltype(m_user_buffers)::iterator m_user_buffer_iterator{m_user_buffers.end()};
```
mkornaukhov: Wouldn't `auto` instead of `decltype(m_user_buffers)::iterator` better? Puke syntax

runtime-light/server/http/init-functions.cpp
```diff
@@ -329,12 +331,23 @@ void init_server(tl::TLBuffer& tlb) noexcept {
...
   auto& http_server_instance_st{HttpServerInstanceState::get()};
 
   string body{};
   if (http_server_instance_st.http_method != method::head) {
-    body = output.str();
+    auto& output_instance_st{OutputInstanceState::get()};
+    const auto system_buffer{output_instance_st.output_buffers.system_buffer()};
+    const auto user_buffers{output_instance_st.output_buffers.user_buffers()};
+
+    const auto body_size{std::ranges::fold_left(user_buffers | std::views::transform([](const auto& buffer) noexcept { return buffer.size(); }),
```
mkornaukhov, line 343: It's visual garbage, please, lets rewrite with for-loop

mkornaukhov03 (approved):

runtime-light/stdlib/output/output-state.h
apolyakov: `auto` is not allowed for non-static members

PR #1356, opened 2025-06-28
[k2] use ordered containers instead of unordered ones in the scheduler

apolyakov:
`perf` shows that ordered containers in scheduler produce less page faults compared to unordered ones


apolyakov:
Closed since it doesn't have expected effect on production code

PR #1363, opened 2025-07-09
Add kml_available_models()

mkornaukhov03:


runtime/kphp_ml/kphp_ml_interface.cpp
```diff
@@ -125,6 +125,15 @@ bool f$kml_model_exists(const string& model_name) {
   return kphp_ml_find_loaded_model_by_name(model_name) != nullptr;
 }
 
+array<string> f$kml_available_models() {
+  array<string> response;
+  response.reserve(loaded_models.size(), true);
```
Danil42Russia, line 130: you can shorten it to `array<string> response(array_size( loaded_models.size(), true))`, but that's up to you :)

PetrShumilov (approved):
LGTM

PR #1364, opened 2025-07-09
[k2] new coroutine scheduler

apolyakov:


compiler/code-gen/vertex-compiler.cpp
```diff
@@ -923,7 +923,7 @@ void compile_func_call(VertexAdaptor<op_func_call> root, CodeGenerator &W, func_
 
     if (mode == func_call_mode::fork_call) {
       if (func->is_interruptible) {
-        W << "(co_await start_fork_t{" << FunctionName(func);
```
astrophysik, line 926: Since now fork starting isn't coroutine what about to don't mark function, that call `fork` interruptible? 
This happens in `calc-bad-vars.cpp:561`. 
In this example
```php
function foo() {
    sched_yield();
    return null;
}

function start() {
    $f = fork(lol());
    return $f;
}

$f = start();
wait($f);
```
Body of function `start` will be
```c++
kphp::coro::task<future< Optional < bool > >> f$start() noexcept  {
  future< Optional < bool > > v$f = 0;
  v$f = (kphp::forks::start(f$lol()));
  co_return v$f;
}
```
But it doesn't await anything

runtime-light/coroutine/concepts.h
```diff
@@ -0,0 +1,72 @@
...
+#include "runtime-light/coroutine/async-stack.h"
+#include "runtime-light/coroutine/shared-task.h"
+#include "runtime-light/coroutine/task.h"
+
+namespace kphp::coro::concepts {
+
+namespace detail {
+
+template<typename type, typename... types>
+concept in_types = (std::same_as<type, types> || ...);
```
astrophysik: NIT: maybe move `in_types` in common concepts? To use it other contexts

runtime-light/coroutine/detail/poll-info.h
```diff
@@ -0,0 +1,65 @@
...
+#include "runtime-light/coroutine/poll.h"
+#include "runtime-light/k2-platform/k2-api.h"
+
+namespace kphp::coro::detail {
+
+struct poll_info {
+  using timed_events = kphp::stl::multimap<k2::TimePoint, detail::poll_info&, kphp::memory::script_allocator>;
+  using awaiting_polls = kphp::stl::multimap<k2::descriptor, detail::poll_info&, kphp::memory::script_allocator>;
+
+  k2::descriptor m_descriptor{k2::INVALID_PLATFORM_DESCRIPTOR};
```
astrophysik, line 21: I see that, `m_descriptor` doesn't changed. What about to make it `const`? Same for m_poll_op

runtime-light/coroutine/detail/poll-info.h
```diff
@@ -0,0 +1,65 @@
...
+#include <optional>
+
+#include "runtime-common/core/allocator/script-allocator.h"
+#include "runtime-common/core/std/containers.h"
+#include "runtime-light/coroutine/poll.h"
+#include "runtime-light/k2-platform/k2-api.h"
+
+namespace kphp::coro::detail {
+
+struct poll_info {
```
astrophysik, line 17: Are we worried about poll_info size? What about sorting his fields to reduce it?

runtime-light/coroutine/io-scheduler.h
```diff
@@ -0,0 +1,562 @@
...
+      case k2::UpdateStatus::UpdateExisted:
+        process_update(descriptor);
+        break;
+      case k2::UpdateStatus::NewDescriptor:
+        process_accept(descriptor);
+        break;
+      case k2::UpdateStatus::NoUpdates:
+        [[fallthrough]];
+      default:
+        std::unreachable();
```
astrophysik: I think better kphp::assert(false) or something else. std::unreachable literally calls ub

runtime-light/streams/stream.h
```diff
@@ -0,0 +1,229 @@
...
+      break;
+    case kphp::coro::poll_status::closed:
+      [[likely]] co_return std::expected<void, int32_t>{};
+    case kphp::coro::poll_status::error:
+      co_return std::unexpected{k2::errno_efault};
+    case kphp::coro::poll_status::timeout:
+      co_return std::unexpected{k2::errno_ecanceled};
+    }
+  }
+  std::unreachable();
```
astrophysik: Also, as for me it is not good practice

runtime-light/coroutine/detail/task-self-deleting.h
```diff
@@ -0,0 +1,103 @@
...
+  }
+};
+
+class task_self_deleting {
+  promise_self_deleting& m_promise;
+
+public:
+  using promise_type = promise_self_deleting;
+
+  explicit task_self_deleting(promise_self_deleting& promise) noexcept
```
astrophysik, line 67: Maybe hide constructor of task_self_deleting? If users should create it via `make_task_self_deleting`

runtime-light/coroutine/detail/when-all.h
```diff
@@ -0,0 +1,290 @@
...
+  auto start(when_all_latch& latch, kphp::coro::async_stack_frame& caller_async_stack_frame, void* return_address) noexcept -> void {
+    m_coroutine.promise().start(latch, caller_async_stack_frame, return_address);
+  }
+
+public:
+  // to be able to call start()
+  template<typename task_container_type>
+  friend class when_all_ready_awaitable;
+
+  explicit when_all_task(std::coroutine_handle<promise_type> coroutine) noexcept
```
astrophysik, line 259: Let's also hide constructor of `when_all_task` if it is created via `make_when_all_task`

runtime-light/coroutine/detail/task-self-deleting.h
apolyakov: I think it's ok to keep it as it is. It's hidden in `detail` namespace

runtime-light/coroutine/detail/when-all.h
apolyakov: Same as for `task-self-deleting`

runtime-light/coroutine/detail/poll-info.h
apolyakov: It's actually changed by `io_scheduler::process_accept` for example

astrophysik (approved):

PR #1367, opened 2025-07-17
fixed an error when giving `OpenMetrics` statistics

Danil42Russia:


common/server/stats.cpp
```diff
@@ -39,9 +39,8 @@ const char* engine_default_char_stats() {
   return engine_default_prepare_stats(tl_stats_t{}, NULL);
 }
 
-const char* engine_default_open_metrics_stat_stats() {
-  // TODO: also return the size
-  return engine_default_prepare_stats(open_metrics_stats_t{}, nullptr);
+std::pair<std::string_view, std::size_t> engine_default_open_metrics_stat_stats() {
```
astrophysik: Let's use just `std::string_view`. It is already contains size

astrophysik: NIT: It is better to mark function as `noexcept`

server/php-master.cpp
```diff
@@ -1083,7 +1083,7 @@ int php_master_rpc_stats(const std::optional<std::vector<std::string>> &sorted_f
   return 0;
 }
 
-std::string_view get_open_metrics_stats_http() noexcept {
+std::pair<std::string_view, std::size_t> get_open_metrics_stats_http() noexcept {
```
astrophysik: Just `std::string_view`

astrophysik (dismissed):

common/server/stats.h
```diff
@@ -16,7 +16,7 @@ char* get_engine_default_prepare_stats_buffer();
 std::pair<char*, int> engine_default_prepare_stats_with_tag_mask(stats_t&& stats, const char* stats_prefix, unsigned int tag_mask);
 char* engine_default_prepare_stats(stats_t&& stats, const char* stats_prefix);
 const char* engine_default_char_stats();
-const char* engine_default_open_metrics_stat_stats();
+std::string_view engine_default_open_metrics_stat_stats() noexcept;
```
DrDet, line 20: let's add `include <string_view>`

Danil42Russia: done

PR #1375, opened 2025-07-23
[k2] implement error_reporting

astrophysik:


runtime-common/stdlib/error-handling/error-handling-context.h
apolyakov, line 1: nitpick: we already have `diagnostics` folder, why not just add these files there?

runtime-common/stdlib/error-handling/error-handling-context.h
```diff
@@ -0,0 +1,22 @@
...
+#include "common/mixin/not_copyable.h"
+
+namespace php_warning_levels {
+constexpr int32_t PHP_WARNING_MINIMUM_LEVEL = 0;
+constexpr int32_t DEFAULT_WARNING_LEVEL = 2;
+constexpr int32_t PHP_WARNING_MAXIMUM_LEVEL = 3;
+} // namespace php_warning_levels
+
+struct ErrorHandlingContext final : private vk::not_copyable {
+  int32_t php_warning_level{php_warning_levels::DEFAULT_WARNING_LEVEL};
```
apolyakov: Isn't it better to have it as `int64_t`? With `int32_t` there are implicit casts spread across runtime

runtime-light/k2-platform/k2-api.h
```diff
@@ -303,6 +303,10 @@ inline size_t backtrace(std::span<void*> buffer) noexcept {
   return k2_backtrace(buffer.data(), buffer.size());
 }
 
+inline bool is_instance_state_available() noexcept {
```
apolyakov: (◉ _ ◉)

I think this should not exist in k2 API

runtime-light/stdlib/diagnostics/backtrace.cpp
```diff
@@ -44,8 +44,8 @@ size_t async_frames(std::span<void*> addresses, kphp::coro::async_stack_frame* t
 namespace kphp::diagnostic::impl {
 
 size_t async_backtrace(std::span<void*> addresses) noexcept {
-  if (const auto* instance_state{k2::instance_state()}; instance_state != nullptr) [[likely]] {
-    const auto& async_stack_root{instance_state->coroutine_instance_state.coroutine_stack_root};
+  if (k2::is_instance_state_available()) [[likely]] {
```
apolyakov: Instead of `k2::is_instance_state_available` I'd suggest adding following function:
`std::optional<std::reference_wrapper<InstanceState>> InstanceState::try_get()`

runtime-light/stdlib/diagnostics/error-handling-state.cpp
```diff
@@ -0,0 +1,13 @@
+//  Compiler for PHP (aka KPHP)
+//  Copyright (c) 2025 LLC «V Kontakte»
+//  Distributed under the GPL v3 License, see LICENSE.notice.txt
+
+#include "runtime-light/stdlib/diagnostics/error-handling-state.h"
+
+#include "runtime-light/k2-platform/k2-api.h"
```
apolyakov: Looks like redundant includes

apolyakov (changes requested):

runtime-common/stdlib/error-handling/error-handling-context.h
astrophysik: fix

runtime-common/stdlib/error-handling/error-handling-context.h
astrophysik: fix

runtime-light/k2-platform/k2-api.h
astrophysik: remove

runtime-light/stdlib/diagnostics/error-handling-state.cpp
astrophysik: fix

runtime-light/stdlib/diagnostics/error-handling-state.cpp
```diff
@@ -0,0 +1,21 @@
...
+
+#include "runtime-light/stdlib/diagnostics/error-handling-state.h"
+
+#include "runtime-common/core/runtime-core.h"
+#include "runtime-light/state/component-state.h"
+#include "runtime-light/state/instance-state.h"
+
+ErrorHandlingState::ErrorHandlingState() noexcept {
+  const auto& component_st{ComponentState::get()};
+  if (const auto& default_level{component_st.ini_opts.get_value(string(INI_LEVEL_KEY.data(), INI_LEVEL_KEY.size()))};
```
apolyakov: 1. Please, use braced-init everywhere
2. Out of context, `INI_LEVEL_KEY` seems to be an awkward naming. I'd suggest `INI_ERROR_REPORTING_KEY`

apolyakov: We may eliminate one call to `to_int`:

```
auto default_level_str{component_st.ini_opts.get_value(string(INI_LEVEL_KEY.data(), INI_LEVEL_KEY.size()))};
if (default_level_str.empty() || !default_level_str.is_int()) { return; }

const auto default_level{default_level_str.to_int()};
minimum_log_level = static_cast<bool>(SUPPORTED_ERROR_LEVELS & default_level) ? default_level : minimum_log_level;
  }
```

runtime-light/stdlib/diagnostics/logs.h
```diff
@@ -54,7 +55,7 @@ template<typename... Args>
 
 template<typename... Args>
 void warning(std::format_string<impl::wrapped_arg_t<Args>...> fmt, Args&&... args) noexcept {
-  if (std::to_underlying(level::warn) <= k2::log_level_enabled()) {
+  if (std::to_underlying(level::warn) <= k2::log_level_enabled() && (ErrorHandlingState::get().minimum_log_level & E_WARNING) != 0) {
```
apolyakov: It'd be nice to have `ErrorHandlingInstanceState` being able to do something like `k2::log_level_enabled`

runtime-light/stdlib/diagnostics/error-handling-functions.h
```diff
@@ -87,3 +89,34 @@ inline array<array<string>> f$debug_backtrace() noexcept {
   kphp::log::warning("cannot resolve virtual addresses to debug information");
   return {};
 }
+
+inline int64_t f$error_reporting(Optional<int64_t> error_level_opt = {}) noexcept {
+  auto& error_handling_st{ErrorHandlingState::get()};
+  const int64_t prev{error_handling_st.minimum_log_level};
```
apolyakov: No need for variable unless it's necessary. Let's better declare it next to `error_level`. Also, why not name it like `current_error_level`? Then, `error_level` could be renamed to `new_error_level`. It makes the code much more easier to read as we need less context to understand what's going on

PR #1377, opened 2025-07-28
add kphp_slow_curl_response metric

apolyakov:
This PR introduces a new StatsHouse metric to monitor slow CURL responses in KPHP:

- Metric Name: kphp_slow_curl_responses
- Threshold: Records responses that exceed 100ms
- Data Captured:
    - Request duration
    - Target host/endpoint (if available)

### DIFF
--- a/runtime/curl.cpp
+++ b/runtime/curl.cpp
@@ -6,6 +6,7 @@
 
 #include <cstdio>
 #include <cstring>
+#include <string_view>
 
 #include "curl/curl.h"
 #include "curl/easy.h"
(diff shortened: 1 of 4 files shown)


runtime/curl.cpp
```diff
@@ -667,6 +669,8 @@ mixed f$curl_exec(curl_easy easy_id) noexcept {
   easy_context->error_num = dl::critical_section_call(curl_easy_perform, easy_context->easy_handle);
   double request_finish_time = dl_time();
   if (request_finish_time - request_start_time >= long_curl_query) {
+    StatsHouseManager::get().add_slow_net_event_stats(slow_net_event_stats::slow_curl_response_stats{
```
Danil42Russia: let's write a `url` for synchronous `curl`. But for `async` we will have an empty address :(

apolyakov: Done

runtime/curl.cpp
```diff
@@ -667,8 +670,11 @@ mixed f$curl_exec(curl_easy easy_id) noexcept {
   easy_context->error_num = dl::critical_section_call(curl_easy_perform, easy_context->easy_handle);
   double request_finish_time = dl_time();
   if (request_finish_time - request_start_time >= long_curl_query) {
-    kprintf("LONG curl query : %f. Curl id = %d, url = %.100s\n", request_finish_time - request_start_time, easy_context->uniq_id,
-            easy_context->get_info(CURLINFO_EFFECTIVE_URL).as_string().c_str());
+    string curl_url = easy_context->get_info(CURLINFO_EFFECTIVE_URL).as_string();
+    StatsHouseManager::get().add_slow_net_event_stats(
+        slow_net_event_stats::slow_curl_response_stats{slow_net_event_stats::slow_curl_response_stats::curl_kind::sync,
+                                                       std::string_view{curl_url.c_str(), curl_url.size()}, request_finish_time - request_start_time});
```
DrDet, line 676: I think we need to normalize url here somehow to get rid of very long urls. Maybe let's cut it up to the first 100 symbols?

apolyakov: Done

DrDet (approved):

PR #1383, opened 2025-08-05
[k2] add web-builtins

astrophysik:


runtime-light/server/http/http-server-state.h
```diff
@@ -31,6 +32,8 @@ enum status : uint16_t {
   BAD_REQUEST = 400,
 };
 
+enum class response_state : uint8_t { before_send, start_headers_send, headers_sent, body_sent, response_sent };
```
apolyakov: I think this enum has more clarity:
```
enum class response_state : uint8_t {
    not_started,     // Replaces before_send
    sending_headers, // Replaces start_headers_send
    headers_sent,    // Kept (clear enough)
    sending_body,    // More active than body_sent
    completed        // More final than response_sent
};
```

runtime-light/server/http/http-server-state.h
```diff
@@ -60,6 +63,9 @@ struct HttpServerInstanceState final : private vk::not_copyable {
   uint64_t status_code{kphp::http::status::NO_STATUS};
   kphp::http::method http_method{kphp::http::method::other};
   kphp::http::connection_kind connection_kind{kphp::http::connection_kind::close};
+  kphp::http::response_state response_state{kphp::http::response_state::before_send};
+
+  std::optional<kphp::coro::task<>> headers_custom_handler_function;
```
apolyakov: `headers_registered_callback`?

runtime-light/server/http/init-functions.cpp
```diff
@@ -337,6 +337,22 @@ void init_server(kphp::component::stream request_stream) noexcept {
 kphp::coro::task<> finalize_server() noexcept {
   auto& http_server_instance_st{HttpServerInstanceState::get()};
 
+  switch (http_server_instance_st.response_state) {
+  case kphp::http::response_state::before_send:
+  case kphp::http::response_state::start_headers_send:
+  case kphp::http::response_state::headers_sent:
```
apolyakov: It looks like we need to process the case when headers are already sent, so we don't need to send them again in `finalize_server`

runtime-light/server/http/init-functions.cpp
```diff
@@ -337,6 +337,22 @@ void init_server(kphp::component::stream request_stream) noexcept {
...
+  case kphp::http::response_state::body_sent:
+    break;
+  case kphp::http::response_state::response_sent:
+    co_return;
+  }
+
+  if (http_server_instance_st.response_state == kphp::http::response_state::before_send &&
```
apolyakov: It also seems for me that this should better be a part of the state machine in your `switch`

runtime-light/server/http/init-functions.cpp
```diff
@@ -337,6 +337,22 @@ void init_server(kphp::component::stream request_stream) noexcept {
...
+      http_server_instance_st.headers_custom_handler_function.has_value()) {
+    http_server_instance_st.response_state = kphp::http::response_state::start_headers_send;
+    co_await *std::exchange(http_server_instance_st.headers_custom_handler_function, std::nullopt);
+  }
+
   string body{};
```
apolyakov: Sending a body also can be made part of the state machine

apolyakov (approved):

PR #1386, opened 2025-08-08
[k2] implement user-defined tags in logs

astrophysik:


runtime-light/state/instance-state.h
```diff
@@ -103,6 +104,7 @@ struct InstanceState final : vk::not_copyable {
   CurlInstanceState curl_instance_state;
   StringInstanceState string_instance_state;
   SystemInstanceState system_instance_state;
+  kphp::log::Logger instance_logger;
```
apolyakov: I think that loggers should be initialized just after allocators

runtime-light/stdlib/diagnostics/diagnostics.h
apolyakov, line 1: I thinks `logs.h` is the perfect name for that file. `diagnostics.h` is too general I think

runtime-light/stdlib/diagnostics/logger.h
```diff
@@ -0,0 +1,127 @@
...
+}
+
+template<typename T>
+using wrapped_arg_t = std::invoke_result_t<decltype(impl::wrap_log_argument<T>), T>;
+
+} // namespace impl
+
+enum class Level : size_t { error = 1, warn, info, debug, trace }; // NOLINT
+
+struct Record {
```
apolyakov: We use small case for types except for states

runtime-light/stdlib/diagnostics/logger.h
```diff
@@ -0,0 +1,127 @@
...
+  Level level;
+  std::string_view message;
+  std::optional<std::span<void* const>> backtrace;
+};
+
+struct Logger final : vk::not_copyable {
+
+  static void log(Record record) noexcept;
```
apolyakov: it looks like we can simplify the public API of the logger. For example, I think that most of the current API is actually redundant, e.g. `enabled`, `set_*` etc. Speaking about `extra_*`, I guess may just have methods `add_extra_tag`, `remove_extra_tag`, `clear` which are way more general

runtime-light/stdlib/diagnostics/logger.h
```diff
@@ -0,0 +1,127 @@
...
+  static bool enabled(Level level) noexcept;
+
+  Logger& set_extra_tags(std::string_view extra_tags_view) noexcept;
+  Logger& set_extra_info(std::string_view extra_info_view) noexcept;
+  Logger& set_environment(std::string_view environment_view) noexcept;
+
+  static std::optional<std::reference_wrapper<Logger>> try_get() noexcept;
+
+private:
+  static void stateless_log(Record record) noexcept {
```
apolyakov: confusing naming

runtime-light/stdlib/diagnostics/detail/logs.h
```diff
@@ -0,0 +1,65 @@
...
+    const auto [trace_out, trace_size]{std::format_to_n(trace_buffer.data(), trace_buffer.size() - 1, "\n{}", backtrace_symbols)};
+    *trace_out = '\0';
+    return trace_size;
+  } else if (auto backtrace_addresses{kphp::diagnostic::backtrace_addresses(raw_trace)}; !backtrace_addresses.empty()) {
+    const auto [trace_out, trace_size]{std::format_to_n(trace_buffer.data(), trace_buffer.size() - 1, "{}", backtrace_addresses)};
+    *trace_out = '\0';
+    return trace_size;
+  } else {
+    static constexpr std::string_view DEFAULT_TRACE = "[]";
+    std::memcpy(trace_buffer.data(), DEFAULT_TRACE.data(), DEFAULT_TRACE.size() + 1); // add last \0
```
apolyakov: I think you miss the check for `trace_buffer`'s capacity here

runtime-light/stdlib/diagnostics/contextual-logger.cpp
```diff
@@ -0,0 +1,53 @@
...
+
+#include <cstddef>
+#include <format>
+#include <functional>
+#include <optional>
+#include <span>
+#include <string_view>
+#include <utility>
+
+#include "runtime-common/core/allocator/script-allocator.h"
```
apolyakov, line 14: Please remove redundant includes

runtime-light/stdlib/diagnostics/contextual-logger.cpp
```diff
@@ -0,0 +1,53 @@
...
+#include "runtime-light/state/image-state.h"
+#include "runtime-light/state/instance-state.h"
+#include "runtime-light/stdlib/diagnostics/backtrace.h"
+#include "runtime-light/stdlib/diagnostics/detail/logs.h"
+
+namespace kphp::log {
+
+std::optional<std::reference_wrapper<contextual_logger>> contextual_logger::try_get() noexcept {
+  if (const auto* instance_state_ptr{k2::instance_state()}; instance_state_ptr != nullptr) [[likely]] {
+    return const_cast<contextual_logger&>(instance_state_ptr->instance_logger);
```
apolyakov: Why is `const_cast` here? It seems enough to just remove `const` from `instance_state_ptr` declaration

runtime-light/stdlib/diagnostics/contextual-logger.cpp
```diff
@@ -0,0 +1,53 @@
...
+void contextual_logger::log_with_tags(kphp::log::level level, std::optional<std::span<void* const>> trace, std::string_view message) const noexcept {
+  kphp::stl::vector<k2::LogTaggedEntry, kphp::memory::script_allocator> tagged_entries{};
+  if (level == level::warn || level == level::error) {
+    tagged_entries.reserve(extra_tags.size());
+    for (const auto& [key, value] : extra_tags) {
+      tagged_entries.push_back(k2::LogTaggedEntry{.key = key.data(), .value = value.data(), .key_len = key.size(), .value_len = value.size()});
+    }
+  }
+  if (trace.has_value()) {
+    static constexpr std::string_view BACKTRACE_KEY = "trace";
```
apolyakov: Such constants are perfect candidates for declaring them inside the `contextual_logger`. Thus, there will be a single place to check what tags we add to logs

runtime-light/stdlib/diagnostics/detail/logs.h
apolyakov: Can you point where last `'0` is added then?

runtime-light/stdlib/diagnostics/contextual-logger.cpp
```diff
@@ -0,0 +1,53 @@
...
+    static constexpr size_t BACKTRACE_BUFFER_SIZE = 1024UZ * 4UZ;
+    std::array<char, BACKTRACE_BUFFER_SIZE> backtrace_buffer; // NOLINT
+    size_t backtrace_size{impl::resolve_log_trace(backtrace_buffer, *trace)};
+    tagged_entries.push_back(
```
apolyakov, line 43: Can't we also reserve space for that at the beginning of the function?

runtime-light/stdlib/diagnostics/contextual-logger.h
```diff
@@ -0,0 +1,106 @@
...
+  }
+}
+
+inline void contextual_logger::remove_extra_tag(std::string_view key) noexcept {
+  if (auto it{extra_tags.find(key)}; it != extra_tags.end()) {
+    extra_tags.erase(it);
+  }
+}
+
+inline void contextual_logger::clear_extra_tag() noexcept {
```
apolyakov: `clear_extra_tags` as you erase all the elements

runtime-light/stdlib/diagnostics/raw-logger.h
```diff
@@ -0,0 +1,50 @@
...
+#include "runtime-light/k2-platform/k2-api.h"
+#include "runtime-light/stdlib/diagnostics/detail/logs.h"
+
+namespace kphp::log {
+
+struct raw_logger final {
+  raw_logger() noexcept = delete;
+
+  template<typename... Args>
+  static void log(level level, std::optional<std::span<void* const>> trace, std::format_string<impl::wrapped_arg_t<Args>...> fmt, Args&&... args) noexcept {
```
apolyakov: nitpick: I think it's better to use fully qualified names as much as it's possible. It's not so obvious what type you use otherwise

PR #1387, opened 2025-08-08
Implement `getrandmax()` and `srand()` for runtime light

mkornaukhov03:


runtime/math_functions.h
```diff
@@ -25,8 +25,6 @@ int64_t f$rand(int64_t l, int64_t r) noexcept;
 
 int64_t f$rand() noexcept;
 
-int64_t f$getrandmax() noexcept;
```
astrophysik, line 28: Why you remove it from old runtime? If this is really necessary, then it is also necessary to remove the declaration from the `_functions.txt`

astrophysik: I see, that you move definition in runtime-common

apolyakov: I guess since it's moved to `runtime-common`

apolyakov (approved):

astrophysik (approved):

PR #1396, opened 2025-08-27
[k2] add partial files support

apolyakov:
This PR adds:

- preliminary files support: `fopen`, `fwrite`, `fread`, `fclose`
- new `resource` that has two subtypes: `sync_resource`, `async_resource`
- updated stream API that allows to read/write an exact number of bytes


runtime-light/stdlib/crypto/crypto-functions.cpp
```diff
@@ -48,11 +51,12 @@ kphp::coro::task<Optional<string>> f$openssl_random_pseudo_bytes(int64_t length)
   }
 
   auto stream{*std::move(expected_stream)};
-  if (!co_await kphp::forks::id_managed(kphp::component::query(stream, tls.view()))) [[unlikely]] {
+  kphp::stl::vector<std::byte, kphp::memory::script_allocator> response{};
```
PetrShumilov, line 54: I suppose It should be a part of instance state or some buffer's pool.

apolyakov: It can be a member of some buffer pool, but we don't have one yet

PR #1398, opened 2025-08-29
change json_decode so that it accepts string_view instead of KPHP string

apolyakov:
Current `json_decode` forces us to create unnecessary KPHP strings from time to time


runtime-common/stdlib/serialization/json-functions.cpp
```diff
@@ -557,22 +561,20 @@ bool do_json_decode(const char* s, int s_len, int& i, mixed& v, const char* json
...
-std::pair<mixed, bool> json_decode(const string& v, const char* json_obj_magic_key) noexcept {
-  mixed result;
-  int i = 0;
-  if (do_json_decode(v.c_str(), v.size(), i, result, json_obj_magic_key)) {
-    json_skip_blanks(v.c_str(), i);
-    if (i == static_cast<int>(v.size())) {
-      bool success = true;
-      return {result, success};
+std::optional<mixed> json_decode(std::string_view v, std::string_view json_obj_magic_key) noexcept {
+  mixed result{};
```
PetrShumilov, line 565: Do we need initialization here?

server/php-engine.cpp
```diff
@@ -2187,12 +2187,12 @@ int main_args_handler(int i, const char *long_option) {
       }
       std::stringstream stringstream;
       stringstream << file.rdbuf();
-      auto [config, success] = json_decode(string(stringstream.str().c_str()));
-      if (!success) {
+      auto opt_config = json_decode(stringstream.str());
+      if (!opt_config) {
```
PetrShumilov: I see we can use `has_value` for more explicitness

runtime-common/stdlib/serialization/from-json-processor.h
```diff
@@ -204,12 +204,14 @@ ClassName f$JsonEncoder$$from_json_impl(Tag /*tag*/, const string& json_string,
   auto& msg = SerializationLibContext::get().last_json_processor_error;
   msg = {};
 
-  auto [json, success] = json_decode(json_string, FromJsonVisitor<Tag>::get_json_obj_magic_key());
+  auto opt_json = json_decode({json_string.c_str(), json_string.size()}, {FromJsonVisitor<Tag>::get_json_obj_magic_key()});
 
-  if (!success) {
+  if (!opt_json) {
```
PetrShumilov: I think that better to use `json.has_value()` instead. It's more clear and explicit

runtime-common/stdlib/serialization/json-functions.cpp
apolyakov: It's equivalent to `mixed result;`

PR #1402, opened 2025-09-02
You can now write TL serializers in PHP

hrissan:
Functions to implement by PHP serializers.
```
interface RpcFunction {
...
  public function getTLFunctionMagic() : int;
  public function customStore() : RpcFunctionFetcher;
  public function customFetch() : RpcFunctionFetcher;
```

They should either store/fetch function and return instance of `RpcFunctionFetcher`, or do nothing and return `null`, in this case runtime will call C++ serializer.
```
interface RpcFunctionFetcher {
  public function typedFetch() : RpcFunctionReturnResult;
  public function typedStore($result ::: RpcFunctionReturnResult);
}
```


builtin-functions/kphp-full/_functions.txt
```diff
@@ -990,9 +991,15 @@ function rpc_tl_query_result ($query_ids ::: array) ::: mixed[][];
...
 
-/** @kphp-tl-class */
+/**
+ * @kphp-required
+ * @kphp-tl-class
+ */
 interface RpcFunction {
+  public function getTLFunctionMagic() : int;
   public function getTLFunctionName() : string;
+  public function customStore() : @tl\RpcFunctionFetcher;
```
apolyakov: Can we change the name to just `store/fetch`?

builtin-functions/kphp-full/fetcher.txt
```diff
@@ -0,0 +1,9 @@
+<?php
+
+namespace VK\TL;
+
+/** @kphp-required */
+interface RpcFunctionFetcher {
```
apolyakov: I think it's better to move it somewhere near to RPC types and functions

hrissan: it is in namespace VK/TL, so we have to put it in separate file.

PR #1403, opened 2025-09-02
Preparation step for You can now write TL serializers in PHP

hrissan:
This MR changes tl2php only (and generated code) to add stubs to  `RpcFunction` descendants (and RpcFunction.php, but only inside #ifndef KPHP).

This MR adds no changes to runtime at all.

https://github.com/VKCOM/kphp/pull/1402


common/tl-files/tests.tl
DrDet, line 1: It looks exatcly like [`common.tl`](https://github.com/VKCOM/kphp/blob/927cace039b3328f8473651a190db9852aad1654/common/tl-files/common.tl). Why do we need this?

runtime/tl/rpc_function.h
```diff
@@ -78,6 +78,31 @@ struct C$VK$TL$RpcFunctionReturnResult : abstract_refcountable_php_interface {
   virtual ~C$VK$TL$RpcFunctionReturnResult() = default;
 };
 
+struct C$VK$TL$RpcFunctionFetcher : abstract_refcountable_php_interface {
```
DrDet, line 83: let's add clarifiying comment about what's this for, like above and below

runtime-light/stdlib/rpc/rpc-tl-function.h
```diff
@@ -78,6 +78,31 @@ struct C$VK$TL$RpcFunctionReturnResult : abstract_refcountable_php_interface {
   ~C$VK$TL$RpcFunctionReturnResult() override = default;
 };
 
+struct C$VK$TL$RpcFunctionFetcher : abstract_refcountable_php_interface {
```
DrDet, line 83: and here

runtime-light/stdlib/rpc/rpc-tl-function.h
```diff
@@ -78,6 +78,31 @@ struct C$VK$TL$RpcFunctionReturnResult : abstract_refcountable_php_interface {
...
+  virtual void accept(InstanceDeepDestroyVisitor&) noexcept {}
+
+  virtual size_t virtual_builtin_sizeof() const noexcept {
+    return 0;
+  }
+  virtual C$VK$TL$RpcFunctionFetcher* virtual_builtin_clone() const noexcept {
+    return nullptr;
+  }
+
+  virtual ~C$VK$TL$RpcFunctionFetcher() = default;
```
DrDet:
```suggestion
  ~C$VK$TL$RpcFunctionFetcher() override = default;
```
like above and below

Makefile
apolyakov, line 1: I think it's better to not add such a file:

1. it doesn't take into account user's current dir
2. it uses `make` which is not actually required

hrissan: OK, let's remove. Someone will have to waste time again to recreate command to generate this.

common/tl-files/tests.tl
hrissan: It has less stuff. OK, let's remove it and replace with common.tl

runtime-light/stdlib/rpc/rpc-tl-function.h
hrissan: This is exactly like it is done in RpcFunction and RpcFunctionResult.
Either we change everywhere, or nowhere.
Let's change everywhere in the separate MR.

runtime/tl/rpc_function.h
hrissan: done

runtime-light/stdlib/rpc/rpc-tl-function.h
hrissan: fixed)

DrDet (approved):

DrDet (approved):

PR #1406, opened 2025-09-08
TL byte granularity for internal data structures

hrissan:
For RPC formats with byte granularity we change buffer from int32* into char*.

Modification to RPC code will be done later

### DIFF
--- a/runtime-light/stdlib/rpc/rpc-api.h
+++ b/runtime-light/stdlib/rpc/rpc-api.h
@@ -77,9 +77,17 @@ kphp::coro::task<class_instance<C$VK$TL$RpcResponse>> typed_rpc_tl_query_result_
 
 inline bool f$store_int(int64_t v) noexcept {
   if (tl::is_int32_overflow(v)) [[unlikely]] {
-    kphp::log::warning("integer {} overflows int32, it will be casted to {}", v, static_cast<int32_t>(v));
+    kphp::log::warning("integer {} overflows int32, it will be cast to {}", v, static_cast<int32_t>(v));
   }
   tl::i32{.value = static_cast<int32_t>(v)}.store(RpcServerInstanceState::get().tl_storer);
+  return true;
+}
+
+inline bool f$store_byte(int64_t v) noexcept {
+  if (v < 0 || v > 255) [[unlikely]] {
+    kphp::log::warning("integer {} overflows uint8, it will be cast to {}", v, static_cast<uint8_t>(v));
+  }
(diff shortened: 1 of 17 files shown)


runtime-light/stdlib/rpc/rpc-api.h
```diff
@@ -77,12 +77,20 @@ kphp::coro::task<class_instance<C$VK$TL$RpcResponse>> typed_rpc_tl_query_result_
 
 inline bool f$store_int(int64_t v) noexcept {
   if (tl::is_int32_overflow(v)) [[unlikely]] {
-    kphp::log::warning("integer {} overflows int32, it will be casted to {}", v, static_cast<int32_t>(v));
+    kphp::log::warning("integer {} overflows int32, it will be cast to {}", v, static_cast<int32_t>(v));
```
DrDet, line 80: Looks, it was grammatically correct before

runtime/rpc.cpp
```diff
@@ -130,12 +156,26 @@ void last_rpc_error_reset() {
...
 }
 
-void rpc_parse(const int32_t* new_rpc_data, int32_t new_rpc_data_len) {
+#ifdef TL1_MIRROR_TESTING
+void check_mirror() {
+  php_assert(rpc_data_len_old * 4 == rpc_data_len);
+  php_assert((rpc_data_old - rpc_data_begin_old) * 4 == rpc_data - rpc_data_begin);
+  // uncomment this heavy check if some unpredictable behavior happens again
+  // remove this function once we are sure new code works
+  // php_assert(memcmp(rpc_data_old, rpc_data, rpc_data_len) == 0);
```
DrDet, line 165: why is it commented? looks like neccessary check for mirror testing

runtime/rpc.cpp
```diff
@@ -175,96 +222,177 @@ bool f$rpc_parse(const Optional<string>& new_rpc_data) noexcept {
...
+  auto pos2 = static_cast<int>(rpc_data_old - rpc_data_begin_old) * 4;
+  php_assert(pos == pos2);
+#endif
+  return pos;
 }
 
-bool rpc_set_pos(int32_t pos) {
-  if (pos < 0 || rpc_data_begin + pos > rpc_data) {
+bool rpc_set_pos(int pos) {
+  if (pos < 0) {
```
DrDet: why don't we check right boundary overflow anymore?

hrissan: added check

DrDet (approved):

PR #1411, opened 2025-09-18
KML in k2

mkornaukhov03:


runtime-common/stdlib/kml/kphp_ml_init.cpp
```diff
@@ -0,0 +1,104 @@
...
+    return;
+  }
+
+  traverse_kml_dir(kml_models_context.kml_directory);
+
+  php_info("loaded %d kml models from %s\n", static_cast<int>(kml_models_context.loaded_models.size()), kml_models_context.kml_directory);
+}
+
+void init_kphp_ml_runtime_in_worker() {
+  KmlInferenceContext::get().mutable_buffer_in_worker = kphp::memory::platform_allocator<char>{}.allocate(KmlModelsContext::get().max_mutable_buffer_size);
```
PetrShumilov: Use std::byte instead

runtime-light/stdlib/kml/kml-instance-state.cpp
```diff
@@ -0,0 +1,18 @@
...
+KmlInstanceState& KmlInstanceState::get() noexcept {
+  return InstanceState::get().kml_instance_state;
+}
+
+KmlInferenceContext& KmlInferenceContext::get() noexcept {
+  return KmlInstanceState::get().kml_inference_context;
+}
+
+void enable_malloc_in_inference() {
+  InstanceState::get().instance_allocator_state.enable_libc_alloc();
```
apolyakov: You should not call `enable_libc_alloc` mannualy. There is already a API for that: `kphp::memory::libc_alloc_guard`

runtime-common/stdlib/kml/kphp_ml_interface.cpp
```diff
@@ -159,9 +161,10 @@ Optional<string> f$kml_get_custom_property(const string& model_name, const strin
     return {};
   }
 
-  auto guard = dl::CriticalSectionGuard();
-
+  enable_malloc_in_inference();
   auto inner_result = p_kml->get_custom_property(std::string(property_name.c_str()));
```
apolyakov: Is there really a need for `std::string`?

runtime-common/stdlib/kml/kml-files-reader.cpp
```diff
@@ -23,87 +27,75 @@
 
 namespace {
 
-// This class is used to free memory in the end of the program
-// It is utilized to allocate memory for 'offset_in_vec' and 'reindex_map_int2int'
+// Not really efficient, but it's OK as reading ML models once at start
 template<typename T>
-class GlobalLifetimeResource {
-  std::vector<T*> managed_resources;
+using Result = std::variant<T, std::string>;
```
PetrShumilov: General question: why do we use `std::string` instead of our strings?

runtime-light/state/component-state.h
```diff
@@ -25,6 +29,14 @@ struct ComponentState final : private vk::not_copyable {
...
     parse_env();
     parse_args();
+
+    auto kml_dir = ini_opts[string("kml_dir")];
+    char* buffer = static_cast<char*>(k2::alloc(kml_dir.size() + 1));
+    strncpy(buffer, kml_dir.c_str(), kml_dir.size() + 1); // Copy null terminator, too
+    kml_component_state.kml_models_context.kml_directory = buffer;
+    component_allocator_state.enable_libc_alloc();
+    init_kphp_ml_runtime_in_master();
+    component_allocator_state.disable_libc_alloc();
```
PetrShumilov: Does it be expressed more elegant? For example, in separate function `kml_init`

runtime-light/stdlib/kml/kml-instance-state.cpp
mkornaukhov: It's used in `runtime-common` code, there is no API for that as `kphp::memory::libc_alloc_guard` is defined only in `runtime-light`.

apolyakov: Let's move `kphp::memory::libc_alloc_guard` to `runtime-common` then

mkornaukhov: Done

runtime-common/stdlib/kml/kphp_ml_init.cpp
mkornaukhov: Fixed

runtime-light/state/component-state.h
mkornaukhov: I've used guard and renamed initialization function

runtime-common/stdlib/kml/kml-files-reader.cpp
mkornaukhov: There is kphp-runtime-agnostic code in kml reader and MLModel. This is done by design to keep it more reusable.

runtime-common/stdlib/kml/kphp_ml_interface.cpp
mkornaukhov: Yes, it's required for lookup in hash table

PR #1412, opened 2025-09-18
rebalance available CPUs to build and run parallel tests

Danil42Russia:
Roughly calculated that giving `95%` to run tests and `27%` to build is a rational value. Assuming the build server has `32` cores, we will give `30` cores to run the tests, and `8` cores to build.

### DIFF
--- a/tests/kphp_ci_pipeline_runner.py
+++ b/tests/kphp_ci_pipeline_runner.py
@@ -235,8 +235,8 @@ def parse_args():
 
 def _calculate_pytest_jobs_count(default_percent: int = 95) -> int:
     """
-    What percentage of the total number of cores do we use to run parallel tests.
-    If you set a large value, the cores may start to idle.
+    Determine the percentage of the total number of CPU cores to use for running parallel Python tests.
+    It is recommended not to allocate more than 95% of available cores to avoid idle time.
     """
 
     cpu_count = multiprocessing.cpu_count()
--- a/tests/python/lib/kphp_builder.py
+++ b/tests/python/lib/kphp_builder.py
@@ -121,8 +121,8 @@ class KphpBuilder:
 
     def _calculate_kphp_jobs_count(self, default_percent: int = 25) -> int:
         """
-        What percentage of the total number of cores do we use to run parallel kphp compilation.
-        You don't want to give more than 1/4 of the number of cores, or you will starvation.
+        Determine the percentage of the total number of CPU cores to use for running parallel KPHP compilation.
+        It is recommended not to allocate more than 25% of the available cores to avoid resource starvation.
         """
 
         cpu_count = multiprocessing.cpu_count()


tests/kphp_ci_pipeline_runner.py
```diff
@@ -232,6 +233,14 @@ def parse_args():
 
     return parser.parse_args()
 
+def _calculate_pytest_jobs_count(percent: int = 5) -> int:
```
DrDet: This naming looks confusing. As far as I can see, `percent` defines part of cores which will NOT be used for pytest jobs.

Danil42Russia: Yes, I agree, I made a reverse percentage by mistake :(

tests/kphp_ci_pipeline_runner.py
```diff
@@ -232,6 +233,16 @@ def parse_args():
...
+def _calculate_pytest_jobs_count(default_percent: int = 95) -> int:
+    """
+    What percentage of the total number of cores do we use to run parallel tests.
+    If you set a large value, the cores may start to idle.
+    """
+
+    cpu_count = multiprocessing.cpu_count()
+    percent = int(os.getenv("FORCE_PYTEST_JOBS_PERCENT", default_percent))
+
+    return math.floor(cpu_count * percent / 100)
```
DrDet, line 245: what happens if it's `== 0`?

tests/python/lib/kphp_builder.py
```diff
@@ -117,6 +119,17 @@ def _move_sanitizer_logs_to_artifacts(self, sanitizer_glob_mask, proc, sanitizer
                 return self._move_to_artifacts(sanitizer_log_name, proc.returncode, file=sanitizer_log)
         return None
 
+    def _calculate_kphp_jobs_count(self, default_percent: int = 25) -> int:
+        """
+        What percentage of the total number of cores do we use to run parallel kphp compilation.
+        You don't want to give more than 1/4 of the number of cores, or you will starvation.
```
apolyakov:
```
Determine the percentage of the total number of CPU cores to use for running parallel KPHP compilation. It is recommended not to allocate more than 25% of the available cores to avoid resource starvation.
```

tests/kphp_ci_pipeline_runner.py
```diff
@@ -232,6 +233,16 @@ def parse_args():
 
     return parser.parse_args()
 
+def _calculate_pytest_jobs_count(default_percent: int = 95) -> int:
+    """
+    What percentage of the total number of cores do we use to run parallel tests.
```
apolyakov: Same as above

tests/kphp_ci_pipeline_runner.py
Danil42Russia: I haven't checked, but most likely nothing will be collected or there will be an error. But whatever happens is unlikely

PR #1414, opened 2025-09-24
use local tcp server in tests

Shamzik:
The goal is to enable TCP tests.

What's been done:
- Added a local TCP server;
- Fixed the fgets implementation.

### DIFF
--- a/runtime/tcp.cpp
+++ b/runtime/tcp.cpp
@@ -65,7 +65,8 @@ Optional<int64_t> connect_to_address(const string& host, int64_t port, double en
     return {};
   }
 
-#ifndef NDEBUG
+  // The following code block prints the host resolving result for debugging.
+#if 0
   string ai_host(NI_MAXHOST, false);
   string ai_port(NI_MAXSERV, false);
   for (auto ai = result; ai != nullptr; ai = ai->ai_next) {


tests/python/lib/kphp_run_once.py
```diff
@@ -59,6 +82,7 @@ def _get_extensions(self):
 
         return extensions
 
+    @_with_tcp
```
PetrShumilov: I suppose that decorators isn't appropriate choice. Make it simple, move starting and shut-downing routines directly inside  `KphpRunOnce` and `PhpRunOnce`

PetrShumilov (changes requested):

tests/python/lib/kphp_run_once.py
Shamzik: I rewrote the decorator as a context manager and moved it to tests/kphp_tester.py to avoid duplicate code.

See the latest changes

runtime/tcp.cpp
```diff
@@ -63,6 +64,22 @@ Optional<int64_t> connect_to_address(const string& host, int64_t port, double en
     faulter(-3, string("Can't resolve host \"%s\""), host);
     return {};
   }
+
+#ifndef NDEBUG
```
PetrShumilov: 1) Make a separate macros 
2) Add comments here for clarify why we need it

PetrShumilov (changes requested):

runtime/tcp.cpp
Shamzik: I think we don't need unused macros. See fix

PetrShumilov (approved):
LGTM

PR #1418, opened 2025-10-03
[k2] add await set primitive

astrophysik:


runtime-light/coroutine/await-set.h
```diff
@@ -0,0 +1,62 @@
...
+#include "runtime-light/coroutine/async-stack.h"
+#include "runtime-light/coroutine/concepts.h"
+#include "runtime-light/coroutine/coroutine-state.h"
+#include "runtime-light/coroutine/detail/await-set.h"
+#include "runtime-light/coroutine/type-traits.h"
+
+namespace kphp::coro {
+
+template<typename return_type>
+class await_set {
```
apolyakov, line 22: This class is missing an explicit destructor.

runtime-light/coroutine/await-set.h
```diff
@@ -0,0 +1,62 @@
...
+
+public:
+  template<typename T>
+  friend class kphp::coro::detail::await_set::await_broker;
+
+  explicit await_set(await_set_waiting_policy waiting_policy = await_set_waiting_policy::suspend_on_empty) noexcept
+      : m_await_broker(*this, waiting_policy) {}
+
+  await_set(const await_set&) = delete;
+  await_set(await_set&&) = delete;
```
apolyakov: It seems reasonable for `await_set` to be movable

runtime-light/coroutine/await-set.h
```diff
@@ -0,0 +1,62 @@
...
+template<typename return_type>
+class await_set {
+  kphp::stl::list<detail::await_set::await_set_task<return_type>, kphp::memory::script_allocator> m_tasks_storage{};
+  detail::await_set::await_broker<return_type> m_await_broker;
+
+public:
+  template<typename T>
+  friend class kphp::coro::detail::await_set::await_broker;
+
+  explicit await_set(await_set_waiting_policy waiting_policy = await_set_waiting_policy::suspend_on_empty) noexcept
```
apolyakov: 1. Why is it declared in `detail/` if it's a part of public API? Also, wouldn't it be better to use it like `kphp::coro::await_set::waiting_policy`? It's only applicable to `await_set`, so I'm not sure it's the best choice to declare in `kphp::coro` namespace
2. I think `resume_on_empty` should be default. It's better to instantly return `std::nulllopt` than accidentally await forever

runtime-light/coroutine/await-set.h
```diff
@@ -0,0 +1,62 @@
...
+namespace kphp::coro {
+
+template<typename return_type>
+class await_set {
+  kphp::stl::list<detail::await_set::await_set_task<return_type>, kphp::memory::script_allocator> m_tasks_storage{};
+  detail::await_set::await_broker<return_type> m_await_broker;
+
+public:
+  template<typename T>
+  friend class kphp::coro::detail::await_set::await_broker;
```
apolyakov, line 28: There is no need to declare friends public

runtime-light/coroutine/await-set.h
```diff
@@ -0,0 +1,62 @@
+// Compiler for PHP (aka KPHP)
+// Copyright (c) 2025 LLC «V Kontakte»
+// Distributed under the GPL v3 License, see LICENSE.notice.txt
+
+#pragma once
+
```
apolyakov, line 6: At least `<utility>` and `<cstddef>` headers are missing. Please, check for others too

runtime-light/coroutine/await-set.h
```diff
@@ -0,0 +1,62 @@
...
+  await_set(await_set&&) = delete;
+
+  await_set& operator=(const await_set&) = delete;
+  await_set& operator=(await_set&& other) = delete;
+
+  template<typename awaitable_type>
+  requires kphp::coro::concepts::awaitable<awaitable_type> && std::is_same_v<typename awaitable_traits<awaitable_type>::awaiter_return_type, return_type>
+  void push(awaitable_type awaitable) noexcept {
+    m_tasks_storage.push_front(detail::await_set::make_await_set_task(std::move(awaitable)));
+    auto task_iterator{m_tasks_storage.begin()};
```
apolyakov: nit: `m_tasks_storage.front()` instead of `begin()`? It corresponds to previous `push_front`

runtime-light/coroutine/await-set.h
```diff
@@ -0,0 +1,62 @@
...
+    auto& coroutine_stack_root{CoroutineInstanceState::get().coroutine_stack_root};
```
apolyakov: I don't insist on that, but I think it's ok to have the `m_coroutine_stack_root` as a `await_set`'s member

runtime-light/coroutine/detail/await-set.h
```diff
@@ -0,0 +1,365 @@
...
+    auto* ready_task{m_ready_tasks};
+    m_ready_tasks = m_ready_tasks->m_next;
+    auto task_iterator{ready_task->m_storage_location};
+    auto result{task_iterator->result()};
+    m_await_set.m_tasks_storage.erase(task_iterator);
+    return result_t{std::move(result)};
+  }
+
+  void detach_awaiters() noexcept {
+    while (m_waiters != nullptr) {
```
apolyakov: It may be dangerous. Imagine that the first awaiter you've detached pushes a new task to the await set?

runtime-light/coroutine/detail/await-set.h
```diff
@@ -0,0 +1,365 @@
...
+#include "runtime-light/coroutine/type-traits.h"
+#include "runtime-light/coroutine/void-value.h"
+#include "runtime-light/stdlib/diagnostics/logs.h"
+
+namespace kphp::coro {
+
+enum class await_set_waiting_policy : uint8_t { suspend_on_empty, resume_on_empty };
+
+template<typename return_type>
+class await_set;
```
apolyakov, line 22: missing `NL`

runtime-light/coroutine/detail/await-set.h
```diff
@@ -0,0 +1,365 @@
...
+} // namespace kphp::coro
+
+namespace kphp::coro::detail::await_set {
+
+struct await_set_waiter {
+  await_set_waiter* m_next{};
+  await_set_waiter* m_prev{};
+  std::coroutine_handle<> m_continuation{};
```
apolyakov: nit: redundant init

runtime-light/coroutine/detail/await-set.h
```diff
@@ -0,0 +1,365 @@
...
+};
+
+template<typename return_type>
+class await_set_task;
+
+template<typename return_type>
+struct await_set_ready_task_element {
+  await_set_ready_task_element* m_next{};
+  std::list<await_set_task<return_type>>::iterator m_storage_location{};
```
apolyakov: Why `std::list`?

runtime-light/coroutine/detail/await-set.h
```diff
@@ -0,0 +1,365 @@
...
+#include "runtime-common/core/allocator/script-malloc-interface.h"
+#include "runtime-light/coroutine/async-stack.h"
+#include "runtime-light/coroutine/concepts.h"
+#include "runtime-light/coroutine/type-traits.h"
+#include "runtime-light/coroutine/void-value.h"
+#include "runtime-light/stdlib/diagnostics/logs.h"
+
+namespace kphp::coro {
+
+enum class await_set_waiting_policy : uint8_t { suspend_on_empty, resume_on_empty };
```
apolyakov: What do you think about making the policy a part of `await_set`'s type? Having it that way may lead to unexpected behavior (not a C++ UB). For example:
```
void foo(kphp::coro::await_set as) {
  co_await as.next(); // It's unknown that will happen since we don't know the policy
}

```

runtime-light/coroutine/detail/await-set.h
```diff
@@ -0,0 +1,365 @@
...
+        m_await_set(await_set) {}
+
+  await_broker(const await_broker&) = delete;
+  await_broker(await_broker&&) = delete;
+
+  await_broker& operator=(const await_broker&) = delete;
+  await_broker& operator=(await_broker&& other) = delete;
+
+  void push_ready_task(await_set_ready_task_element<return_type>& ready_task) noexcept {
+    ready_task.m_next = m_ready_tasks;
```
apolyakov: nit: `ready_task.m_next = std::exchange(m_ready_tasks, std::addressof(ready_task))` to explicitly point that this is an "atomic" action

runtime-light/coroutine/detail/await-set.h
```diff
@@ -0,0 +1,365 @@
...
+
+private:
+  void resume_first_awaiter() noexcept {
+    auto* waiter{m_waiters};
+    m_waiters = waiter->m_next;
+    if (m_waiters != nullptr) {
+      m_waiters->m_prev = nullptr;
+    }
+
+    waiter->m_continuation.resume();
```
apolyakov: Let's add `kphp::log::assertion(waiter != nullptr)`

runtime-light/coroutine/detail/await-set.h
```diff
@@ -0,0 +1,365 @@
...
+  public:
+    explicit awaiter(await_broker<return_type>& await_broker) noexcept
+        : m_await_broker(await_broker) {}
+
+    awaiter(awaiter&& other) noexcept = delete;
+    awaiter(const awaiter& other) = delete;
+    awaiter& operator=(const awaiter& other) = delete;
+    awaiter& operator=(awaiter&& other) = delete;
+
+    bool await_ready() noexcept {
```
apolyakov: `constexpr bool await_ready() const noexcept`

runtime-light/coroutine/detail/await-set.h
```diff
@@ -0,0 +1,365 @@
...
+
+    std::optional<result_type> await_resume() noexcept {
+      // restore caller async stack frame
+      kphp::log::assertion(caller_frame != nullptr);
+      auto* async_stack_root{caller_frame->async_stack_root};
+      kphp::log::assertion(async_stack_root != nullptr);
+      async_stack_root->top_async_stack_frame = caller_frame;
+
+      m_suspended = false;
+      return std::move(m_await_broker.try_get_result());
```
apolyakov: No need for move here as `try_get_result()` is already rvalue

runtime-light/coroutine/detail/await-set.h
```diff
@@ -0,0 +1,365 @@
...
+      : m_await_broker(await_broker) {}
+
+  await_set_awaitable(const await_set_awaitable&) = delete;
+  await_set_awaitable(await_set_awaitable&& other) noexcept
+      : m_await_broker(other.m_await_broker) {}
+
+  await_set_awaitable& operator=(const await_set_awaitable&) = delete;
+  await_set_awaitable& operator=(await_set_awaitable&& other) = delete;
+
+  auto operator co_await() const noexcept {
```
apolyakov: What do you think about defining it for `&&` only?

runtime-light/coroutine/await-set.h
astrophysik: Okay

runtime-light/coroutine/await-set.h
astrophysik: No, it cannot be. The main issue is that await_broker is stored inside await_set. When it is moved, the pointers to await_broker need to be updated in all running await_set_task and in all await_set_awaitable that have already been returned through next().

runtime-light/coroutine/await-set.h
astrophysik: Okay

runtime-light/coroutine/await-set.h
astrophysik: fixed

runtime-light/coroutine/await-set.h
astrophysik: I use begin here to get an iterator to the element and later remove it in O(1) time. Rewrite it in another way to avoid push_front

runtime-light/coroutine/detail/await-set.h
astrophysik: fix

runtime-light/coroutine/detail/await-set.h
astrophysik: fix

runtime-light/coroutine/detail/await-set.h
astrophysik: fix

runtime-light/coroutine/detail/await-set.h
astrophysik: Yeah, good point. I rework this with a change to the shutdown semantics

runtime-light/coroutine/detail/await-set.h
astrophysik: Good point, now await set has type `await_set<return_type, await_set_waiting_policy>`

runtime-light/coroutine/detail/await-set.h
astrophysik: This breaks the match for awaitable concept

runtime-light/coroutine/await-set.h
```diff
@@ -0,0 +1,81 @@
+// Compiler for PHP (aka KPHP)
+// Copyright (c) 2025 LLC «V Kontakte»
+// Distributed under the GPL v3 License, see LICENSE.notice.txt
+
+#pragma once
+
+#include <cstddef>
+#include <cstdint>
```
apolyakov: unnecessary include

runtime-light/coroutine/await-set.h
```diff
@@ -0,0 +1,81 @@
...
+  friend class kphp::coro::detail::await_set::await_broker;
+
+  void abort_all() noexcept {
+    m_tasks_storage.clear();
+    m_await_broker.release();
+  }
+
+public:
+  explicit await_set() noexcept
+      : m_await_broker(new detail::await_set::await_broker<return_type>(*this)),
```
apolyakov: nit: it's better to use `make_unique` or even `await_broker::create -> std::optional<await_broker>`

runtime-light/coroutine/await-set.h
```diff
@@ -0,0 +1,81 @@
...
+        m_coroutine_stack_root(CoroutineInstanceState::get().coroutine_stack_root) {}
+
+  await_set(await_set&& other) noexcept
+      : m_tasks_storage(std::move(other.m_tasks_storage)),
+        m_await_broker(std::move(other.m_await_broker)),
+        m_coroutine_stack_root(other.m_coroutine_stack_root) {}
+
+  await_set& operator=(await_set&& other) noexcept {
+    abort_all();
+    m_tasks_storage = std::move(other.m_tasks_storage);
```
apolyakov: what if `this == std::addressof(other)`

runtime-light/coroutine/detail/await-set.h
```diff
@@ -0,0 +1,377 @@
+// Compiler for PHP (aka KPHP)
+// Copyright (c) 2025 LLC «V Kontakte»
+// Distributed under the GPL v3 License, see LICENSE.notice.txt
+
+#pragma once
+
+#include <concepts>
+#include <coroutine>
+#include <cstdint>
```
apolyakov: unnecessary include

runtime-light/coroutine/detail/await-set.h
```diff
@@ -0,0 +1,377 @@
...
+
+template<typename return_type>
+struct await_set_ready_task_element {
+  await_set_ready_task_element* m_next{};
+  kphp::stl::list<await_set_task<return_type>, kphp::memory::script_allocator>::iterator m_storage_location{};
+};
+
+template<typename return_type>
+class await_broker {
+  using await_set_t = kphp::coro::await_set<return_type>;
```
apolyakov: Is it really needed?

runtime-light/coroutine/detail/await-set.h
```diff
@@ -0,0 +1,377 @@
...
+  await_set_t& m_await_set;
+  await_set_waiter* m_waiters{};
```
apolyakov: `m_awaiters`?

runtime-light/coroutine/detail/await-set.h
```diff
@@ -0,0 +1,377 @@
...
+  auto final_suspend() const noexcept {
+    struct completion_notifier {
+
+      constexpr auto await_ready() const noexcept {
+        return false;
+      }
+
+      auto await_suspend(std::coroutine_handle<promise_type> coroutine) noexcept -> void {
+        auto& promise{coroutine.promise()};
+        auto await_broker{promise.m_await_broker};
```
apolyakov: let's rename to `opt_await_broker`

runtime-light/coroutine/detail/await-set.h
```diff
@@ -0,0 +1,377 @@
...
+    std::optional<T> m_result;
+
+  public:
+    auto yield_value(return_type&& return_value) noexcept {
+      m_result = std::move(return_value);
+      return await_set_task_promise_common::final_suspend();
+    }
+
+    T result() noexcept {
+      return std::move(*m_result);
```
apolyakov, line 248: let's add a check that `m_result` is present

PR #1466, opened 2025-11-26
[k2] Add support for multiple sync/async stack switches in backtrace algorithm

denisichh:
**General**

This PR improves collecting of async stack traces in runtime-light

**Problem**

The current collecting backtrace doesn't handle jumps from an async stack to a sync stack and back, as only 1 async stack root is supported

**Solution**

Added creating of new async stack roots and added forwarding of pointers to the next async stack root


runtime-light/coroutine/async-stack.h
```diff
@@ -4,9 +4,6 @@
 
 #pragma once
 
-#include <coroutine>
-#include <utility>
-
 /**
  * This header defines the data structures used to represent a coroutine asynchronous stack.
```
astrophysik: Let's update this section for the new algorithm

runtime-light/coroutine/async-stack-methods.h
```diff
@@ -0,0 +1,36 @@
...
+#include "runtime-light/coroutine/async-stack.h"
+
+#include <coroutine>
+#include <utility>
+
+#include "runtime-light/coroutine/coroutine-state.h"
+
+namespace kphp::coro {
+
+using ::CoroutineInstanceState;
```
astrophysik: Why do we need this using? Can we remove it?

runtime-light/coroutine/async-stack-methods.h
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * The `resume` function is responsible for storing the current synchronous stack frame
+ * in async_stack_root::stop_sync_frame before resuming the coroutine and also creates a new async_stack_root. This allows
+ * capturing one of the stack frames in the synchronous stack trace and jump between parts of the async_stack.
+ */
+[[clang::noinline]] inline void resume(std::coroutine_handle<> handle, async_stack_root* stack_root = nullptr) noexcept {
```
astrophysik: Let’s rename `stack_root` to `current_stack_root` or `caller_stack_root`. The current name doesn’t make it clear which stack_root it refers to.

runtime-light/coroutine/async-stack-methods.h
```diff
@@ -0,0 +1,36 @@
...
+  if (stack_root == nullptr) {
```
astrophysik: I think it’s better to remove this. Let the code that uses `resume` pass the required stack root itself.

silenthill07 (changes requested):

scutuatua-crypto (approved):

PR #1601, opened 2026-04-07
[k2] add temporary stats for confdata and instance cache calls

apolyakov:


runtime-light/state/instance-state.cpp
```diff
@@ -244,4 +244,8 @@ kphp::coro::task<> InstanceState::run_instance_epilogue() noexcept {
   while (!ignore_answer_request_await_set.empty()) {
     co_await ignore_answer_request_await_set.next();
   }
+
+  // TEMPORARY: log instance-cache and confdata timings
+  kphp::log::info("instance stats (approximately): confdata -> {}ms, instance cache -> {}ms", confdata_instance_state.time_ns >> 20,
+                  instance_cache_instance_state.time_ns >> 20);
```
DrDet: let's add ratio to total request time?

PR #1607, opened 2026-04-10
[k2] implement random_int

apolyakov:


Shamzik (dismissed):
What about adding some tests?

apolyakov:
> What about adding some tests?

There's already `dl/386_random_int.php`

Shamzik:
> > What about adding some tests?
> 
> There's already `dl/386_random_int.php`

Can we remove the "php8" tag?

Shamzik (dismissed):
Love it!

Shamzik (approved):

PR #1612, opened 2026-04-21
[k2] fix reference counter function overloads

astrophysik:


runtime-light/core/reference-counter/reference-counter-functions.h
```diff
@@ -7,11 +7,23 @@
 #include <utility>
 
 #include "common/php-functions.h"
+#include "runtime-common/core/core-types/decl/optional.h"
```
apolyakov: Redundant include here. `Optional` is included from `runtime-core.h`

apolyakov (approved):

PR #1610, opened 2026-04-16
[k2] make init instance memory size adjustable

PetrShumilov:
Key changes:
* Set initial instance memory size to 64Mb (instead of 128 Mb previously) 
* Set min instance extra memory size to 64Mb (instead of 1Mb previously) 
* Made initial instance memory size adjustable via `initial-instance-memory-size` component argument
* Made  min instance extra memory size adjustable via `min-instance-extra-memory-size` component argument

PR #1615, opened 2026-04-22
[k2] undo revert of  "make init instance memory size adjustable" and add small fixes

PetrShumilov:
* Undo revert #1614 and reapply changes of #1610
* Decreased the minimum amount of extra memory requested from the platform. 64 Mb -> 1Mb
* Fix small remarks

### DIFF
--- a/runtime-common/core/allocator/runtime-allocator.h
+++ b/runtime-common/core/allocator/runtime-allocator.h
@@ -13,7 +13,7 @@ struct RuntimeAllocator final : vk::not_copyable {
   static RuntimeAllocator& get() noexcept;
 
   RuntimeAllocator() = default;
-  RuntimeAllocator(size_t script_mem_size, size_t oom_handling_mem_size);
+  RuntimeAllocator(size_t script_mem_size, size_t min_extra_mem_size, size_t oom_handling_mem_size);
 
   void init(void* buffer, size_t script_mem_size, size_t oom_handling_mem_size);
   void free();
@@ -28,5 +28,12 @@ struct RuntimeAllocator final : vk::not_copyable {
   void* realloc_global_memory(void* mem, size_t new_size, size_t old_size) noexcept;
   void free_global_memory(void* mem, size_t size) noexcept;
 
+private:
+  void request_extra_memory(size_t requested_size, size_t min_extra_mem_size) noexcept;
+
+public:
   memory_resource::unsynchronized_pool_resource memory_resource;
+
+private:
+  size_t min_extra_mem_size{0};
 };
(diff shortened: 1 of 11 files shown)


runtime-common/core/allocator/runtime-allocator.h
```diff
@@ -28,5 +28,12 @@ struct RuntimeAllocator final : vk::not_copyable {
...
   void free_global_memory(void* mem, size_t size) noexcept;
 
+private:
+  void request_extra_memory(size_t requested_size) noexcept;
+
+public:
   memory_resource::unsynchronized_pool_resource memory_resource;
+
+private:
+  size_t min_extra_mem_size{0};
```
apolyakov: Let's rename this to `m_min_extra_size` so that it doesn't interfere with method parameters

runtime-light/allocator/allocator-state.h
```diff
@@ -11,6 +11,8 @@
 #include "runtime-common/core/allocator/runtime-allocator.h"
 #include "runtime-light/stdlib/diagnostics/logs.h"
 
+inline constexpr auto DEFAULT_MIN_EXTRA_MEMORY_POOL_SIZE{static_cast<size_t>(1 * 1024U * 1024U)}; // 1mb
```
apolyakov: nit: `1 MiB`

runtime-light/allocator/runtime-light-allocator.cpp
```diff
@@ -113,3 +100,19 @@ void* RuntimeAllocator::realloc_global_memory(void* old_mem, size_t new_size, si
 void RuntimeAllocator::free_global_memory(void* mem, size_t /*unused*/) noexcept {
   k2::free(mem);
 }
+
+void RuntimeAllocator::request_extra_memory(size_t requested_size) noexcept {
+  // Extra mem size have to be greater than max chunk block
+  const auto min_size{std::max(min_extra_mem_size, memory_resource::unsynchronized_pool_resource::MAX_CHUNK_BLOCK_SIZE)};
+
+  // If the requested size is greater than or equal to half of min_size, it’s more efficient to allocate a multiple of the requested size.
+  size_t extra_mem_size{std::max(min_size, 2 * requested_size)};
```
apolyakov, line 109: Isn't it better to allocate closest power of 2 that is bigger than `requested_size`?

apolyakov (commented):
Let's also rename this PR since it doesn't revert a commit actually. It also adds some unique changes

runtime-light/allocator/runtime-light-allocator.cpp
PetrShumilov: In separate PR

apolyakov (approved):

PR #1616, opened 2026-04-22
add parent pid and registers collecting

denisichh:


common/server/crash-dump.h
```diff
@@ -3,5 +3,18 @@
...
 
 #pragma once
+#include <cstddef>
+#include <cstdint>
 
+struct crash_dump_buffer_t {
+  char scratchpad[1024]{};
+  size_t position{0};
+
+  void reset() {
```
astrophysik: NIT: not very useful addition as for me since since working with the class still occurs through explicit access to the fields

server/json-logger.cpp
```diff
@@ -348,12 +351,30 @@ void JsonLogger::write_general_info(JsonBuffer *json_out_it, int type, int64_t c
     json_out_it->append_key("logname_id").append_integer(logname_id);
   }
   json_out_it->append_key("pid").append_integer(pid);
+  json_out_it->append_key("ppid").append_integer(ppid);
   json_out_it->append_key("cluster").append_string(vk::singleton<ServerConfig>::get().get_cluster_name());
   json_out_it->append_raw(uncaught ? R"json("uncaught":true)json" : R"json("uncaught":false)json");
   json_out_it->finish<'}'>();
 
   if (extra_info_available_) {
```
astrophysik, line 359: Why not write in tags where the internal data is already written?

server/json-logger.cpp
```diff
@@ -348,12 +351,31 @@ void JsonLogger::write_general_info(JsonBuffer *json_out_it, int type, int64_t c
...
-    json_out_it->append_key("extra_info").start<'{'>().append_raw(extra_info_).finish<'}'>();
+    json_out_it->append_key("extra_info").start<'{'>().append_raw(extra_info_);
+    if (ucontext != nullptr) {
```
astrophysik, line 361: Let's do it like
```c++
json_out_it->append_key("registers").append_string(format_registers(ucontext));
json_out_it->append_key("registers").append_string(format_cr2(ucontext));
```
So that all specific logic will be moved in formatted functions

server/json-logger.cpp
```diff
@@ -348,12 +351,31 @@ void JsonLogger::write_general_info(JsonBuffer *json_out_it, int type, int64_t c
...
+      #define LITERAL_WITH_LENGTH(literal) literal, sizeof(literal) - 1
```
astrophysik: Why you choose such define instead of `static constexpr std::string_view{"0x"}`?

server/json-logger.cpp
```diff
@@ -348,12 +351,31 @@ void JsonLogger::write_general_info(JsonBuffer *json_out_it, int type, int64_t c
...
+      const auto* ucp = static_cast<ucontext_t*>(ucontext);
+      crash_dump_buffer_t buffer{};
+
+#if defined(__x86_64__) && !defined(__APPLE__)
+      crash_dump_write_reg(LITERAL_WITH_LENGTH("0x"), ucp->uc_mcontext.gregs[REG_CR2], std::addressof(buffer));
+      assert(buffer.position < sizeof(buffer.scratchpad));
```
astrophysik: I don't think these assertions are needed here. `crash_dump_write` functions should do same checks, not the caller

server/json-logger.cpp
denisichh: done

server/json-logger.cpp
denisichh: done

astrophysik (approved):
LGTM

PR #1625, opened 2026-05-02
support op_add in CollectConstVarsPass

apolyakov:
This PR extends the `CollectConstVarsPass` compiler optimization to support the `+` operator (array union/addition) for constant expressions. This moves evaluation of constant `arr + arr` expressions from request execution time to kPHP initialization stage.

Implementation Details:
The optimization is currently limited to `arr + arr` case only. While the optimization would be safe for some other cases as well (the runtime handles type coercion), `K2` runtime's implementation of string addition requires `RuntimeContext` to be initialized (which is not initialized at the time of constant initialization stage). To ensure compatibility with both legacy and `K2` runtimes, we restrict this optimization to array addition only.

Example:
Before this change, an expression like `['a' => 1] + ['b' => 2]` would be evaluated on every request. After this change, it's evaluated once during initialization and stored as a constant variable.

### DIFF
--- a/compiler/pipes/collect-const-vars.cpp
+++ b/compiler/pipes/collect-const-vars.cpp
@@ -37,8 +37,6 @@ struct VertexVisitor {
       return Derived::on_string_build(v.as<op_string_build>());
     case op_define_val:
       return Derived::on_define_val(v.as<op_define_val>());
-    case op_var:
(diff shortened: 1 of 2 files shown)


PetrShumilov (approved):
OK

PR #1626, opened 2026-05-05
[k2] update k2-header.h

Shamzik:
This PR updates k2 header


apolyakov (changes requested):
Looks like you've forgotten to update `compiler/code-gen/files/init-scripts.cpp:300`

apolyakov (approved):

PR #1629, opened 2026-05-18
[k2] use initial-exec TLS model for performance critical thread locals

apolyakov:
The `initial-exec` TLS model provides faster access to thread-local variables by using a more efficient addressing mode that avoids dynamic TLS resolution at runtime. Since these state pointers are accessed frequently in hot paths, caching them in thread-local storage with `initial-exec` model improves performance.

The state pointers are now set once during lifecycle transitions (image/component/instance creation and initialization, and at poll entry) and then accessed directly from TLS rather than calling through the C API layer each time.

### Note

The `initial-exec` TLS model is generally unsafe for dynamically loaded libraries (it requires TLS to be allocated at program startup). This is acceptable for K2 because the runtime is loaded in a controlled environment where this constraint is satisfied. In case of any problems we can try `-mtls-dialect=gnu2` which is safe but a bit slower than `initial-exec` TLS model.

### DIFF
--- a/runtime-light/k2-platform/k2-api.h
+++ b/runtime-light/k2-platform/k2-api.h
@@ -29,11 +29,17 @@
 
 namespace k2 {
 
-namespace k2_impl_ {
+namespace details {
 
 inline constexpr size_t DEFAULT_MEMORY_ALIGN = 16;
 
-} // namespace k2_impl_
+#define TLS_INITIAL_EXEC __attribute__((tls_model("initial-exec")))
+
+TLS_INITIAL_EXEC inline thread_local const ImageState* image_state_ptr{};         // NOLINT
+TLS_INITIAL_EXEC inline thread_local const ComponentState* component_state_ptr{}; // NOLINT
+TLS_INITIAL_EXEC inline thread_local InstanceState* instance_state_ptr{};         // NOLINT
+
+} // namespace details
 
 inline constexpr int32_t errno_ok = 0;
 inline constexpr int32_t errno_e2big = E2BIG;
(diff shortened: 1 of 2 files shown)


runtime-light/k2-platform/k2-api.h
```diff
@@ -29,11 +29,17 @@
...
 
 inline constexpr size_t DEFAULT_MEMORY_ALIGN = 16;
 
-} // namespace k2_impl_
+#define TLS_INITIAL_EXEC __attribute__((tls_model("initial-exec")))
+
+TLS_INITIAL_EXEC inline thread_local const ImageState* image_state_ptr{};         // NOLINT
+TLS_INITIAL_EXEC inline thread_local const ComponentState* component_state_ptr{}; // NOLINT
+TLS_INITIAL_EXEC inline thread_local InstanceState* instance_state_ptr{};         // NOLINT
+
```
astrophysik, line 41: Maybe add `#undef` here?

runtime-light/runtime-light.cpp
```diff
@@ -15,50 +16,54 @@
 #define VISIBILITY_DEFAULT __attribute__((visibility("default")))
 
 VISIBILITY_DEFAULT ImageState* k2_create_image() {
-  kphp::log::debug("start image state creation, requested {} bytes", sizeof(ImageState));
```
astrophysik, line 18: Shouldn't we explicit zeroing undefined thread locals? Can they contain values of previous instance?

astrophysik (approved):

PR #1633, opened 2026-05-25
[k2] add `stream_set_blocking`

denisichh:


runtime-light/streams/stream.h
```diff
@@ -27,12 +27,16 @@
...
 
 class stream {
+  bool m_non_blocking{false};
   k2::descriptor m_descriptor{k2::INVALID_PLATFORM_DESCRIPTOR};
   kphp::coro::io_scheduler& m_scheduler{kphp::coro::io_scheduler::get()};
 
   explicit stream(k2::descriptor descriptor) noexcept
       : m_descriptor(descriptor) {}
 
+  auto read_non_blocking(std::span<std::byte> buf) const noexcept -> kphp::coro::task<std::expected<size_t, int32_t>>;
```
apolyakov: It looks like these don't need to be coroutines

runtime-light/streams/stream.h
```diff
@@ -72,6 +77,20 @@ class stream {
 
 // ================================================================================================
 
+inline auto stream::read_non_blocking(std::span<std::byte> buf) const noexcept -> kphp::coro::task<std::expected<size_t, int32_t>> {
+  if (status().read_status != k2::IOStatus::IOAvailable) {
+    co_return std::expected<size_t, int32_t>{0};
+  }
+  co_return std::expected<size_t, int32_t>{k2::read(m_descriptor, buf)};
```
apolyakov: We can just `return k2::read(...)`

runtime-light/streams/stream.h
denisichh: done

runtime-light/streams/stream.h
denisichh: done

apolyakov (approved):
LGTM

PR #1634, opened 2026-05-27
[k2] fix f$_microtime_string function

Shamzik:
Use system_clock instead of high_resolution_clock


DrDet (changes requested):
Let's add tests for it

DrDet (approved):

PR #1635, opened 2026-06-15
[k2] rpc client API

T-y-c-o-o-n:
- added functions for rpc client API in k2 header: `k2_rpc_send_request`, `k2_rpc_get_response_size`, `k2_rpc_fetch_response`.
- added new K2 descriptor type: rpc-query. `k2_stream_status` accepts this descriptor. `k2_take_update` returns update on this descriptor.
- built-in functions `f$rpc_send_requests` and `f$rpc_send_typed_query_requests` uses new rpc client API instead of streams.


runtime-light/stdlib/rpc/rpc-api.cpp
```diff
@@ -273,86 +355,41 @@ kphp::coro::task<class_instance<C$VK$TL$RpcResponse>> typed_rpc_tl_query_result_
...
-    const vk::final_action finalizer{[&tl_storer, &rpc_server_instance_st] noexcept {
-      if (tl_storer.capacity() > rpc_server_instance_st.tl_storer.capacity()) {
-        std::swap(tl_storer, rpc_server_instance_st.tl_storer);
-      }
-    }};
-
-    // prepare and send RPC request
-    // 'request_buf' will look like this:
-    //    [ RpcExtraHeaders (optional) ] [ payload ]
-    if (const auto& [opt_new_extra_header, cur_extra_header_size]{kphp::rpc::regularize_extra_headers(tl_storer.view(), ignore_answer)}; opt_new_extra_header) {
```
T-y-c-o-o-n, line 301: В новом коде  нет вызова `regularize_extra_headers(...)`

runtime-light/stdlib/rpc/rpc-queue-functions.h
```diff
@@ -18,20 +18,27 @@
...
+    co_await m_scheduler.poll(request_info.rpc_d, kphp::coro::poll_op::read, timeout);
     co_return request_id;
   }};
 
   auto& rpc_client_instance_st{RpcClientInstanceState::get()};
 
-  const auto it_awaiter_task{rpc_client_instance_st.response_awaiter_tasks.find(request_id)};
-  if (it_awaiter_task == rpc_client_instance_st.response_awaiter_tasks.end()) [[unlikely]] {
-    kphp::log::warning("could not find rpc query with id {} in pending queries", queue_id);
+  const auto it_rpc_request_info{rpc_client_instance_st.rpc_requests_infos.find(request_id)};
```
T-y-c-o-o-n: кажется, это не будет работать для ignore_answer запросов (если для них, вообще, вызывается `rpc_queue_push`)

builtin-functions/kphp-light/stdlib/rpc.txt
```diff
@@ -40,15 +40,15 @@ final class KphpRpcRequestsExtraInfo {
     public function get ();
 }
 
-/** @kphp-extern-func-info interruptible */
+/** @kphp-extern-func-info */
```
apolyakov: We can completely remove this now

runtime-light/k2-platform/k2-header.h
```diff
@@ -316,6 +309,37 @@ int32_t k2_unlink(const char* path, size_t path_len);
  */
 int32_t k2_component_access(size_t name_len, const char* name);
 
+/**
+ * Try to send rpc request. If case of success write descriptor of rpc request to `rpc_d`, otherwise return `errno` != 0.
+ * which should be later used to call `k2_rpc_fetch_response`.
```
apolyakov: nit: fix grammar

runtime-light/k2-platform/k2-header.h
```diff
@@ -316,6 +309,37 @@ int32_t k2_unlink(const char* path, size_t path_len);
...
+ *
+ * @return return `0` on success. libc-like `errno` otherwise
+ *
+ * Possible `errno` values:
+ * `EAI_MEMORY` => max descriptors count achieved
+ * `EINVAL` => invalid `actor_name` or request, or connection pool is empty for this actor.
+ */
+int32_t k2_rpc_send_request(const char* actor_name, size_t actor_name_len, const void* request_ptr, size_t request_size, uint64_t* rpc_d);
```
apolyakov: 1. Maybe rename to `k2_rpc_send`?
2. Do we really need `actor_name`? Can't we just use something like `int64_t actor_id`?
3. Shoudn't we consider multiple RPC implementations here, e.g. `TL RPC`, `gRPC` etc?

runtime-light/k2-platform/k2-header.h
```diff
@@ -316,6 +309,37 @@ int32_t k2_unlink(const char* path, size_t path_len);
...
+
+/**
+ * Get response size for corresponding request of this `rpc_d`. Write 0 to `response_size` if response is not ready.
+ * Write response size value to `response_size` when response is ready.
```
apolyakov: `when response is ready` => `if response is ready`

runtime-light/k2-platform/k2-header.h
```diff
@@ -316,6 +309,37 @@ int32_t k2_unlink(const char* path, size_t path_len);
...
+ *
+ * @return return `0` on success. libc-like `errno` otherwise
+ *
+ * `EINVAL` => invalid `rpc_d` descriptor, for example, it is unknown descriptor, or not rpc descriptor.
```
apolyakov, line 461: Let's add something like ` * Possible errno values:
`

runtime-light/k2-platform/k2-header.h
```diff
@@ -316,6 +309,37 @@ int32_t k2_unlink(const char* path, size_t path_len);
...
+ */
+int32_t k2_rpc_get_response_size(uint64_t rpc_d, size_t* response_size);
+
+/**
+ * Write response for corresponding request of this `rpc_d` to `buf`. Does nothing if response is not ready.
```
apolyakov: IMO it's a bad design choice to have this function do nothing in case a response is not ready yet. We need to indicate this case explicitly

runtime-light/stdlib/rpc/rpc-api.cpp
```diff
@@ -111,79 +112,107 @@ class_instance<RpcTlQuery> store_function(const mixed& tl_object) noexcept {
...
 
     RpcClientInstanceState::get().response_fetcher_instances.emplace(query_info.id, std::move(rpc_tl_query));
   }
-  co_return query_info;
+  return query_info;
 }
 
+namespace {
+
+static std::expected<string, std::pair<int, string>> get_rpc_response(int64_t query_id, k2::descriptor rpc_d, bool collect_responses_extra_info) {
```
apolyakov: 1. We use sized integral types in `runtime-light`
2. We definitely need an RAII-like type for the RPC descriptor to prevent leaks that cause the instance to hang.
3. No need for `static` in anonymous namespace inside a `cpp` file

runtime-light/stdlib/rpc/rpc-api.cpp
```diff
@@ -111,79 +112,107 @@ class_instance<RpcTlQuery> store_function(const mixed& tl_object) noexcept {
...
+  std::expected<size_t, int32_t> first_response_size{k2::rpc_get_response_size(rpc_d)};
+  if (!first_response_size) {
+    return std::unexpected{std::make_pair(TL_ERROR_INTERNAL, string{"error fetching rpc response"})};
+  }
+  string response{reinterpret_cast<char*>(k2::alloc(*first_response_size)), static_cast<string::size_type>(*first_response_size)};
```
apolyakov: Looks like we can have a single buffer for RPC response to reuse

runtime-light/stdlib/rpc/rpc-api.cpp
```diff
@@ -196,8 +225,34 @@ kphp::coro::task<array<mixed>> rpc_tl_query_result_one_impl(int64_t query_id) no
...
+  std::optional<string> opt_response{std::nullopt};
+
+  kphp::log::assertion(opt_rpc_request_info.has_value());
+  auto rpc_request_info{*opt_rpc_request_info};
+  k2::TimePoint now_instant{};
+  k2::instant(std::addressof(now_instant));
+  std::chrono::nanoseconds now_ns{now_instant.time_point_ns};
+  std::chrono::nanoseconds timeout{rpc_request_info.deadline - now_ns};
+  kphp::coro::io_scheduler& m_scheduler{kphp::coro::io_scheduler::get()};
+  switch (co_await m_scheduler.poll(rpc_request_info.rpc_d, kphp::coro::poll_op::read, timeout)) {
```
apolyakov: It's quite hard to understand what's going on here. Can you please make it more human readable? For example, I think we don't need to directly use the scheduler's API here

runtime-light/stdlib/rpc/rpc-api.cpp
```diff
@@ -196,8 +225,34 @@ kphp::coro::task<array<mixed>> rpc_tl_query_result_one_impl(int64_t query_id) no
...
-  kphp::log::assertion(opt_awaiter_task.has_value());
-  auto opt_response{co_await kphp::forks::id_managed(*std::exchange(opt_awaiter_task, std::nullopt))};
+  std::optional<string> opt_response{std::nullopt};
+
+  kphp::log::assertion(opt_rpc_request_info.has_value());
+  auto rpc_request_info{*opt_rpc_request_info};
+  k2::TimePoint now_instant{};
+  k2::instant(std::addressof(now_instant));
+  std::chrono::nanoseconds now_ns{now_instant.time_point_ns};
+  std::chrono::nanoseconds timeout{rpc_request_info.deadline - now_ns};
```
apolyakov: What will happen if nobody will fetch response? I'm afraid we might hang in this case

apolyakov (changes requested):

runtime-light/k2-platform/k2-header.h
T-y-c-o-o-n: fixed

runtime-light/k2-platform/k2-header.h
T-y-c-o-o-n: added

builtin-functions/kphp-light/stdlib/rpc.txt
T-y-c-o-o-n: removed

runtime-light/k2-platform/k2-header.h
T-y-c-o-o-n: 1. renamed
2. some clusters have equal actor_id and can be identified only by actor_name or ip:port (for example, "config", "mc_local", "statshouse", "dolphin" all have actor_id=0). So we must either pass rpc connection (like in old runtime) or pass some identifier (like actor name), wich can be mapped to actor_id and ip:port or connection pool.
3. added RpcKind enum

runtime-light/k2-platform/k2-header.h
T-y-c-o-o-n: fixed

runtime-light/stdlib/rpc/rpc-api.cpp
T-y-c-o-o-n: 1. replaced int with int32_t
2. Encaplusated all k2-rpc-api logic into `kphp::rpc::query_handle` class
3. removed

runtime-light/stdlib/rpc/rpc-api.cpp
T-y-c-o-o-n: Tried to refactor and improve code readability:
- k2-rpc-api logic refactored to separate `kphp::rpc::query_handle` class
- time juggling moved to util functions

runtime-light/k2-platform/k2-header.h
T-y-c-o-o-n: Now k2_rpc_get_response_size and k2_rpc_fetch_response return `EAGAIN`, if response is not ready

runtime-light/k2-platform/k2-header.h
```diff
@@ -316,6 +320,42 @@ int32_t k2_unlink(const char* path, size_t path_len);
...
+ * Try to send rpc request. In case of success write descriptor of rpc request to `rpc_d`, otherwise return `errno` != 0,
+ * which should be later used to call `k2_rpc_fetch_response`.
+ *
+ * @return return `0` on success. libc-like `errno` otherwise
+ *
+ * Possible `errno` values:
+ * `EAI_MEMORY` => max descriptors count achieved
+ * `EINVAL` => invalid `actor_name` or request, or connection pool is empty for this actor.
+ */
+int32_t k2_rpc_send(const char* actor_name, size_t actor_name_len, const void* request_ptr, size_t request_size, enum RpcKind rpc_kind, uint64_t* rpc_d);
```
apolyakov: Let's do either:

- rename `k2_rpc_send` -> `k2_rpc_send_request`
- rename `k2_rpc_fetch_response` -> `k2_rpc_fetch`, `k2_rpc_get_response_size` -> `k2_rpc_response_size`

runtime-light/k2-platform/k2-api.h
```diff
@@ -210,6 +210,30 @@ inline int32_t component_access(std::string_view component_name) noexcept {
   return k2_component_access(component_name.size(), component_name.data());
 }
 
+inline std::expected<uint64_t, int32_t> rpc_send_request(std::string_view actor_name, std::span<const std::byte> request_buffer) noexcept {
```
apolyakov: - we have `k2::descriptor` type that should be used here instead of `uint64_t`
- shouldn't we add `rpc_kind` parameter?

runtime-light/stdlib/time/util.h
apolyakov, line 1: Let's put these functions to `kphp::time` namespace and name them like `kphp::time::remaining` and `kphp::time::expires_at`.

P.S. we try to not introduce `util` as it leads to a bunch of disparate functionality.

T-y-c-o-o-n: Done

runtime-light/k2-platform/k2-api.h
T-y-c-o-o-n: Done both

runtime-light/k2-platform/k2-header.h
T-y-c-o-o-n: I've chosen first variant

runtime-light/k2-platform/k2-header.h
```diff
@@ -316,6 +320,42 @@ int32_t k2_unlink(const char* path, size_t path_len);
...
+ *
+ * Possible `errno` values:
+ * `EINVAL` => invalid `rpc_d` descriptor, for example, it is unknown descriptor, or not rpc descriptor.
+ * `EAGAIN` => response is not ready yet.
+ */
+int32_t k2_rpc_get_response_size(uint64_t rpc_d, size_t* response_size);
+
+/**
+ * Write response for corresponding request of this `rpc_d` to `buf`. Return `EAGAIN` if response is not ready.
+ * If `buf_size` < response size, then write first `buf_size` bytes of response to `buf`.
```
apolyakov: I think it's pointless to write first `buf_size` bytes unless the next call to `k2_rpc_fetch_response` writes second `buf_size` bytes

runtime-light/k2-platform/k2-api.h
```diff
@@ -210,6 +212,31 @@ inline int32_t component_access(std::string_view component_name) noexcept {
   return k2_component_access(component_name.size(), component_name.data());
 }
 
+inline std::expected<k2::descriptor, int32_t> rpc_send_request(std::string_view actor_name, std::span<const std::byte> request_buffer,
+                                                               RpcKind rpc_kind) noexcept {
```
apolyakov: nit: it's better to use qualified name `k2::RpcKind`

runtime-light/stdlib/rpc/rpc-client-state.h
apolyakov: Looks like it's unused now

runtime-light/stdlib/rpc/rpc-client-state.h
```diff
@@ -16,18 +16,19 @@
...
+#include "runtime-light/stdlib/rpc/rpc-query-handle.h"
 #include "runtime-light/stdlib/rpc/rpc-tl-defs.h"
 #include "runtime-light/stdlib/rpc/rpc-tl-query.h"
 
 struct RpcClientInstanceState final : private vk::not_copyable {
   CurrentTlQuery current_client_query{};
   int64_t current_query_id{kphp::rpc::VALID_QUERY_ID_RANGE_START};
 
+  kphp::stl::unordered_map<int64_t, kphp::rpc::query_handle, kphp::memory::script_allocator> rpc_query_handles;
   kphp::stl::unordered_map<int64_t, kphp::coro::shared_task<std::optional<string>>, kphp::memory::script_allocator> response_awaiter_tasks;
```
apolyakov: Looks like it's unused now

runtime-light/stdlib/rpc/rpc-query.h
```diff
@@ -0,0 +1,116 @@
+// Compiler for PHP (aka KPHP)
+// Copyright (c) 2026 LLC «V Kontakte»
+// Distributed under the GPL v3 License, see LICENSE.notice.txt
+
+#pragma once
+
+#include <chrono>
```
apolyakov, line 7: Could you please run and check `clangd` diagnostics? There are 5 warnings and errors in these includes

runtime-light/stdlib/rpc/rpc-query-handle.h
```diff
@@ -0,0 +1,116 @@
...
+#include "runtime-light/coroutine/task.h"
+#include "runtime-light/k2-platform/k2-api.h"
+#include "runtime-light/stdlib/diagnostics/logs.h"
+#include "runtime-light/stdlib/rpc/rpc-client-state.h"
+#include "runtime-light/stdlib/time/time-functions.h"
+
+namespace kphp::rpc {
+
+class query_handle {
+  k2::descriptor rpc_d{k2::INVALID_PLATFORM_DESCRIPTOR};
```
apolyakov: nit:

- we prefer `m_` prefix for members
- we prefer meaningful names, e.g. `m_descriptor` instead of `rpc_d`. We already know we work with `rpc::query_handle` so it's useless to repeat `rpc` one more time

runtime-light/stdlib/rpc/rpc-query-handle.h
```diff
@@ -0,0 +1,116 @@
...
+  int64_t query_id;
+  std::chrono::nanoseconds deadline{};
+  bool collect_responses_extra_info{false};
+
+public:
+  query_handle() = delete;
+
+  query_handle(k2::descriptor _rpc_d, int64_t _query_id, std::chrono::nanoseconds _deadline, bool _collect_responses_extra_info) noexcept
```
apolyakov: Let's use `Rust`-like pattern for constructors to check `query_handle`'s invariants.

runtime-light/stdlib/rpc/rpc-query-handle.h
```diff
@@ -0,0 +1,116 @@
...
+      : rpc_d{_rpc_d},
+        query_id{_query_id},
+        deadline{_deadline},
+        collect_responses_extra_info{_collect_responses_extra_info} {}
+
+  query_handle(query_handle&& other) noexcept
+      : rpc_d{std::exchange(other.rpc_d, k2::INVALID_PLATFORM_DESCRIPTOR)},
+        query_id{other.query_id},
```
apolyakov: It might be worth it to reset `other.query_id` here

runtime-light/stdlib/rpc/rpc-query-handle.h
```diff
@@ -0,0 +1,116 @@
...
+      collect_responses_extra_info = other.collect_responses_extra_info;
+    }
+    return *this;
+  }
+
+  query_handle(const query_handle& other) = delete;
+
+  query_handle& operator=(const query_handle& other) = delete;
+
+  ~query_handle() noexcept {
```
apolyakov: Destructors in C++ are implicitly `noexcept`

runtime-light/stdlib/rpc/rpc-query-handle.h
```diff
@@ -0,0 +1,116 @@
...
+    drop();
+  }
+
+  void drop() noexcept {
```
apolyakov: Do we need this function to be in `query_handle`'s public API?

runtime-light/stdlib/rpc/rpc-query-handle.h
```diff
@@ -0,0 +1,116 @@
...
+    if (rpc_d != k2::INVALID_PLATFORM_DESCRIPTOR) {
+      k2::free_descriptor(std::exchange(rpc_d, k2::INVALID_PLATFORM_DESCRIPTOR));
+    }
+  }
+
+  kphp::coro::task<void> wait_for_response() noexcept;
+
+  kphp::coro::task<std::expected<string, std::pair<int32_t, string>>> get_response() noexcept;
```
apolyakov: Let's try to get rid of core `kPHP` types here

runtime-light/stdlib/rpc/rpc-query-handle.h
```diff
@@ -0,0 +1,116 @@
...
+
+inline kphp::coro::task<void> query_handle::wait_for_response() noexcept {
+  if (rpc_d == k2::INVALID_PLATFORM_DESCRIPTOR) {
+    co_return;
+  }
+
+  kphp::coro::io_scheduler& m_scheduler{kphp::coro::io_scheduler::get()};
+  std::chrono::nanoseconds timeout{kphp::time::remaining(deadline)};
+
+  co_await m_scheduler.poll(rpc_d, kphp::coro::poll_op::read, timeout);
```
apolyakov: Looks like it's possible for `timeout` to become zero or negative here. Please, read `io_scheduler::poll` documentation for timeout

runtime-light/stdlib/rpc/rpc-query-handle.h
```diff
@@ -0,0 +1,116 @@
...
+    drop();
+  }
+
+  void drop() noexcept {
+    if (rpc_d != k2::INVALID_PLATFORM_DESCRIPTOR) {
+      k2::free_descriptor(std::exchange(rpc_d, k2::INVALID_PLATFORM_DESCRIPTOR));
+    }
+  }
+
+  kphp::coro::task<void> wait_for_response() noexcept;
```
apolyakov: Let's take one more step if you started splitting declarations and definitions:

1. let's have `query_handle`'s body only include declarations with doc comments
2. let's have all implementation defined out-of-class

runtime-light/stdlib/rpc/rpc-query-handle.h
```diff
@@ -0,0 +1,116 @@
...
+    co_return get_ready_response();
+  case kphp::coro::poll_status::closed:
+  case kphp::coro::poll_status::timeout:
+    co_return std::unexpected{std::make_pair(TL_ERROR_QUERY_TIMEOUT, string{"rpc response timeout"})};
+  case kphp::coro::poll_status::error:
+    co_return std::unexpected{std::make_pair(TL_ERROR_INTERNAL, string{"error fetching rpc response"})};
+  }
+}
+
+std::expected<query_handle, int32_t> send_and_get_handle(std::string_view actor, bool collect_responses_extra_info, bool ignore_answer,
```
apolyakov: nit: why not just `send`? Do we really need to highlight that we get handle as a return value?

I don't understand why do we need this function at all. It seems like it's literally part of `kphp::rpc::send_request`

runtime-light/stdlib/rpc/rpc-queue-functions.h
```diff
@@ -17,24 +17,23 @@
...
 inline void rpc_queue_push(int64_t queue_id, int64_t request_id) noexcept {
-  static constexpr auto rpc_queue_wrapper_task{[](kphp::coro::shared_task<> awaiter_task, int64_t request_id) noexcept -> kphp::coro::task<int64_t> {
-    co_await std::move(awaiter_task).when_ready();
+  static constexpr auto rpc_queue_wrapper_task{[](int64_t request_id) noexcept -> kphp::coro::task<int64_t> {
+    auto& rpc_client_instance_st{RpcClientInstanceState::get()};
+    const auto it_rpc_request_handle{rpc_client_instance_st.rpc_query_handles.find(request_id)};
+    if (it_rpc_request_handle == rpc_client_instance_st.rpc_query_handles.end()) [[unlikely]] {
+      co_return request_id;
+    }
+    // FIXME after response fetch finished `it_rpc_request_handle` will become invalid iterator
```
apolyakov: So, let's fix this :)

runtime-light/stdlib/time/time-functions.h
```diff
@@ -29,7 +32,39 @@ int64_t fix_year(int64_t year) noexcept;
...
+/**
+ * Calculate time remaining to the deadline.
+ */
+template<kphp::concepts::duration duration_type>
+duration_type remaining(duration_type deadline) noexcept {
+  k2::TimePoint now_instant{};
+  k2::instant(std::addressof(now_instant));
+
+  std::chrono::nanoseconds now_ns{now_instant.time_point_ns};
+  std::chrono::nanoseconds deadline_ns{duration_cast<std::chrono::nanoseconds>(deadline)};
```
apolyakov: 1. What is `duration_cast` here?
2. It looks like you use duration as time point. Please, correct me if I'm wrong

apolyakov (changes requested):

runtime-light/stdlib/rpc/rpc-queue-functions.h
T-y-c-o-o-n: WE discussed, that wait_for_response may await inside and overlife object destruction, but this is not a problem, because after wait it does not touch object fields

runtime-light/stdlib/rpc/rpc-query-handle.h
T-y-c-o-o-n: - Renamed it with send
- I've moved logic about regularize_extra_headers and collect_response_extra to `kphp::rpc::send_request`, and leaved here only logic about creating query. That allowed me to avoid `rpc_query_handle.cpp` file

runtime-light/stdlib/rpc/rpc-query-handle.h
T-y-c-o-o-n: Done. Only constructors, destructor and assign operators have definition inside class

runtime-light/k2-platform/k2-api.h
T-y-c-o-o-n: done

runtime-light/stdlib/rpc/rpc-client-state.h
T-y-c-o-o-n: removed

runtime-light/stdlib/rpc/rpc-query-handle.h
T-y-c-o-o-n: refactored

runtime-light/stdlib/rpc/rpc-query-handle.h
T-y-c-o-o-n: yeap, made this constructor private and made `send` static method inside `query_handle`

runtime-light/stdlib/rpc/rpc-query-handle.h
T-y-c-o-o-n: removed completely `id` from `query_handle`

runtime-light/stdlib/rpc/rpc-query-handle.h
T-y-c-o-o-n: removed

runtime-light/stdlib/rpc/rpc-query-handle.h
T-y-c-o-o-n: Yeah, i've decided to leave it public, because outer code can call this function without invariants violation and because it can be useful API for canceling requests.

runtime-light/stdlib/rpc/rpc-query-handle.h
T-y-c-o-o-n: resplaced strings with std::span<std::byte> for responses and std::string_view for error descriptions

runtime-light/stdlib/rpc/rpc-query-handle.h
T-y-c-o-o-n: Yeap, handled `timeout <= 0` case manually

runtime-light/stdlib/time/time-functions.h
T-y-c-o-o-n: You'r right. Replaced deadlines types with `std::chrono::steady_clock::time_point`

runtime-light/stdlib/rpc/rpc-query-handle.h
```diff
@@ -0,0 +1,155 @@
...
+#include "common/rpc-error-codes.h"
+#include "runtime-common/stdlib/string/string-context.h"
+#include "runtime-light/coroutine/io-scheduler.h"
+#include "runtime-light/coroutine/task.h"
+#include "runtime-light/k2-platform/k2-api.h"
+#include "runtime-light/stdlib/time/time-functions.h"
+
+namespace kphp::rpc {
+
+inline constexpr std::string_view EMPTY_QUERY_ERROR_DESCRIPTION = "fetching rpc response from empty query";
```
apolyakov: Do we need this? Isn't it enough to return an error code?

runtime-light/stdlib/rpc/rpc-query-handle.h
```diff
@@ -0,0 +1,155 @@
...
+inline constexpr std::string_view TIMEOUT_ERROR_DESCRIPTION = "rpc response timeout";
+inline constexpr std::string_view INTERNAL_ERROR_DESCRIPTION = "internal error while fetching rpc response";
+
+class query_handle {
+  k2::descriptor m_descriptor{k2::INVALID_PLATFORM_DESCRIPTOR};
+  std::chrono::steady_clock::time_point m_deadline{};
```
apolyakov: nit: no need for explicit init here

runtime-light/stdlib/rpc/rpc-query-handle.h
T-y-c-o-o-n: laeved only rpc error code

runtime-light/stdlib/rpc/rpc-query-handle.h
T-y-c-o-o-n: removed

runtime-light/k2-platform/k2-api.h
```diff
@@ -210,6 +213,32 @@ inline int32_t component_access(std::string_view component_name) noexcept {
...
+  k2::descriptor descriptor{};
+  if (auto error_code{
+          k2_rpc_send_request(actor_name.data(), actor_name.size(), request_buffer.data(), request_buffer.size(), rpc_kind, std::addressof(descriptor))};
+      error_code != k2::errno_ok) {
+    return std::unexpected{error_code};
+  }
+  return {descriptor};
+}
+
+inline std::expected<size_t, int32_t> rpc_get_response_size(uint64_t rpc_d) noexcept {
```
apolyakov: Let's use `k2::descriptor` instead of `uint64_t`

runtime-light/k2-platform/k2-header.h
```diff
@@ -316,6 +320,43 @@ int32_t k2_unlink(const char* path, size_t path_len);
  */
 int32_t k2_component_access(size_t name_len, const char* name);
 
+/**
+ * Try to send rpc request. In case of success write descriptor of rpc request to `rpc_d`, otherwise return `errno` != 0,
```
apolyakov: Could you please make these docs more clear?

For example, for `k2_rpc_send_request` it might look like:
```
* Try to send an RPC request. On success, write the request descriptor to `rpc_d`;
* this descriptor is later used when calling `k2_rpc_fetch_response`.
* Otherwise, return a non-zero `errno`.
...
```

Please, take a look at other API functions you've added and check their docs too

runtime-light/stdlib/rpc/rpc-client-state.h
```diff
@@ -23,11 +23,12 @@ struct RpcClientInstanceState final : private vk::not_copyable {
   CurrentTlQuery current_client_query{};
   int64_t current_query_id{kphp::rpc::VALID_QUERY_ID_RANGE_START};
 
-  kphp::stl::unordered_map<int64_t, kphp::coro::shared_task<std::optional<string>>, kphp::memory::script_allocator> response_awaiter_tasks;
+  kphp::stl::unordered_map<int64_t, kphp::coro::shared_task<std::expected<kphp::stl::vector<std::byte, kphp::memory::script_allocator>, int32_t>>,
```
apolyakov: This is not a good idea to have `shared_task` returning something like `vector`. `shared_task` returns a copy of its result on each `co_await`, so you'll end up with multiple copies of the vector

runtime-light/stdlib/rpc/rpc-query.h
```diff
@@ -0,0 +1,136 @@
...
+class query {
+  k2::descriptor m_descriptor{k2::INVALID_PLATFORM_DESCRIPTOR};
+  std::chrono::steady_clock::time_point m_deadline;
+
+  query(k2::descriptor descriptor, std::chrono::steady_clock::time_point deadline) noexcept
+      : m_descriptor{descriptor},
+        m_deadline{deadline} {}
+
+  template<std::invocable<size_t> B>
+  requires std::is_convertible_v<std::invoke_result_t<B, size_t>, std::span<std::byte>>
```
apolyakov: Let's make it `std::same_as<std::span<byte>>`

runtime-light/stdlib/rpc/rpc-query.h
```diff
@@ -0,0 +1,136 @@
...
+  auto get_ready_response(B response_buffer_provider) noexcept -> std::expected<std::span<std::byte>, int32_t>;
```
apolyakov: Why do we need to accept this parameter by value?

runtime-light/stdlib/rpc/rpc-query.h
```diff
@@ -0,0 +1,136 @@
...
+
+  query(const query& other) = delete;
+  query& operator=(const query& other) = delete;
+
+  static auto send(std::string_view actor, std::chrono::milliseconds timeout,
+                   std::span<const std::byte> request_buffer) noexcept -> std::expected<query, int32_t>;
+
+  template<std::invocable<size_t> B>
+  requires std::is_convertible_v<std::invoke_result_t<B, size_t>, std::span<std::byte>>
+  auto get_response(B response_buffer_provider) noexcept -> kphp::coro::task<std::expected<std::span<std::byte>, int32_t>>;
```
apolyakov: Just a note to think about: we can get rid of this method. Instead, we might have an interface like this:

```
kphp::rpc::query q{create_some_query()};
co_await q(my_response_buffer_provider);
```

What do you think?

runtime-light/stdlib/rpc/rpc-query.h
```diff
@@ -0,0 +1,136 @@
...
+    switch (response_size_exp.error()) {
+    case k2::errno_eagain:
+      return std::unexpected{TL_ERROR_QUERY_TIMEOUT};
+    default:
+      return std::unexpected{TL_ERROR_INTERNAL};
+    }
+  }
+  size_t response_size{*response_size_exp};
+
+  std::span<std::byte> response_buffer{std::invoke(response_buffer_provider, response_size)};
```
apolyakov: I think we need to check `response_buffer`'s size here

runtime-light/stdlib/rpc/rpc-query.h
```diff
@@ -0,0 +1,136 @@
...
+  }
+
+  return {response_buffer};
+}
+
+template<std::invocable<size_t> B>
+requires std::is_convertible_v<std::invoke_result_t<B, size_t>, std::span<std::byte>>
+auto query::get_response(B response_buffer_provider) noexcept -> kphp::coro::task<std::expected<std::span<std::byte>, int32_t>> {
+  if (m_descriptor == k2::INVALID_PLATFORM_DESCRIPTOR) {
+    co_return std::unexpected{TL_ERROR_INTERNAL};
```
apolyakov: We can use `TL_ERROR_INVALID_CONNECTION_ID` to improve observability

runtime-light/stdlib/rpc/rpc-query.h
```diff
@@ -0,0 +1,136 @@
...
+  }
+
+  kphp::coro::io_scheduler& m_scheduler{kphp::coro::io_scheduler::get()};
+  std::chrono::nanoseconds timeout{kphp::time::remaining(m_deadline)};
+
+  // TODO DISCUSS: may be completely remove timeout monitoring in kphp and leave it in k2-node's rpc client ???
```
apolyakov: Seems reasonable now. There are two ways the platform can notify us about RPC timeout:

1. close the stream -> aren't there other errors that can lead to stream being closed?
2. write some "there was a timeout" answer to the stream -> requires a bit more complex processing since such an answer will vary depending on RPC kind we use

runtime-light/stdlib/time/time-functions.h
```diff
@@ -29,7 +32,33 @@ int64_t fix_year(int64_t year) noexcept;
 
 string date(const string& format, const tm& t, int64_t timestamp, bool local) noexcept;
 
-} // namespace kphp::time::impl
+} // namespace impl
+
+/**
+ * Calculate time remaining to the deadline.
+ */
+inline std::chrono::nanoseconds remaining(std::chrono::steady_clock::time_point deadline) noexcept {
```
apolyakov: What do you think about following implementation:
```
/**             
   * Returns the current steady-clock time point from the K2 platform.                                                                                                                                                                                         
   */             
  inline std::chrono::steady_clock::time_point now() noexcept /* iff k2::instant is noexcept */ {
    k2::TimePoint now_instant{};
    k2::instant(std::addressof(now_instant));
    return std::chrono::steady_clock::time_point{std::chrono::nanoseconds{now_instant.time_point_ns}};
  }

  /**
   * Calculate time remaining until the deadline.
   */
  template <typename Duration>
  inline auto remaining(std::chrono::time_point<std::chrono::steady_clock, Duration> deadline) noexcept {
    return deadline - now();
  }

  /**
   * Converts a timeout to the time point at which it elapses, i.e. the deadline.
   */
  template <typename Rep, typename Period>
  inline auto expires_at(std::chrono::duration<Rep, Period> timeout) noexcept {
    return now() + timeout;
  }
```

It makes client code easier by removing mandatory conversions from any duration to nanoseconds

apolyakov (changes requested):

runtime-light/stdlib/rpc/rpc-api.h
```diff
@@ -287,9 +286,8 @@ inline int64_t f$rpc_tl_pending_queries_count() noexcept {
 
 // === client untyped =============================================================================
 
-inline kphp::coro::task<array<int64_t>> f$rpc_send_requests(string actor, array<mixed> tl_objects, Optional<double> timeout, bool ignore_answer,
-                                                            class_instance<C$KphpRpcRequestsExtraInfo> requests_extra_info,
-                                                            bool need_responses_extra_info) noexcept {
+inline array<int64_t> f$rpc_send_requests(string actor, array<mixed> tl_objects, Optional<double> timeout, bool ignore_answer,
```
apolyakov: As it's not longer a coroutine, we don't need to pass parameters by value

runtime-light/stdlib/rpc/rpc-query.h
```diff
@@ -0,0 +1,137 @@
...
+class query {
+  k2::descriptor m_descriptor{k2::INVALID_PLATFORM_DESCRIPTOR};
+  std::chrono::steady_clock::time_point m_deadline;
+
+  query(k2::descriptor descriptor, std::chrono::steady_clock::time_point deadline) noexcept
+      : m_descriptor{descriptor},
+        m_deadline{deadline} {}
+
+  template<std::invocable<size_t> B>
+  requires std::is_same_v<std::invoke_result_t<B, size_t>, std::span<std::byte>>
```
apolyakov: missing includes for `concepts` and `type_traits`

runtime-light/stdlib/rpc/rpc-query.h
```diff
@@ -0,0 +1,137 @@
...
+
+inline auto query::drop() noexcept -> void {
+  if (m_descriptor != k2::INVALID_PLATFORM_DESCRIPTOR) {
+    k2::free_descriptor(std::exchange(m_descriptor, k2::INVALID_PLATFORM_DESCRIPTOR));
+  }
+}
+
+inline auto query::send(std::string_view actor, std::chrono::milliseconds timeout,
+                        std::span<const std::byte> request_buffer) noexcept -> std::expected<query, int32_t> {
+  auto descriptor_exp{k2::rpc_send_request(actor, request_buffer, RpcKind::TL_RPC)};
```
apolyakov: I think `send` needs to accept RPC kind as a parameter

runtime-light/stdlib/rpc/rpc-query.h
```diff
@@ -0,0 +1,137 @@
...
+}
+
+template<std::invocable<size_t> B>
+requires std::is_same_v<std::invoke_result_t<B, size_t>, std::span<std::byte>>
+auto query::response(B response_buffer_provider) && noexcept -> kphp::coro::task<std::expected<std::span<std::byte>, int32_t>> {
+  if (m_descriptor == k2::INVALID_PLATFORM_DESCRIPTOR) {
+    co_return std::unexpected{TL_ERROR_INVALID_CONNECTION_ID};
+  }
+
+  kphp::coro::io_scheduler& m_scheduler{kphp::coro::io_scheduler::get()};
```
apolyakov: nit: no need to get the scheduler unless it's really needed

runtime-light/stdlib/rpc/rpc-query.h
```diff
@@ -0,0 +1,137 @@
...
+  auto timeout{kphp::time::remaining(m_deadline)};
+
+  if (timeout <= std::chrono::nanoseconds::zero()) {
+    co_return std::move(*this).get_ready_response<B>(std::move(response_buffer_provider));
```
apolyakov: nit: I think it's better to declare `get_ready_response` and `static constexpr` lambda inside `response`

runtime-light/stdlib/rpc/rpc-api.cpp
```diff
@@ -111,50 +113,50 @@ class_instance<RpcTlQuery> store_function(const mixed& tl_object) noexcept {
   return rpc_tl_query;
 }
 
-kphp::coro::task<kphp::rpc::query_info> rpc_tl_query_one_impl(std::string_view actor, mixed tl_object, std::optional<double> opt_timeout,
-                                                              bool collect_resp_extra_info, bool ignore_answer) noexcept {
+kphp::rpc::query_info rpc_tl_query_one_impl(std::string_view actor, mixed tl_object, std::optional<double> opt_timeout, bool collect_resp_extra_info,
```
apolyakov: No need to accept `tl_object` by value

runtime-light/stdlib/rpc/rpc-api.cpp
```diff
@@ -273,42 +275,44 @@ kphp::coro::task<class_instance<C$VK$TL$RpcResponse>> typed_rpc_tl_query_result_
...
+
   const size_t request_size{rpc_server_instance_st.tl_storer.view().size_bytes()};
   const auto timestamp{std::chrono::duration<double>{std::chrono::system_clock::now().time_since_epoch()}.count()};
 
-  auto expected_stream{kphp::component::stream::open(actor, k2::stream_kind::component)};
-  if (!expected_stream) [[unlikely]] {
-    co_return kphp::rpc::query_info{
-        .id = kphp::rpc::INVALID_QUERY_ID, .request_size = rpc_server_instance_st.tl_storer.view().size_bytes(), .timestamp = timestamp};
-  }
+  auto tl_storer{std::exchange(rpc_server_instance_st.tl_storer, tl::storer{0})};
```
apolyakov: We don't need this anymore as `send_request` is not a coroutine now. Hence it's impossible that this buffer is going to be used concurrently

runtime-light/stdlib/rpc/rpc-api.cpp
```diff
@@ -273,42 +275,44 @@ kphp::coro::task<class_instance<C$VK$TL$RpcResponse>> typed_rpc_tl_query_result_
...
+  if (const auto& [opt_new_extra_header, cur_extra_header_size]{kphp::rpc::regularize_extra_headers(tl_storer.view(), ignore_answer)}; opt_new_extra_header) {
+    std::span<const std::byte> new_header{reinterpret_cast<const std::byte*>(std::addressof(*opt_new_extra_header)),
+                                          sizeof(std::remove_cvref_t<decltype(*opt_new_extra_header)>)};
+    std::span<const std::byte> request_body{tl_storer.view().subspan(cur_extra_header_size)};
+
+    std::span<std::byte> new_request_buffer{};
+    size_t request_and_headers_size{new_header.size() + request_body.size()};
+    if (request_and_headers_size <= StringLibContext::STATIC_BUFFER_LENGTH) {
+      // we have enough space in static buffer to store request with regularized headers
+      auto& string_lib_ctx{StringLibContext::get()};
```
apolyakov, line 292: Please use `StringInstanceState` isntead of `StringLibContext`

runtime-light/stdlib/rpc/rpc-api.cpp
```diff
@@ -273,42 +275,44 @@ kphp::coro::task<class_instance<C$VK$TL$RpcResponse>> typed_rpc_tl_query_result_
...
 
-  auto stream{*std::move(expected_stream)};
-  {
-    auto tl_storer{std::exchange(rpc_server_instance_st.tl_storer, tl::storer{0})};
-    const vk::final_action finalizer{[&tl_storer, &rpc_server_instance_st] noexcept {
-      if (tl_storer.capacity() > rpc_server_instance_st.tl_storer.capacity()) {
-        std::swap(tl_storer, rpc_server_instance_st.tl_storer);
-      }
-    }};
+  // if we have to allocate memory for request buffer, it will be in this vector, and it will be freed at the end of the function
```
apolyakov, line 279: I think it's better to use the `tl_storer` buffer for this

runtime-light/stdlib/rpc/rpc-api.cpp
```diff
@@ -318,37 +322,45 @@ kphp::coro::task<kphp::rpc::query_info> send_request(std::string_view actor, std
...
 
-  // create a task to wait for RPC response. we need to do it even if 'ignore_answer' is 'true' to make sure
-  // that the stream will not be closed too early. otherwise, platform may even not send RPC request
-  static constexpr auto awaiter_coroutine{[](int64_t query_id, kphp::component::stream stream, std::chrono::milliseconds timeout,
-                                             bool collect_responses_extra_info) noexcept -> kphp::coro::shared_task<std::optional<string>> {
-    std::optional<string> opt_response{std::in_place};
-    auto fetch_task{kphp::component::fetch_response(stream, kphp::component::read_ext::append(*opt_response))};
-    if (auto expected{co_await kphp::coro::io_scheduler::get().schedule(std::move(fetch_task), timeout)}; !expected) [[unlikely]] {
-      opt_response = std::nullopt;
+  static constexpr auto awaiter_coroutine{[](query q, int64_t query_id, bool collect_responses_extra_info) mutable noexcept
```
apolyakov: Why is it `mutable`?

runtime-light/stdlib/rpc/rpc-api.cpp
```diff
@@ -318,37 +322,45 @@ kphp::coro::task<kphp::rpc::query_info> send_request(std::string_view actor, std
...
+                                          -> kphp::coro::shared_task<std::expected<string, int32_t>> {
+    string resp_buf{};
+    const auto response_buffer_provider{[&resp_buf](size_t size) noexcept -> std::span<std::byte> {
+      resp_buf.assign(size, true);
```
apolyakov: `reserve_at_least` better expresses what you do here

runtime-light/stdlib/rpc/rpc-api.cpp
```diff
@@ -318,37 +322,45 @@ kphp::coro::task<kphp::rpc::query_info> send_request(std::string_view actor, std
...
+      return {reinterpret_cast<std::byte*>(resp_buf.buffer()), size};
+    }};
+
+    auto fetch_task{std::move(q).response<decltype(response_buffer_provider)>(response_buffer_provider)};
```
apolyakov: No need for explicit `decltype` here

runtime-light/stdlib/rpc/rpc-api.cpp
```diff
@@ -318,37 +322,45 @@ kphp::coro::task<kphp::rpc::query_info> send_request(std::string_view actor, std
...
-    co_return std::move(opt_response);
+    co_return std::move(resp_buf);
   }};
 
-  static constexpr auto ignore_answer_awaiter_coroutine{
-      [](kphp::component::stream stream, std::chrono::milliseconds timeout) noexcept -> kphp::coro::shared_task<> {
-        auto fetch_task{kphp::component::fetch_response(stream, [](std::span<const std::byte>) noexcept {})};
-        std::ignore = co_await kphp::coro::io_scheduler::get().schedule(std::move(fetch_task), timeout);
-      }};
+  static constexpr auto ignore_answer_awaiter_coroutine{[](query q) noexcept -> kphp::coro::shared_task<> {
```
apolyakov: Do we even need it now?

T-y-c-o-o-n: Yeap, leaved a comment about it. We need it to prevent destrucing `query`, leading to descriptor free, leading to request drop

runtime-light/stdlib/rpc/rpc-api.cpp
T-y-c-o-o-n: +

runtime-light/stdlib/rpc/rpc-api.cpp
T-y-c-o-o-n: I don't think so, because we need to increase size of string to return response anyway

runtime-light/stdlib/rpc/rpc-api.cpp
T-y-c-o-o-n: removed

runtime-light/stdlib/rpc/rpc-api.cpp
T-y-c-o-o-n: used

runtime-light/stdlib/rpc/rpc-api.cpp
T-y-c-o-o-n: used

runtime-light/stdlib/rpc/rpc-api.cpp
T-y-c-o-o-n: removed

runtime-light/stdlib/rpc/rpc-api.cpp
T-y-c-o-o-n: added references

runtime-light/stdlib/rpc/rpc-query.h
T-y-c-o-o-n: declared

runtime-light/stdlib/rpc/rpc-query.h
T-y-c-o-o-n: fixed

runtime-light/stdlib/rpc/rpc-query.h
T-y-c-o-o-n: added

runtime-light/stdlib/rpc/rpc-query.h
T-y-c-o-o-n: added missing includes also for span and invoke

runtime-light/stdlib/rpc/rpc-api.h
T-y-c-o-o-n: added const references, but not for class_instance, because it prohibits passing null to it

runtime-light/stdlib/rpc/rpc-api.cpp
T-y-c-o-o-n: We decided to removed it and to fix `regularize_extra_headers` to pass no_result bit to k2-node, so that `query` drop will not reject sending request

apolyakov (dismissed):
LGTM

apolyakov (approved):

PR #1642, opened 2026-06-29
add php_code_commit_hash

Shamzik:
This PR adds new kphp option which can be used to specify php code commit hash

### DIFF
--- a/compiler/code-gen/files/init-scripts.cpp
+++ b/compiler/code-gen/files/init-scripts.cpp
@@ -5,7 +5,10 @@
 #include "compiler/code-gen/files/init-scripts.h"
 
 #include <chrono>
+#include <cstdint>
 #include <string>
+
+#include "common/algorithms/string-algorithms.h"
 
 #include "compiler/code-gen/common.h"
 #include "compiler/code-gen/const-globals-batched-mem.h"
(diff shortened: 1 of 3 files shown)


compiler/code-gen/files/init-scripts.cpp
```diff
@@ -5,8 +5,11 @@
 #include "compiler/code-gen/files/init-scripts.h"
 
 #include <chrono>
+#include <cstdint>
 #include <string>
 
+#include "common/algorithms/string-algorithms.h"
```
apolyakov, line 11: No need to empty line between `#include "common/..."` and `#include "compiler/..."`

compiler/code-gen/files/init-scripts.cpp
```diff
@@ -297,6 +300,7 @@ void CppMainFile::compile(CodeGenerator &W) const {
 
 void ComponentInfoFile::compile(CodeGenerator &W) const {
   kphp_assert(G->is_output_mode_k2());
+  kphp_assert(G->settings().php_code_commit_hash.get().size() == 40);
```
apolyakov: It looks weird that the compiler crashes in case of "invalid" commit hash

apolyakov (dismissed):

compiler/code-gen/files/init-scripts.cpp
Shamzik: fixed

compiler/code-gen/files/init-scripts.cpp
Shamzik: take first 40 instead

astrophysik (approved):

PR #1650, opened 2026-07-09
Get rid of static as a storage for the empty value

PetrShumilov:
Fix segfault which is actual for both runtimes.

In some cases modified empty value can leak over one request to another request.


runtime-common/core/core-types/definition/mixed.inl
```diff
@@ -177,18 +184,74 @@ template<typename T>
 T& mixed::empty_value() noexcept {
   static_assert(vk::is_type_in_list<T, bool, int64_t, double, string, mixed, array<mixed>>{} || is_type_acceptable_for_mixed<T>::value, "unsupported type");
 
-  static T value;
-  value = T{};
-  return value;
+  auto& ctx{RuntimeContext::get()};
+  if constexpr (std::is_same_v<T, bool>) {
+    if (ctx.empty_value.bool_v == nullptr) {
+      ctx.empty_value.bool_v = new (RuntimeAllocator::get().alloc_script_memory(sizeof(T))) T{};
```
apolyakov: Isn't it possible to get rid of allocations in simple cases like `bool`, `double`, `array<mixed>` etc? For example, we may have empty values stored in `RuntimeContext::empty_value`

runtime-common/core/core-types/definition/mixed.inl
```diff
@@ -177,18 +184,74 @@ template<typename T>
...
+      php_assert(raw_mem);
+      return *reinterpret_cast<T*>(type2value->insert({type_id, new (raw_mem) T{}}).first->second);
+    }
+
+    *reinterpret_cast<T*>(it->second) = T{};
+    return *reinterpret_cast<T*>(it->second);
+  }
 }
 
 inline void mixed::reset_empty_values() noexcept {
```
apolyakov, line 232: I think we need a cleanup for K2 mode too. Moreover, why don't we just reset these values with default values instead of `nullptr`?

runtime-common/core/core-types/definition/mixed.inl
```diff
@@ -4,10 +4,17 @@
 
 #pragma once
 
+#include <cstddef>
```
apolyakov, line 7: missing include for `<new>`

runtime-common/core/core-types/definition/mixed.inl
```diff
@@ -177,18 +184,74 @@ template<typename T>
...
+  auto& ctx{RuntimeContext::get()};
+  if constexpr (std::is_same_v<T, bool>) {
+    if (ctx.empty_value.bool_v == nullptr) {
+      ctx.empty_value.bool_v = new (RuntimeAllocator::get().alloc_script_memory(sizeof(T))) T{};
+    }
+    *reinterpret_cast<T*>(ctx.empty_value.bool_v) = T{};
+    return *reinterpret_cast<T*>(ctx.empty_value.bool_v);
+  } else if constexpr (std::is_same_v<T, int64_t>) {
+    if (ctx.empty_value.int64_v == nullptr) {
+      ctx.empty_value.int64_v = new (RuntimeAllocator::get().alloc_script_memory(sizeof(T))) T{};
```
apolyakov: Looks like we need to check the result of allocation for `nullptr`

runtime-common/core/core-types/definition/mixed.inl
```diff
@@ -177,18 +184,74 @@ template<typename T>
 T& mixed::empty_value() noexcept {
   static_assert(vk::is_type_in_list<T, bool, int64_t, double, string, mixed, array<mixed>>{} || is_type_acceptable_for_mixed<T>::value, "unsupported type");
 
-  static T value;
-  value = T{};
-  return value;
+  auto& ctx{RuntimeContext::get()};
+  if constexpr (std::is_same_v<T, bool>) {
```
apolyakov: nit: create helper for repeated fast-path branches

runtime-common/core/core-types/definition/mixed.inl
```diff
@@ -177,18 +184,74 @@ template<typename T>
...
+    }
+    *reinterpret_cast<T*>(ctx.empty_value.mixed_v) = T{};
+    return *reinterpret_cast<T*>(ctx.empty_value.mixed_v);
+  } else if constexpr (std::is_same_v<T, array<mixed>>) {
+    if (ctx.empty_value.array_v == nullptr) {
+      ctx.empty_value.array_v = new (RuntimeAllocator::get().alloc_script_memory(sizeof(T))) T{};
+    }
+    *reinterpret_cast<T*>(ctx.empty_value.array_v) = T{};
+    return *reinterpret_cast<T*>(ctx.empty_value.array_v);
+  } else {
```
apolyakov, line 210: Let's add some guard that we only expect class instances here

runtime-common/core/core-context.h
```diff
@@ -17,6 +17,15 @@ struct RuntimeContext final : vk::not_copyable {
   int32_t show_migration_php8_warning{};
   int32_t php_disable_warnings{};
   uint32_t empty_obj_count{};
+  struct {
+    void* bool_v{};
```
apolyakov: Is there a point to have `void*` instead of `bool*`?

tests/phpt/mixed/primitive/001_set_to_empty_value.php
```diff
@@ -0,0 +1,12 @@
+@ok
+<?php
```
apolyakov: Let's add classes test case at least. Even better to add all cases

runtime-common/core/core-types/definition/mixed.inl
```diff
@@ -177,18 +184,74 @@ template<typename T>
 T& mixed::empty_value() noexcept {
   static_assert(vk::is_type_in_list<T, bool, int64_t, double, string, mixed, array<mixed>>{} || is_type_acceptable_for_mixed<T>::value, "unsupported type");
 
-  static T value;
-  value = T{};
-  return value;
+  auto& ctx{RuntimeContext::get()};
```
apolyakov, line 186: I think we may get rid of RTTI completely. Take a look at following example implementation:
```
  namespace {                                                                                                                                                                                                                                                  
  template<typename T>
  inline constexpr char empty_value_type_tag{};                                                                                                                                                                                                                
  }                                                                                                                                                                                                                                                            

  template<typename T>
  T& mixed::empty_value() noexcept {
    static_assert(vk::is_type_in_list<T, bool, int64_t, double, string, mixed, array<mixed>>{} || is_type_acceptable_for_mixed<T>::value, "unsupported type");

    auto& ctx{RuntimeContext::get()};

    auto init_empty_value = [](void*& ptr) -> T& {
      if (ptr == nullptr) {
        auto* raw_mem{RuntimeAllocator::get().alloc_script_memory(sizeof(T))};
        php_assert(raw_mem);
        ptr = new (raw_mem) T{};
      }
      *reinterpret_cast<T*>(ptr) = T{};
      return *reinterpret_cast<T*>(ptr);
    };

    if constexpr (std::is_same_v<T, bool>) {
      return init_empty_value(ctx.empty_value.bool_v);
    } else if constexpr (std::is_same_v<T, int64_t>) {
      return init_empty_value(ctx.empty_value.int64_v);
    } else if constexpr (std::is_same_v<T, double>) {
      return init_empty_value(ctx.empty_value.double_v);
    } else if constexpr (std::is_same_v<T, string>) {
      return init_empty_value(ctx.empty_value.string_v);
    } else if constexpr (std::is_same_v<T, mixed>) {
      return init_empty_value(ctx.empty_value.mixed_v);
    } else if constexpr (std::is_same_v<T, array<mixed>>) {
      return init_empty_value(ctx.empty_value.array_v);
    } else {
      using type2value_t = kphp::stl::unordered_map<const void*, void*, kphp::memory::script_allocator>;
      if (ctx.empty_value.objects == nullptr) {
        auto* raw_mem{RuntimeAllocator::get().alloc_script_memory(sizeof(type2value_t))};
        php_assert(raw_mem);
        ctx.empty_value.objects = new (raw_mem) type2value_t{};
      }

      auto* type2value{reinterpret_cast<type2value_t*>(ctx.empty_value.objects)};
      const auto type_id{static_cast<const void*>(&empty_value_type_tag<T>)};
      const auto it{type2value->find(type_id)};
      if (it == type2value->end()) {
        auto* raw_mem{RuntimeAllocator::get().alloc_script_memory(sizeof(T))};
        php_assert(raw_mem);
        return *reinterpret_cast<T*>(type2value->insert({type_id, new (raw_mem) T{}}).first->second);
      }

      *reinterpret_cast<T*>(it->second) = T{};
      return *reinterpret_cast<T*>(it->second);
    }
  }
```

apolyakov (changes requested):

runtime-common/core/core-types/definition/mixed.inl
PetrShumilov: Unfortunately, this isn’t possible. It only works for k2, where the RuntimeContext lifetime matches the request lifetime. In baseline KPHP, we’d run into the same problem we already have.

runtime-common/core/core-types/definition/mixed.inl
PetrShumilov: It's not related on this PR

runtime-common/core/core-types/definition/mixed.inl
PetrShumilov: Will be fixed

runtime-common/core/core-types/definition/mixed.inl
PetrShumilov: OK, agree

runtime-common/core/core-types/definition/mixed.inl
PetrShumilov: No need, we already have a guard at the top of function

runtime-common/core/core-context.h
PetrShumilov: Thanks

tests/phpt/mixed/primitive/001_set_to_empty_value.php
PetrShumilov: Will be added

runtime-common/core/core-types/definition/mixed.inl
PetrShumilov: Fiixed

runtime-common/core/core-types/definition/mixed.inl
PetrShumilov: Fixed

runtime-common/core/core-types/definition/mixed.inl
PetrShumilov: Fixed

runtime-common/core/core-context.h
PetrShumilov: fixed

tests/phpt/mixed/primitive/001_set_to_empty_value.php
PetrShumilov: Added

runtime-common/core/core-types/definition/mixed.inl
PetrShumilov: Took into account

runtime-common/core/core-context.h
```diff
@@ -9,6 +9,10 @@
 
 #include "common/mixin/not_copyable.h"
```
apolyakov, line 11: No need for newline here

apolyakov (approved):

runtime-common/core/core-context.h
PetrShumilov: I see that need

PR #1651, opened 2026-07-10
implement array_all function

LLirikkkk:


runtime-common/stdlib/array/array-functions.h
```diff
@@ -49,6 +49,17 @@ array<T> array_diff(const array<T>& a1, const array<T1>& a2, const Proj& key_pro
   return result;
 }
 
+template<class T, class CallbackT>
```
apolyakov: Reducing code duplication is nice, but I don't think it's worth it to deduplicate a single loop over array

apolyakov (approved):
LGTM

PR #1653, opened 2026-07-14
[k2] add check if expression contains call of interruptible function and update code gen of NullCoalesce 

LLirikkkk:


apolyakov (approved):
(👍🏻ᴗ _ᴗ)👍🏻

PR #1664, opened 2026-07-28
implement intrusive list

LLirikkkk:


common/containers/intrusive-list.h
```diff
@@ -0,0 +1,491 @@
...
+
+    return *this;
+  }
+
+  ~list_node_base() {
+    unlink();
+  }
+
+  auto is_linked() const noexcept -> bool {
+    return m_prev != this;
```
apolyakov, line 76: Let's add a comment explaining why it's ok to only check `m_prev` here

common/containers/intrusive-list.h
```diff
@@ -0,0 +1,491 @@
...
+    return m_value;
+  }
+
+  auto value() const noexcept -> const T& {
+    return m_value;
+  }
+};
+
+template<typename T, typename... Tags>
+auto make_list_node(T value) noexcept -> list_node<T, Tags...> {
```
apolyakov: Looks like we don't need this one in public API. `list_node`'s constructors are enough. What do you think?

Also, I think it's worth it to add an in-place constructor like:
```
template<typename... Args>
  explicit list_node(std::in_place_t, Args&&... args)
      : m_value{std::forward<Args>(args)...} {}
```
so we can eliminate redundant move in some cases

common/containers/intrusive-list.h
```diff
@@ -0,0 +1,491 @@
...
+  using iterator_category = std::bidirectional_iterator_tag;
+
+private:
+  static_assert(details::is_list_node_v<Node>, "Node must be a specialization of list_node");
+  static_assert(details::is_tag_of_v<Tag, Node>, "Tag is not one of Node's tags");
+
+  details::list_node_base* m_curr{nullptr};
+
+  explicit list_iterator(const details::list_node_base* node) noexcept
+      : m_curr{const_cast<details::list_node_base*>(node)} {}
```
apolyakov, line 184: Let's add a comment justifying this `const_cast`

common/containers/intrusive-list.h
```diff
@@ -0,0 +1,491 @@
...
+    return !(*this == other);
+  }
+};
+
+template<typename Node, typename Tag = default_tag>
+class list {
+public:
+  using value_type = typename Node::value_type;
+  using size_type = size_t;
+  using difference_type = ptrdiff_t;
```
apolyakov: Let's use `std::ptrdiff_t` as you do in the iterator

runtime-common/core/std/containers.h
apolyakov, line 1: I think we can easily get rid of it. `vk::intrusive::list` is enough

apolyakov (changes requested):

common/containers/intrusive-list-test.cpp
apolyakov, line 1: 1. missing copyright header
2. let's reorder includes here: take a look at `zlib-functions.h` for reference
3. let's also add a test that covers `list_node`'s default constructor

common/containers/intrusive-list.h
```diff
@@ -0,0 +1,491 @@
...
+    return !m_sentinel.is_linked();
+  }
+
+  // complexity O(n)
+  auto size() const noexcept -> size_type {
+    return std::distance(begin(), end());
+  }
+
+  auto clear() noexcept -> void {
+    m_sentinel.unlink();
```
apolyakov, line 399: Should we add a comment that nodes keep technically linked after this operation?

common/containers/intrusive-list.h
```diff
@@ -0,0 +1,491 @@
...
+struct tagged_hook : public list_node_base {};
+
+template<typename... Tags>
+struct tagged_hooks : public tagged_hook<Tags>... {};
+
+} // namespace details
+
+struct default_tag {};
+
+template<typename T, typename... Tags>
```
apolyakov, line 98: What will happen if `Tags` have a duplicate?

common/containers/intrusive-list.h
LLirikkkk: Fixed.

common/containers/intrusive-list.h
LLirikkkk: I agree. Let's make it.

common/containers/intrusive-list.h
LLirikkkk: Fixed.

common/containers/intrusive-list.h
LLirikkkk: Fixed.

runtime-common/core/std/containers.h
LLirikkkk: Ok.

common/containers/intrusive-list.h
LLirikkkk: I think, it's good idea.

common/containers/intrusive-list.h
LLirikkkk: I've added type trait is_unique, to check if Tags have a duplicate. If so, there will be readable compilation error.

common/containers/intrusive-list-test.cpp
LLirikkkk: Fixed.

apolyakov (approved):
Nice work!

PR #1667, opened 2026-08-01
fix erroneous type inference for static class fields of array type

apolyakov:
A static field's init value is stored in php globals and is reset via `hard_reset_var()` on every request startup, where memory allocations are forbidden; its constant initializer must therefore have exactly the same C++ type as the field itself (otherwise a converting `copy_from()` is generated, which allocates memory and crashes)


PetrShumilov (approved):
LGTM

PR #1668, opened 2026-08-04
[k2] add a notification when the stream is closed, fix register_abort_handler

denisichh:


runtime-light/coroutine/poll.h
```diff
@@ -11,6 +11,7 @@ namespace kphp::coro {
 enum class poll_op : uint8_t {
   read,
   write,
+  is_closed,
```
apolyakov: Just to keep naming pattern the same: `is_closed` -> `close`

denisichh: done

runtime-light/coroutine/poll.h
```diff
@@ -11,6 +11,7 @@ namespace kphp::coro {
 enum class poll_op : uint8_t {
   read,
   write,
+  close_writer,
```
apolyakov: Can't we name it just `close`?

apolyakov (approved):
Nice work ⭐

PR #1676, opened 2026-08-12
[k2] add tl::RpcDestActorFlags type

apolyakov:
Add new `tl::RpcDestActorFlags` class that we'll need in the future for confdata component

### DIFF
--- a/runtime-light/tl/tl-types.h
+++ b/runtime-light/tl/tl-types.h
@@ -1090,6 +1090,14 @@ struct uuid final {
   bool fetch(tl::fetcher& tlf) noexcept {
     return lo.fetch(tlf) && hi.fetch(tlf);
   }
+
+  void store(tl::storer& tls) const noexcept {
+    lo.store(tls), hi.store(tls);
+  }
+
+  constexpr size_t footprint() const noexcept {
+    return lo.footprint() + hi.footprint();
+  }
 };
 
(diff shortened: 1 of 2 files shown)


runtime-light/tl/tl-types.h
```diff
@@ -1253,17 +1325,56 @@ class rpcInvokeReqExtra final {
...
 };
 
 struct RpcInvokeReqExtra final {
   tl::rpcInvokeReqExtra inner{};
 
-  bool fetch(tl::fetcher& tlf, const tl::mask& flags) noexcept {
+  bool fetch(tl::fetcher& tlf) noexcept {
     tl::magic magic{};
-    return magic.fetch(tlf) && magic.expect(TL_RPC_INVOKE_REQ_EXTRA) && inner.fetch(tlf, flags);
+    return magic.fetch(tlf) && magic.expect(TL_RPC_INVOKE_REQ_EXTRA) && inner.fetch(tlf);
```
Shamzik: I think the RpcInvokeReqExtra magic should be fetched after flags

Shamzik (changes requested):

runtime-light/tl/tl-types.h
apolyakov: You're right, thanks!

Shamzik (approved):

PR #1678, opened 2026-08-20
add return_code tag

Shamzik:
This PR adds return_code tag to http metrics


server/php-worker.h
```diff
@@ -40,7 +40,7 @@ class PhpWorker {
   php_query_data_t data;
 
   bool paused;
-  bool flushed_http_connection;
```
apolyakov, line 43: We already have a global containing the HTTP code. Can we use it instead?

server/statshouse/statshouse-manager.cpp
```diff
@@ -135,7 +138,10 @@ void StatsHouseManager::add_request_stats(uint64_t script_time_ns, uint64_t net_
   client.metric("kphp_by_host_request_cpu_time", true).tag("system").tag(worker_type).tag(status).write_value(script_system_time_ns);
   client.metric("kphp_by_host_request_init_time", true).tag(worker_type).tag(status).write_value(script_init_time);
   if (process_type == ProcessType::http_worker) {
-    client.metric("kphp_by_host_http_connection_process_time", true).tag(status).write_value(http_connection_process_time);
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Warray-bounds"
```
apolyakov: Let's try finding a way to not use this hack

server/php-worker.h
Shamzik: Done

server/statshouse/statshouse-manager.cpp
Shamzik: see last commit

PR #1683, opened 2026-08-27
add blazingly fast instance_cache implementation for K2

denisichh:
Rewrites instance_cache_store/fetch/update_ttl/delete in the K2 runtime: instead of msgpack-serializing and shipping instances to a separate cache process over RPC, instances are now deep-copied directly into shared memory via the platform's k2::alloc_shared_memory/publish_shared_memory/get_shared_memory.

  Storage layout: class_name_hash(u64) | class_instance shell | inner data. On fetch, the block is reinterpreted in place rather than deserialized.

  Key pieces

  - New "construct-in-place" API on core types (class_instance::clone_in/alloc, array::allocation/copy_in, string::copy_in) -- lets objects be constructed directly into a caller-provided memory span instead of the script memory pool.
  - New shared CRTP visitor base kphp::visitors::instance_deep_basic_visitor<Child> (runtime-common/stdlib/visitors/instance deep-basic-visitor.h), factored out of the legacy runtime's copy processor so both runtimes share the same traversal logic. Legacy's InstanceReferencesCountingVisitor/InstanceDeepCopyVisitor/InstanceDeepDestroyVisitor now build on it too.
  - Two new K2 visitors: instance_deep_copy_visitor (copies an instance graph into a caller-provided block, pinning copies with ExtraRefCnt::for_instance_cache) and instance_deep_estimate_size_visitor (computes the exact size needed upfront so the shared-memory block can be allocated once).                   
  - Alignment: added virtual_builtin_alignof() (dispatches to the dynamic type's alignment for polymorphic clones) plus alignment() helpers on class_instance/array/string, and a padding-aware align_for_chunk(size, align) used with std::align() when carving memory from a pool.                                                                                
  - Removed is_serializable/@kphp-serializable compiler checks for instance-cache calls.


compiler/code-gen/declarations.cpp
```diff
@@ -734,6 +736,12 @@ void ClassDeclaration::compile_get_hash(CodeGenerator& W, ClassPtr klass) {
   compile_class_method(FunctionSignatureGenerator(W).set_const_this(), klass, "int get_hash()", klass->get_hash());
 }
 
+void ClassDeclaration::compile_class_name_hash(CodeGenerator& W, ClassPtr klass) {
+  // hash of the class name, computed once at compile time -- same for every instance,
```
PetrShumilov, line 765: I see that it would be better not to introduce a new field. We already have `get_hash` in `class_instance`, which can be helpful, but it has a specific corner case: `class_instance<T>{}.get_hash() == nullptr, while class_instance<T>{}.alloc().get_hash() != nullptr`. I think you have two possible options:
* Adjust the existing get_hash to provide the required semantics.
* Add a new method (e.g., get_type_hash or similar).

compiler/code-gen/declarations.cpp
```diff
@@ -942,8 +956,13 @@ void ClassDeclaration::compile_virtual_builtin_functions(CodeGenerator& W, Class
 
   compile_class_method(FunctionSignatureGenerator(W).set_const_this(), klass, "size_t virtual_builtin_sizeof()", "sizeof(*this)");
 
+  compile_class_method(FunctionSignatureGenerator(W).set_const_this(), klass, "size_t virtual_builtin_alignof()", "alignof(" + klass->src_name + ")");
```
PetrShumilov: Why do we need this?

runtime-common/core/class-instance/class-instance-decl.inl
```diff
@@ -71,8 +74,17 @@ public:
 
   inline class_instance& operator=(const Optional<bool>& null) noexcept;
   inline class_instance clone() const;
+  // copies the instance into externally provided memory (no allocation/ownership)
+  // memory must be aligned to alignof(T) and >= estimate_memory_usage() bytes
+  // caller must pin it with a special ExtraRefCnt (e.g. for_instance_cache), since the instance never frees it.
+  // Returns a null instance if memory is unfit.
+  inline class_instance clone_in(vk::span<std::byte> memory) const noexcept;
```
PetrShumilov: Do we have an ability to return something like variant?

runtime-common/core/class-instance/class-instance-decl.inl
```diff
@@ -71,8 +74,17 @@ public:
...
   template<class... Args>
   inline class_instance<T> alloc(Args&&... args) __attribute__((always_inline));
+  // constructs an instance in externally provided memory (no allocation/ownership)
+  // leaves it null if memory is smaller than sizeof(T) or misaligned
+  template<class... Args>
+  inline class_instance<T> alloc(vk::span<std::byte> memory, Args&&... args) noexcept __attribute__((always_inline));
```
PetrShumilov: I think that would be better to rename into `alloc_in` based on 2 reasons:
* More regular and coherent with clone_in
* Do not produce a new overload

compiler/code-gen/declarations.cpp
```diff
@@ -942,8 +956,13 @@ void ClassDeclaration::compile_virtual_builtin_functions(CodeGenerator& W, Class
...
+
   compile_class_method(FunctionSignatureGenerator(W).set_const_this(), klass, klass->src_name + "* virtual_builtin_clone()",
                        "new " + klass->src_name + "{*this}");
+
+  compile_class_method(FunctionSignatureGenerator(W).set_const_this(), klass, klass->src_name + "* virtual_builtin_construct_at(void* ptr)",
+                       "new (ptr) " + klass->src_name + "{*this}");
```
PetrShumilov: UB here, placement new returns actual address where the object has been constructed. Need to return pointer from `virtual_builtin_construct_at`

runtime-common/core/class-instance/class-instance-decl.inl
```diff
@@ -113,6 +140,19 @@ public:
     return res;
   }
 
+  template<class S = T>
+  std::enable_if_t<std::is_polymorphic<S>{}, class_instance> virtual_builtin_clone_in(vk::span<std::byte> memory) const noexcept {
+    class_instance res;
+    if (o) {
+      if (unlikely(memory.size() < o->virtual_builtin_sizeof() || reinterpret_cast<std::uintptr_t>(memory.data()) % o->virtual_builtin_alignof() != 0)) {
```
PetrShumilov: I recommend to add nullptr check for `memory.data()`

PetrShumilov (changes requested):

compiler/code-gen/declarations.cpp
denisichh: done

runtime-common/core/class-instance/class-instance-decl.inl
denisichh: done

runtime-common/core/class-instance/class-instance-decl.inl
denisichh: done

runtime-common/core/class-instance/class-instance-decl.inl
denisichh: done

runtime-common/core/core-types/decl/array_decl.inl
```diff
@@ -186,6 +189,36 @@ private:
     inline array_inner& operator=(const array_inner& other) = delete;
   };
 
+  class allocation {
```
PetrShumilov: What do you think about `memory` instead of `allocation`?  

As I see such class should be interpreted as resource.

runtime-common/core/core-types/decl/array_decl.inl
```diff
@@ -186,6 +189,36 @@ private:
...
+          is_vector_{is_vector} {}
+
+  public:
+    // allocates script memory for an array of the given size
+    inline static allocation allocate(int64_t new_int_size, bool is_vector) noexcept;
+    // takes ownership of the beginning of externally provided memory (no allocation):
+    // returns std::nullopt if the memory is smaller than estimate_size(new_int_size, is_vector) or misaligned
+    inline static std::optional<allocation> from_external(vk::span<std::byte> memory, int64_t new_int_size, bool is_vector) noexcept;
+
+    vk::span<std::byte> memory() const noexcept {
```
PetrShumilov: `get`?

runtime-common/core/core-types/decl/array_decl.inl
```diff
@@ -186,6 +189,36 @@ private:
     inline array_inner& operator=(const array_inner& other) = delete;
   };
 
+  class allocation {
+    vk::span<std::byte> mem_;
+    int64_t int_size_{0};
```
PetrShumilov: please, rename to `size_`

runtime-common/core/core-types/decl/array_decl.inl
```diff
@@ -186,6 +189,36 @@ private:
     inline array_inner& operator=(const array_inner& other) = delete;
   };
 
+  class allocation {
+    vk::span<std::byte> mem_;
```
PetrShumilov: maybe `buffer`?

runtime-common/core/core-types/decl/array_decl.inl
```diff
@@ -228,6 +267,12 @@ public:
   template<class T1, class = enable_if_constructible_or_unknown<T, T1>>
   inline array(array<T1>&& other) noexcept __attribute__((always_inline));
 
+  // copies other into externally provided memory (no allocation/ownership).
+  // Memory must be aligned to alignof(array_inner) and have at least other.calculate_memory_for_copying() bytes.
+  // The array never frees this memory, so the caller must protect it with a special ExtraRefCnt (e.g. for_instance_cache).
+  // Returns std::nullopt if the memory is unfit.
+  inline static std::optional<array> copy_in(vk::span<std::byte> memory, const array& other) noexcept;
```
PetrShumilov: `clone_in`?

runtime-common/core/core-types/decl/string_decl.inl
```diff
@@ -57,7 +62,8 @@ private:
     inline char* ref_data() const;
 
     inline static size_type new_capacity(size_type requested_capacity, size_type old_capacity);
-    inline static string_inner* create(size_type requested_capacity, size_type old_capacity);
+    inline static string_inner* create(size_type requested_capacity, size_type old_capacity) noexcept;
+    inline static string_inner* create(vk::span<std::byte> memory, size_type requested_capacity, size_type old_capacity) noexcept;
```
PetrShumilov: `create_in`?

runtime-light/stdlib/instance-cache/instance-cache-functions.h
```diff
@@ -4,166 +4,154 @@
...
   if (ttl < 0) [[unlikely]] {
-    kphp::log::warning("ttl can't be negative: ttl -> {}, key -> {}", ttl, key.c_str());
-    co_return false;
+    kphp::log::warning("instance_cache_store. ttl less than 0, key will be stored forever: ttl -> {}, key -> {}", ttl, key.c_str());
+    ttl = 0;
   }
-  if (ttl > std::numeric_limits<uint32_t>::max()) [[unlikely]] {
-    kphp::log::warning("ttl exceeds maximum allowed value, key will be stored forever: ttl -> {}, max -> {}, key -> {}", ttl,
-                       std::numeric_limits<uint32_t>::max(), key.c_str());
+  if (constexpr int64_t max_ttl{std::numeric_limits<int64_t>::max() / 1000}; ttl > max_ttl) [[unlikely]] {
```
PetrShumilov: Provide a comment regarding sec and milliseconds or rename `max_ttl` to `max_ttl_sec` and `ttl` to `ttl_sec`

runtime-light/stdlib/instance-cache/instance-cache-functions.h
```diff
@@ -4,166 +4,154 @@
...
-  cache_store.store(tls);
-
-  auto expected_stream{kphp::component::stream::open(kphp::instance_cache::details::COMPONENT_NAME, k2::stream_kind::component)};
-  if (!expected_stream) [[unlikely]] {
-    co_return false;
+  const size_t estimated_size{estimate_size_visitor.get_estimated_size()};
+  constexpr size_t instance_size{sizeof(class_instance<InstanceType>)};
+  constexpr size_t hash_size{sizeof(uint64_t)};
+
+  auto alloc_result{k2::alloc_shared_memory(hash_size + instance_size + estimated_size, alignof(std::max_align_t))};
```
PetrShumilov: alignment should be fixed via constant

runtime-light/stdlib/instance-cache/instance-cache-functions.h
```diff
@@ -4,166 +4,154 @@
...
-  if (!co_await kphp::forks::id_managed(kphp::component::query(stream, tls.view(), response))) [[unlikely]] {
-    co_return false;
+  std::byte* mem{static_cast<std::byte*>(alloc_result.value())};
+
+  const uint64_t class_name_hash{InstanceType::get_class_name_hash()};
+  std::memcpy(mem, &class_name_hash, hash_size);
+  // deep-copies the object graph into the inner area and rewrites the instance's fields to point at the copies,
+  // so the whole graph ends up inside the shared-memory block.
+  // All copies are pinned with ExtraRefCnt::for_instance_cache and are never freed individually -- the platform owns the block.
+  kphp::visitors::instance_deep_copy_visitor copy_visitor{std::span{mem + hash_size + instance_size, estimated_size}, ExtraRefCnt::for_instance_cache};
```
PetrShumilov: use `std::next`

runtime-light/stdlib/instance-cache/instance-cache-functions.h
```diff
@@ -4,166 +4,154 @@
...
+  }
+  std::construct_at(reinterpret_cast<class_instance<InstanceType>*>(mem + hash_size), std::move(instance));
+
+  // the platform expects ttl in milliseconds, while the PHP API accepts seconds
+  if (auto publish_result{k2::publish_shared_memory(std::string_view{key.c_str(), key.size()}, mem, ttl * 1000, false, true)}; publish_result.has_value()) {
+    InstanceCacheInstanceState::get().request_cache.insert_or_assign(key, std::span{mem, hash_size + instance_size + estimated_size});
+    return true;
+  } else {
+    // publish is expected to always succeed here (ignore_if_exist=true, valid key/memory), so this should never actually happen.
+    kphp::log::assertion(k2::release_shared_memory(mem).has_value());
```
PetrShumilov, line 82: Does assertion is mandatory ?

runtime-light/stdlib/instance-cache/instance-cache-functions.h
```diff
@@ -4,166 +4,154 @@
...
+  static_assert(is_class_instance_v<ClassInstanceType>, "class_instance<> type expected");
+  constexpr size_t hash_size{sizeof(uint64_t)};
+  constexpr size_t instance_size{sizeof(ClassInstanceType)};
+
+  if (key.empty()) [[unlikely]] {
+    kphp::log::warning("instance_cache_fetch. empty key is not supported");
+    return {};
   }
+  // unwraps and validates a shared memory block: returns a null instance if the block is malformed or belongs to another class
+  const auto unwrap{[&class_name, &key](std::span<const std::byte> mem) noexcept -> ClassInstanceType {
```
PetrShumilov: what about `materialize`?

PetrShumilov (changes requested):

runtime-light/stdlib/instance-cache/instance-cache-functions.h
denisichh: done

runtime-light/stdlib/instance-cache/instance-cache-functions.h
denisichh: done

runtime-light/stdlib/instance-cache/instance-cache-functions.h
denisichh: done

runtime-light/stdlib/instance-cache/instance-cache-functions.h
denisichh: yes

runtime-light/stdlib/instance-cache/instance-cache-functions.h
denisichh: done

PetrShumilov (approved):
OK

PR #1684, opened 2026-08-28
[compiler] reformat with clang-format

denisichh:
- reformat compiler/ with clang-format                                                                                                                                                                             
- update CI format check                                                                                                                                                                                          
- added .clang-format-ignore
- disable ReflowComments


ZemlyakovDmitry (dismissed):

.git-blame-ignore-revs
Danil42Russia, line 1: You can only do this after you’ve committed the PR
Because you have a squash merge, which combines the commits into one and will result in a different SHA

denisichh: ok

Danil42Russia: For example:
- Updated code style: #1281
- Updated `.git-blame-ignore-revs`: #1282

.github/workflows/pr-code-format.yml
Danil42Russia, line 37: Why not add `compiler` here as well? :)

denisichh: I forgot ;). done

PetrShumilov (approved):
LGTM

ZemlyakovDmitry (approved):

PR #1685, opened 2026-08-28
[k2] implement shared-memory confdata

apolyakov:
Implement `kphp-confdata.so` to synchronize snapshots and incremental updates from `confdata-proxy` and expose shared-memory data to PHP instances.

Add wildcard lookups, reader leases that preserve snapshots across piece rotation, memory-budget checks, and capacity metrics. Include the component in build and installation targets.


runtime-light/components/confdata/state/component-state.cpp
```diff
@@ -4,23 +4,110 @@
...
+  const std::string_view storage_view{m_predefined_wildcards_storage};
+  size_t line_number{1};
+  size_t line_begin{};
+  while (line_begin < storage_view.size()) {
+    const size_t line_end{storage_view.find('\n', line_begin)};
+    const auto wildcard{storage_view.substr(line_begin, line_end - line_begin)};
+    if (wildcard.empty()) [[unlikely]] {
+      kphp::log::error("{} contains an empty line: line -> {}", PREDEFINED_WILDCARDS_ARG, line_number);
+    }
+    if (wildcard.contains('\r')) [[unlikely]] {
```
denisichh, line 53: how about doing just one pass over the string instead of two. try std::find_if

runtime-light/stdlib/confdata/confdata-storage.cpp
```diff
@@ -0,0 +1,724 @@
...
+  if (previous != nullptr && equals(*previous, value)) {
+    return false;
+  }
+
+  retire_for_shallow_destruction(mixed{entries});
+  if (previous == nullptr) {
+    entries.set_value(handles.make_remainder_copy(), value);
+  } else {
+    retire_value(*previous);
+    entries.mutate_if_shared();
```
denisichh, line 369: mutate_if_shared()/set_value() here force a full COW copy of the section's whole array. No is_enough_memory_for() guard. Since this is a fixed external pool, overflow falls into request_extra_memory -> assertion. A single-key update on a large wildcard section can crash the whole component. Maybe it would be worth adding a check for sufficient memory before copying and fail the update instead of asserting ?

denisichh (changes requested):

runtime-light/components/confdata/state/component-state.cpp
apolyakov: Thanks for noting, fixed

runtime-light/stdlib/confdata/confdata-storage.cpp
apolyakov: Now we use an approach similar to kPHP's confdata

denisichh (approved):

runtime-light/components/confdata/state/instance-state.cpp
```diff
@@ -43,59 +241,279 @@ auto InstanceState::init() noexcept -> void {
...
+  }
+
+  reader_session session{*this, std::prev(m_confdata_pieces.end())};
+  kphp::log::debug("issuing reader lease: sample -> {}, sections -> {}", session.sample_id(),
+                   m_confdata_pieces.back().storage().values(session.sample_id()).size());
+  const auto lease{kphp::confdata::reader_lease::create(session.shared_memory_name(), session.sample_id())};
+  // The generated shared-memory name and internally acquired sample ID must be valid.
+  // Failure here is an internal invariant violation, not a recoverable client error.
+  kphp::log::assertion(lease.has_value());
+  if (const auto written{co_await connection->get_stream().write_all(std::as_bytes(std::span{std::addressof(*lease), 1}))}; !written) [[unlikely]] {
```
PetrShumilov, line 352: Need to reorder write_all and register_abort_handler.

If `register_abort_handler` fails, `lease` will be destroyed, but `write_all` has already replied to the KPHP instance that everything is OK. This can leave the instance with access to a shared-memory segment that has already been released. Potential segfault

PetrShumilov (changes requested):

PR #1686, opened 2026-08-28
fix test_backtrace

Shamzik:
Enriches KPHP_EXTRA_CXXFLAGS with env


kiry130874-hub:
> Enriches KPHP_EXTRA_CXXFLAGS with env

PR #1690, opened 2026-09-04
[compiler] add missing includes and forward declarations

denisichh:


compiler/code-gen/files/function-header.cpp
```diff
@@ -4,14 +4,23 @@
 
 #include "compiler/code-gen/files/function-header.h"
 
+#include <map>
+#include <set>
+#include <stddef.h>
```
Shamzik, line 9: why not `<cstddef>`?

denisichh: done ;)

apolyakov (approved):

